PR #38, opened 2022-07-22
Update of the workchance of the Abnormalities

Sombrahide:
## About The Pull Request
Work chance changed in some abnormalities so the probability is the same has the game.
Used the wiki (https://lobotomycorp.fandom.com) has guide for the changes.

## Why It's Good For The Game
It's simply more true to the original game, and the works chance will change more depending the level of the worker.

## Changelog
:cl:
balance: rebalanced the work chance of some abnormalities so is more true to the game of Lobotomy Corporation.
/:cl:


.vscode/settings.json
```diff
@@ -7,12 +7,9 @@
 		"tgui/.yarn": true,
 		"tgui/.pnp.*": true
 	},
-	"workbench.editorAssociations": [
```
EgorDinamit, line 10: Why was it changed?

Sombrahide: ops, sorry, in teory i discarded that, i will change it now

Sombrahide: In teory now it works well, it seems like visual studio code made like that when I opened it

PR #39, opened 2022-07-22
Adds Extraction Officer

Kitsunemitsu:
## About The Pull Request
Adds the Extraction Officer to the game, as well as their jacket.
It's 20/20/20/20, so it's not excellent, but they won't be stealing early game ego.

image

## Why It's Good For The Game

So we're a 30+ Pop server pretty regularly.
**One thing that I've noticed is that many people that play captain sit at the Ego console for the entire game**

They shouldn't be doing this, they should be commanding.
So my solution is the Extraction officer! We don't have extraction clerks (and it wouldn't be lore friendly to have them)
So we have a senior Extraction clerk here as an officer, their goal is to recover, buy and distribute ego.

It's better to have a unique job for it, no? We have the pop for it.
With manager, Extraction Officer, 2 captains and 8 Senior Agents, we've got a total of 12 limited roles for 25-40 players. that's enough.

## Changelog
:cl:
add: Added extraction officer
/:cl:


EgorDinamit (changes requested):
1. Why is job_types file called "command.dm"?
2. Why did you add a non-existent job to the list? It will runtime.

Kitsunemitsu:
That runtimes? FUCK
The Command File will be used for all non-manager command if we get any

code/modules/clothing/suits/ego_gear/teth.dm
```diff
@@ -23,3 +23,15 @@
 	Sometimes, the wielder may see things we have long forgotten."
 	icon_state = "fragment"
 	armor = list(RED_DAMAGE = 0, WHITE_DAMAGE = -20, BLACK_DAMAGE = 50, PALE_DAMAGE = 0)
+
+/*Extraction officer coat.
+I don't want them gaming ego right off the bat beacuse this one actually looks pretty okay.
+It's not great though.
+*/
+/obj/item/clothing/suit/armor/ego_gear/extraction
```
EgorDinamit, line 31: I'd suggest you move it outside of EGO gear, since it isn't using ANY of the EGO features, just make it a normal suit.

code/modules/jobs/job_types/agent.dm
```diff
@@ -47,6 +47,7 @@
 	gloves = /obj/item/clothing/gloves/color/black
 	implants = list(/obj/item/organ/cyberimp/eyes/hud/security)
 
+
```
EgorDinamit, line 50: what

Kitsunemitsu: Damn that was a mistake

code/modules/clothing/suits/ego_gear/teth.dm
Kitsunemitsu: Done

PR #66, opened 2022-07-26
Cams updated

PotatoTomahto:
## About The Pull Request
Optimizes camera placements and removes unnecessary ones. Replaces sleepers in welfare and adds vending machines to hallways.

## Why It's Good For The Game
Since Control no longer has advanced camera consoles, they don't have to scroll through 8+ cameras to view an area. Sleepers in welfare were also self-use which is now replaced. Vending machines are cool.

## Changelog
:cl:
add: some more vending machines
del: removed extra cams
tweak: minor adjustments to control, safety, training, and welfare
/:cl:


EgorDinamit:
I don't quite get the point of removing cameras. Did you make sure the entire facility is still viewable from manager's cameras?

Issue #71, opened 2022-07-27
ADD FIRE EXTINGUISHERS TO THE CORP


The-Sun-In-Splendour:
Who would've thought it'd be a good idea to add a fire AOE to white night with no fire extinguishers LMAO
I had to watch +5 people burn to death because unlike on CM you cannot tap people's fire out.

i would do it myself but i refuse to do anything with byond's map editor

PR #72, opened 2022-07-27
Survival box tweak

UnokiAs:
## About The Pull Request

Fixes Issue https://github.com/vlggms/lobotomy-corp13/issues/71
Other fix would be with mapping
Survival box now get a emergency fire extinguisher.
Put a gasmask don't forget to put your gasmask around a certain abnormality :)

## Why It's Good For The Game
Fixes issues with fire with no way to combat it
gasmask is purely flavor 

## Changelog
:cl: Unoki
tweak: tweak the survival box
/:cl:

### DIFF
--- a/code/game/objects/items/storage/boxes.dm
+++ b/code/game/objects/items/storage/boxes.dm
@@ -124,12 +124,13 @@
 
 // Ordinary survival box
 /obj/item/storage/box/survival
-	var/mask_type = /obj/item/clothing/mask/breath
+	var/mask_type = /obj/item/clothing/mask/gas
 	var/internal_type = /obj/item/tank/internals/emergency_oxygen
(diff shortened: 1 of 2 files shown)


NiftyRyanyan:
Why don’t we just put fire lockers in the facility instead, it would make more sense to have these items limited and only grabbed when actually needed instead of handout to everyone.

UnokiAs:
> Why don’t we just put fire lockers in the facility instead, it would make more sense to have these items limited and only grabbed when actually needed instead of handout to everyone.

Mapping issue, This pr is one solution and not the only one

Kitsunemitsu:
This is a fine Solution, I'm approving.

Kitsunemitsu (approved):

PR #84, opened 2022-07-30
Refactors Ego Examine Text [And Ego Diversifacation]

Kitsunemitsu:
## About The Pull Request
Adds a combo system to SST, making it attack faster.
Correctional now has recoil, 
Hornet now shoots faster like originally intended.

Added in an examining weapon feature for special weapons.

## Why It's Good For The Game
Waw+ really should all have some unique features, I'm trying my best.
Weapons are no longer just "This one deals x damage but with a different look" so we need to tell everyone what they actually do

## Changelog
:cl:
add: Adds in a system to examine weapons for a description of what they do
tweak: Correctional has some recoil
tweak: Hornet is now as god intended
tweak: Sword Sharpened with Tears has a combo system
/:cl:


code/game/objects/items/ego_weapons/waw.dm
```diff
@@ -24,18 +24,38 @@
...
 
+//This is like an anime character attacking like 4 times with the 4th one as a finisher attack.
 /obj/item/ego_weapon/despair/melee_attack_chain(mob/user, atom/target, params)
 	..()
-	user.changeNext_move(CLICK_CD_MELEE * 0.75)
-	hitsound = "sound/weapons/ego/rapier[pick(1,2)].ogg"
+	if(combo < 4)
+		force = 20
+		user.changeNext_move(CLICK_CD_MELEE * 0.5)
+		hitsound = "sound/weapons/ego/rapier[pick(1,2)].ogg"
```
EgorDinamit, line 45: Why put it everywhere if the result is always the same?
Also see how Justitia does it, you should've put it in the other proc.

code/game/objects/items/ego_weapons/waw.dm
```diff
@@ -24,18 +24,38 @@
...
+		to_chat(user,"<span class='warning'>You advance with a swing.</span>")
```
EgorDinamit, line 46: This will get annoying REAL fast.

code/game/objects/items/ego_weapons/waw.dm
Kitsunemitsu: Fixed

code/game/objects/items/ego_weapons/waw.dm
Kitsunemitsu: Fixed

EgorDinamit:
> Added in an examining weapon feature for special weapons.
You can already do this with the proc for weapon damage, that's what paradise lost does.

Kitsunemitsu:
Sure showed me then...
I guess the correct name for this PR is "Makes giving egos special ability examine text easier"

Kitsunemitsu:
I need a review and merge, if possible

code/game/objects/items/ego_weapons/_ego_weapon.dm
```diff
@@ -14,6 +14,8 @@
 							JUSTICE_ATTRIBUTE = 0
 							)
 
+	var/special = "This weapon has no special abilities."
```
EgorDinamit, line 17: Variable name is ambiguous, I'd suggest changing it to something like "special_description".

code/game/objects/items/ego_weapons/_ego_weapon.dm
```diff
@@ -22,6 +24,7 @@
 /obj/item/ego_weapon/examine(mob/user)
 	. = ..()
 	. += EgoAttackInfo(user)
+	. += "<span class='notice'>[special]</span>"
```
EgorDinamit, line 27: It'd be a good idea to not display it at all if it's null, which it should be when it doesn't have special stuff.

Kitsunemitsu: Good point, I just had it to set a "This has no special abilitieS" message to keep them all about 4 lines long

code/game/objects/items/ego_weapons/_ego_weapon.dm
```diff
@@ -24,7 +24,8 @@
 /obj/item/ego_weapon/examine(mob/user)
 	. = ..()
 	. += EgoAttackInfo(user)
-	. += "<span class='notice'>[special]</span>"
+	if(special != "This weapon has no special abilities.")	//This is honestly a failsafe for if it doesn't go through. A failsafe debug text, if you would
```
EgorDinamit, line 27: Why not just make special default to null..? :thinking:

Kitsunemitsu: Because the "No abilities" text works to make them have 4 lines by default

PR #85, opened 2022-07-30
Sleeper tweak, pistol buff [BETACORP HOTFIXES]

PotatoTomahto:
## About The Pull Request
Sleepers will now only be able to inject 10 units of probital into someone without probital already. It is still possible to get stamina crit if you return for more probital, but ODs are no longer possible. Soda and Tough pistols have their damage buffed. Removed extra plushies from Safety/Welfare for consistency and added water cooler in Beta training because Alpha has one.

Revamps Tcomms so that the manager's office on Centcom Z-level can actually communicate.

## Why It's Good For The Game
Probital ODs from sleepers are stupid, and there have been too many times where it has ruined me (and other players as well). Pistols do twice the damage of clerk guns so they can have some use as a sidearm or when dual-wielded. Water and plushies are cool.

Manager needs to talk.

## Changelog
:cl:
add: water cooler to training
del: plushies (it had to be done)
tweak: sleepers can no longer give probital ODs (stamina crit still possible)
:cl:

### DIFF
--- a/_maps/map_files/Beta/betacorp.dmm
+++ b/_maps/map_files/Beta/betacorp.dmm
@@ -220,13 +220,9 @@
 /obj/machinery/light{
 	dir = 4
 	},
-/obj/item/paper/guides/jobs/zayin/guide/wellcheers,
+/obj/item/storage/box/drinkingglasses,
 /turf/open/floor/carpet/royalblue,
 /area/department_main/welfare)
-"bm" = (
-/obj/structure/table/wood,
-/turf/open/floor/carpet/orange,
-/area/department_main/training)
 "bn" = (
 /obj/effect/turf_decal/siding/red{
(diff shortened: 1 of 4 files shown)


EgorDinamit:
I don't think pistols need a buff tbh.

EgorDinamit:
They are a ZAYIN weapon. They are not meant to be good.
If anything, they should be slightly better than round-start equipment and that's it.

Kitsunemitsu:
Having pistols at 8 damage is not the worst, considering that often they are sometimes used lategame for people fighting big bird.
Match gun deals 50 on Direct hit and Beak does 6 at rapid fire.
Having 8 damage pistols is not the worst.

EgorDinamit:
Fix conflicts and I'll get it merged.

PR #87, opened 2022-07-31
Agent Rifleman

Kitsunemitsu:
## About The Pull Request
Adds in Agent Rifleman

## Why It's Good For The Game
Since the removal of Sr. Agent we've been missing variety in the agent roles.
So I have two roles planned to add some. The First of which is Agent Rifleman.
They can pick up a 10 red/10 white rifle that is useful for early game, and they're generally just a bit of flavor.
You like ranged: you pick rifleman.

Still to do: Edit maps too

## Changelog
:cl:
add: Added Agent Rifleman
/:cl:


EgorDinamit:
Make sure new R corp gun sprites appear correctly in game.
I am also not sure if the entire rifle thing warrants a new role when it's basically still Agent, but with ability to open lockers without breaking them open.

Kitsunemitsu:
It's more a specialization.
We are really hurting for variety right now, honestly.

EgorDinamit:
If it's absolutely same as agent, why not make it an alt title? I **really** don't think we need a role whose entire difference is ability to open lockers without hitting them with a weapon twice.

Kitsunemitsu:
Give me about 3 days and I'll get this PR in better shape.
The end goal would be a specialist role with various upgradable gear, it wasnt exactly clear in this PR.
It would be an interaction with ordeal butchering that I will implement.

Lockers will be replaced with a vendor you can use once per Ckey.

Kitsunemitsu:
Going to do a future PR for this, it's out of the scope of this one

PR #118, opened 2022-08-07
Morgue Nerf

Kitsunemitsu:
## About The Pull Request
Morgue can only cremate dead people and punishing bird

## Why It's Good For The Game
image
image
image
image

## Changelog
:cl:
balance: Cremators can only cremate dead people and punishing bird
/:cl:

### DIFF
--- a/code/game/objects/structures/morgue.dm
+++ b/code/game/objects/structures/morgue.dm
@@ -266,15 +266,7 @@ GLOBAL_LIST_EMPTY(crematoriums)
 		update_icon()
 
 		for(var/mob/living/M in conts)
-			if (M.stat != DEAD)
-				M.emote("scream")
-			if(user)
-				log_combat(user, M, "cremated")
-			else
-				M.log_message("was cremated", LOG_ATTACK)
-
-			M.death(1)
-			if(M) //some animals get automatically deleted on death.
+			if (M.stat == DEAD)	//Only kill dead people
 				M.ghostize()
 				qdel(M)
 


code/game/objects/structures/morgue.dm
```diff
@@ -266,17 +266,13 @@ GLOBAL_LIST_EMPTY(crematoriums)
 		update_icon()
 
 		for(var/mob/living/M in conts)
-			if (M.stat != DEAD)
+			if (M.stat == DEAD	|| M.name == "Punishing bird")	//Only kill dead people and punishing bird
```
EgorDinamit, line 269: Why are you checking for name?
Also, no need to even add this, punishing bird is being updated already.

code/game/objects/structures/morgue.dm
```diff
@@ -266,17 +266,13 @@ GLOBAL_LIST_EMPTY(crematoriums)
...
+				M.death(1)
```
EgorDinamit, line 270: What's the point of it if mob is already dead?

code/game/objects/structures/morgue.dm
```diff
@@ -266,17 +266,13 @@ GLOBAL_LIST_EMPTY(crematoriums)
...
-			if(user)
-				log_combat(user, M, "cremated")
-			else
-				M.log_message("was cremated", LOG_ATTACK)
-
-			M.death(1)
-			if(M) //some animals get automatically deleted on death.
 				M.ghostize()
 				qdel(M)
+			if (M.stat != DEAD)	//Don't kill dead people
```
EgorDinamit, line 274: Why is it here if you are checking for it above?

code/game/objects/structures/morgue.dm
Kitsunemitsu: Its for old time's sake, honestly.

code/game/objects/structures/morgue.dm
Kitsunemitsu: Good point, sry i coded it at like 6am

code/game/objects/structures/morgue.dm
Kitsunemitsu: I was gonna put a message here, decided against it.

EgorDinamit (changes requested):
Why did you delete all of the above logging stuff? It's useful.

Kitsunemitsu:
It was related to people getting KILLED by the cremator, you no longer can get killed by it, so there's no need. Everyone is dead that is cremated and no one can be revived

PR #127, opened 2022-08-09
Adds Rudolta of the Sleigh

gailelel:
## About The Pull Request
Adds Rudolta and the Ego required.

## Why It's Good For The Game
More abnormalities are always a good thing, and this one is a white/red HE, which we are sorely needing

## Changelog
:cl:
add: Added Rudolta and EGO
/:cl:


Kitsunemitsu:
I personally walked her through this, if you have any questions/concerns, come to me and I'll bring it up.

code/modules/mob/living/simple_animal/abnormality/he/rudolta.dm
```diff
@@ -0,0 +1,70 @@
...
+	return
+
+/mob/living/simple_animal/hostile/abnormality/rudolta/Life()
+	. = ..()
+	if(!.) // Dead
+		return FALSE
+	if((pulse_cooldown < world.time) && !(status_flags & GODMODE))
+		WhitePulse()
+
+/mob/living/simple_animal/hostile/abnormality/rudolta/CanAttack(atom/the_target)
```
EgorDinamit, line 58: So you want it to be able to attack people..?

gailelel: We tested this; it's like this to follow someone around, and doesn't actually attack people. If it wasn't, then Rudolta would just stay in containment.

code/modules/mob/living/simple_animal/abnormality/he/rudolta.dm
```diff
@@ -0,0 +1,70 @@
...
+	if(ishuman(the_target))
```
EgorDinamit, line 59: Regardless, this is absolutely pointless.
There is a variable just for that anyway, and we don't need it to target only humans anyway.

EgorDinamit (changes requested):
Increase its move_to_delay to 5-6.
Make it unable to attack(see how Silent Orchestra and violet ordeals work).

Kitsunemitsu:
Hey, Egor, So
I took this into a local server, and it works exactly as intended.
It can move around, following players, dealing AOE damage.
It does 0 damage, so it can't really "attack people"
Otherwise it just stays in it's room and does nothing.

PR #134, opened 2022-08-10
Extraction-Themed Morgue Slabs

Doctor-Derp:
## About The Pull Request
Adds a variant of morgue slabs to fit the extraction theme. Does not remove normal morgue slabs.
image
image

## Why It's Good For The Game
Themed objects are always nice.

## Changelog
:cl:
add: Adds an extraction themed variant to morgue slabs.
/:cl:


code/game/objects/structures/morgue.dm
```diff
@@ -203,6 +217,58 @@ GLOBAL_LIST_EMPTY(bodycontainers) //Let them act as spawnpoints for revenants an
...
+/obj/structure/bodycontainer/morgue/extraction/Initialize()
+	. = ..()
+	connected = new/obj/structure/tray/e_tray(src)
+	connected.connected = src
+
+/obj/structure/bodycontainer/morgue/extraction/examine(mob/user)
+	. = ..()
+	. += "<span class='notice'>The speaker is [beeper ? "enabled" : "disabled"]. Alt-click to toggle it.</span>"
+
+/obj/structure/bodycontainer/morgue/extraction/AltClick(mob/user)
```
EgorDinamit, line 242: Shouldn't all this stuff be moved to the parent type?

code/game/objects/structures/morgue.dm
```diff
@@ -203,6 +217,58 @@ GLOBAL_LIST_EMPTY(bodycontainers) //Let them act as spawnpoints for revenants an
...
+	..()
+	if(!user.canUseTopic(src, !issilicon(user)))
+		return
+	beeper = !beeper
+	to_chat(user, "<span class='notice'>You turn the speaker function [beeper ? "on" : "off"].</span>")
+
+/obj/structure/bodycontainer/morgue/extraction/update_icon()
+	if (!connected || connected.loc != src) // Open or tray is gone.
+		icon_state = "extract_morgue_open"
```
EgorDinamit, line 251: It'd be a good idea to make it a variable at parent object, to avoid code copy-pasting.

EgorDinamit (changes requested):
Instead of copy-pasting code for each subtype, just make those procs in a parent object.

code/game/objects/structures/morgue.dm
Doctor-Derp: Done!

code/game/objects/structures/morgue.dm
Doctor-Derp: Done, does it look okay?

PR #149, opened 2022-08-12
Attribute level fix, update; Freischutz bullet tweaks & fixes

EgorDinamit:
## About The Pull Request

- Fixes an issue with division by zero in attributes check.
- User's level is now displayed as roman numerals(I - V and EX) in show attributes menu.
- Der Freischutz magic bullet is, potentially, fixed. Additionally, it does slightly more damage.
- EGO Magic bullet now doesn't pass through walls, has decreased fire rate and higher damage.

## Why It's Good For The Game

- A fix.
- I think it just looks neat.
- A fix + balancing; Direct hit will instantly kill a clerk with 0 armor, which seems to be accurate to the original game. Everyone else, however, will most likely survive it.
- No more cheese strategies; Slower fire rate fits more with the firing sound, same with higher damage.


code/modules/clothing/suits/ego_gear/_ego_gear.dm
```diff
@@ -3,6 +3,7 @@
 	desc = "You aren't meant to see this."
 	icon = 'icons/obj/clothing/ego_gear/suits.dmi'
 	worn_icon = 'icons/mob/clothing/ego_gear/suit.dmi'
+	flags_inv = HIDEJUMPSUIT
```
Kitsunemitsu, line 6: You would have to check with Redacted if Remorse is supposed to go above the suit or not.

EgorDinamit: They did mention that it'd be nice if it hid jumpsuit.

PR #154, opened 2022-08-14
Silent Girl and White Lake Fix

dsmith328:
## About The Pull Request
Updated the Records.
Silent Girl can now allow you to be un-panicked.
White Lake no longer puts a permanent AI in you.
Flower Waltz is no longer removable until the Champion is sane or killed.

## Why It's Good For The Game
It Fixes the AI issues on both of these Abnormalities. We could also probably base more off this, like Red Shoes.

## Changelog
:cl:
add: Dropping Restrictions to Flower Waltz
del: Champion Nulling, no longer required.
tweak: AI Removal and Champion State application.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/silent_girl.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/silent_girl.dm
@@ -26,21 +26,25 @@
 	var/mob/living/carbon/human/guilty_people = list()
 	var/mutable_appearance/guilt_icon  // Icon for the effect
 
+/mob/living/simple_animal/hostile/abnormality/silent_girl/Initialize(mapload)
+	. = ..()
+	guilt_icon = mutable_appearance('ModularTegustation/Teguicons/tegu_effects.dmi', "guilt", -MUTATIONS_LAYER) // Doesn't like me assigning this in the definition, so I assign it here.
+
 /mob/living/simple_animal/hostile/abnormality/silent_girl/proc/DriveInsane(mob/living/carbon/human/target)
-	target.adjustSanityLoss(-500)
-	QDEL_NULL(ai_controller)
+	if (!target.sanity_lost)
+		target.adjustSanityLoss(-500)
+	QDEL_NULL(target.ai_controller)
 	target.ai_controller = /datum/ai_controller/insane/release/silent_girl
-	ghostize(1)
 	target.InitializeAIController()
-	addtimer(CALLBACK(src, .proc/killguilty, target), 600) // If panicked after a minute, KILLS THEM | ADDENDUM: Will eventually function like this, given AI removal becomes prevalent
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/abnormality/he/silent_girl.dm
```diff
@@ -86,7 +88,7 @@
 	if (!LAZYLEN(guilty_people)) // No Guilty on 0 counter? Find a random person and take them <3
 		for(var/mob/living/carbon/human/sucker in GLOB.mob_living_list)
 			if(!sucker.sanity_lost && sucker.stat != DEAD)
-				guilty_people += sucker
+				Guilt_Effect(sucker)
 				break
 	for(var/i in guilty_people) // Drive all Guilty people insane on Qliphoth 0
 		DriveInsane(i)
```
EgorDinamit, line 94: This all could be reworked to pick someone randomly out of a list, instead of always picking the first one on the global list.
It would also be nice if DriveInsane was on timer, just so that there's a bit of time between seeing what's about to happen and it actually happening; This would both look better and provide target with an opportunity to yell "Going insane in control".

code/modules/mob/living/simple_animal/abnormality/he/white_lake.dm
```diff
@@ -87,15 +88,34 @@
...
 	//No requirements because who knows who will use it.
 
+// Sets the weapon to not be droppable until it's moved from the main hand. No more leg sweeping.
+/obj/item/ego_weapon/flower_waltz/equipped(mob/user, slot, initial = FALSE)
+	.=..()
+	if (unequipped)
+		return
+	if (slot == ITEM_SLOT_HANDS && !unequipped)
+		RegisterSignal(user, COMSIG_LIVING_DEATH, .proc/Champion_Death)
+		ADD_TRAIT(src, TRAIT_NODROP, src)
```
EgorDinamit, line 109: This all should probably be done in WL's code when it's put into champion's hands.
It will prevent admin-spawned sword from being undroppable.

dsmith328: Due to the variable starting as False, it should never be truly undroppable, as it would simply take a swap to belt/suit storage to permanently remove the trait from it. I felt having all the equip and undroppable code in one spot as well was the best choice, otherwise it would've been entirely the removal code here and entirely the adding code in the champion application.

EgorDinamit: NODROP trait prevents it from being moved anywhere, including backpack, pockets and any other slots on you.

dsmith328: it does not prevent it from being belt-slotted or suit-slotted. I've tested this.

dsmith328:
I have literally no reason why Run Linters didn't work

PR #158, opened 2022-08-14
Made Crew sensors show everyone regardless of the z-level

UnokiAs:
## About The Pull Request
Commented out the code to check for z-level so it show everyone regardless of the z-level. So Manager can use the crewmonitor again.
Why not delete it ?
For easier revert (In case we get multi-z not likely).

## Why It's Good For The Game
Fix manager crewmonitor.

## Changelog
:cl: Unoki
fix: Manager crewmonitor
/:cl:

### DIFF
--- a/code/game/machinery/computer/crew.dm
+++ b/code/game/machinery/computer/crew.dm
@@ -147,10 +147,11 @@ GLOBAL_DATUM_INIT(crewmonitor, /datum/crewmonitor, new)
 		if(!pos)
 			stack_trace("Tracked mob has no loc and is likely in nullspace: [tracked_living_mob] ([tracked_living_mob.type])")
 			continue
-
+		/**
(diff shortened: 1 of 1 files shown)


EgorDinamit:
This isn't a proper solution, as it will show rabbits and admins in thunderdome.
What we need is multi-z station with manager's office on one of them.

UnokiAs:
> This isn't a proper solution, as it will show rabbits and admins in thunderdome. What we need is multi-z station with manager's office on one of them.

It's one of the solutions/temporary, That's why I commented out the code instead of deleting.
Made this PR because I assumed that multi-z level will never happen.

image

Kitsunemitsu:
If it's just the manager's office, IDGAF
It's the RO and EO being moved that I don't like

EgorDinamit (changes requested):
You also uncommented nullspace check, fix it.

PR #171, opened 2022-08-17
Life For A Daredevil Buff and Tweak

dsmith328:
## About The Pull Request
Buffs Life for a Daredevil's parry by making it not punish you if you land it and giving it +0.1 seconds of active time. Also, it can parry attacks such as Goodbye, which it was supposed to do from the start.

## Why It's Good For The Game
It was literally not being used at all due to the 100% debuff rate.

## Changelog
:cl:
balance: Buffed the sword of Crumbling Armor
soundadd: added a Parry sound
/:cl:


EgorDinamit:
No, never use godmode flag.

Kitsunemitsu:
Just set physiology to 0 for all damage types, and remove it after. Godmode is unnecessary. 
You might have to store it in a separate variable due to how multiplying anything by 0 gives you 0. Or just make it a 95% damage reduction

EgorDinamit (changes requested):
The way it's done right now will override any potential changes done to the physiology armor outside of this weapon. This is not good.

PR #184, opened 2022-08-20
Ordeal suppression bonuses

dsmith328:
## About The Pull Request
It's like the Suppression Bonuses, but Gutted to only work on Ordeals.

## Changelog
:cl:
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/_ego_weapon.dm
+++ b/code/game/objects/items/ego_weapons/_ego_weapon.dm
@@ -19,6 +19,11 @@
 /obj/item/ego_weapon/attack(mob/living/target, mob/living/carbon/human/user)
 	if(!CanUseEgo(user))
 		return FALSE
+	if (target.stat != DEAD)
+		if(istype(target, /mob/living/simple_animal/hostile/ordeal))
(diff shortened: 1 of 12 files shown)


EgorDinamit:
Solo killing an amber dawn bug(There's approximately 20 of those) will net you 10 points. This is not a good balance, this should not be done this way.

I believe we should **not reward** people for suppressing anything, we should **punish** them for not doing it. As to how to do it is a more difficult question, but the first thing that comes to mind is halting stat gains while ordeals are active.

EgorDinamit:
Do keep in mind, even if I were to approve the idea, the way it's done isn't exactly good.
It creates the issue where killing ordeals on your own is more profitable than working together, thus providing a needless conflict with "stealing kills" if someone does attempt to help you. The only thing I can see with such a feature is breeding toxic behavior, where people prevent others from finding out the location of ordeal mobs, trying to get the kill on their own.

Auris456852:
If we're going the route of punishment rather than reinforcement... What if you just couldn't work abnormalities without manager approval if something needs suppression? Put a big blob on the reasoning behind it in the #random-ideas Discord channel, because I am not sure if dropping it all in a PR comment is a good idea.

PR #185, opened 2022-08-20
ML can now fire her melting blob point blanc

UnokiAs:
## About The Pull Request
ML now can fire point blanc.
Melting blob projectile now convert on impact with a dead human.
Some walk speed tweak
Remove the mobs pick target (It made her AI very sluggish...)
Reduce the spread a little and buff Adoration damage to not be like der freis with just dot

https://user-images.githubusercontent.com/82665345/185787727-69a84296-a6c1-409e-8c32-193eb66e5b9e.mp4

## Why It's Good For The Game
She can now use her projectile more.

## Changelog
:cl: Unoki
tweak: Melting speed, Melting Blob convert on hit, Melting fire point blanc, Adoration buff

/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/ego_bullets/aleph.dm
+++ b/code/modules/projectiles/projectile/ego_bullets/aleph.dm
@@ -12,7 +12,7 @@
 	damage = 88
 	damage_type = BLACK_DAMAGE
 	flag = BLACK_DAMAGE
-	spread = 15
+	spread = 10
 	speed = 1.2
 	hitsound = "sound/effects/footstep/slime1.ogg"
 
--- a/code/modules/mob/living/simple_animal/abnormality/aleph/melting_love.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/aleph/melting_love.dm
@@ -22,10 +22,10 @@
 	melee_damage_upper = 50
 	projectiletype = /obj/projectile/melting_blob
 	ranged = TRUE
-	minimum_distance = 1
+	minimum_distance = 0
 	ranged_cooldown_time = 5 SECONDS
 	speed = 2
-	move_to_delay = 5
+	move_to_delay = 4
 	/* Works */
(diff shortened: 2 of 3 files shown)


EgorDinamit:
I don't think it's a good idea, it will only make fighting ML easier, since that's how retreat works, it will prevent them from attacking entirely, so you could just block their movement at a corner and get in free hits.

UnokiAs:
> I don't think it's a good idea, it will only make fighting ML easier, since that's how retreat works, it will prevent them from attacking entirely, so you could just block their movement at a corner and get in free hits.

Alright I made her fire point blanc now

code/modules/projectiles/projectile/ego_bullets/aleph.dm
```diff
@@ -9,10 +9,10 @@
 	name = "slime projectile"
 	icon_state = "slime"
 	desc = "A glob of infectious slime. It's going for your heart."
-	damage = 88
+	damage = 100
```
EgorDinamit, line 12: See below. It does damage over time, so with your buff it will be ~150 black damage.

EgorDinamit: I'm not saying it should be nerfed, but I'm not sure if that was your intent here.

UnokiAs: Eh I saw the impact damage of ego_magicbullet (80 Black) while having a better firerate/projectile speed but yeah I can revert the damage buff if needed

EgorDinamit: Wasn't needed, but I'll see how the weapon performs in the game.

UnokiAs: I revert it, I will leave it to someone that eventually do a ego weapon rebalance better have one big pr like the armor one

PR #195, opened 2022-08-23
White Night Rework

EgorDinamit:
## About The Pull Request

Reworks White Night.
Instead of converting random players to its side, it will just spawn the 11 apostles on breach and teleport to random department, periodically sending revival pulse attack. On its own WN will be unable to perform any attacks.
Apostles are simple mobs now as well, with slightly different attacks.

## Why It's Good For The Game

Removes one of the few PVP sources with forced conversion. Additionally, this makes fighting WN much more difficult.


code/modules/mob/living/simple_animal/abnormality/aleph/white_night.dm
```diff
@@ -196,258 +91,386 @@ GLOBAL_LIST_EMPTY(apostles)
...
-				if(ishuman(L))
-					var/mob/living/carbon/human/H = L
-					new /obj/effect/temp_visual/dir_setting/cult/phase(T, H.dir)
-					addtimer(CALLBACK(src, .proc/revive_target, H, i, faction_check))
-					continue
-				// Not a human and an enemy
-				if(!(faction_check in L.faction))
-					playsound(L.loc, 'sound/machines/clockcult/ark_damage.ogg', 35, TRUE, -1)
-					L.adjustPaleLoss(holy_revival_damage)
+			for(var/mob/living/L in T)
```
dsmith328, line 94: Pretty sure this allows people to get multi-hit, as I've seen multiple people report that they ARE being multi-hit by this. Which is not really good considering this can do like, 60 pale damage easily? Not as much of an issue due to the Better Regenerators change but the multi-hit pale is DEFINITELY an issue.

code/modules/mob/living/simple_animal/abnormality/aleph/white_night.dm
```diff
@@ -196,258 +91,386 @@ GLOBAL_LIST_EMPTY(apostles)
...
-		for(var/turf/open/T in fire_zone)
-			new /obj/effect/temp_visual/fire/fast(T)
-			for(var/mob/living/L in T.contents)
-				if("apostle" in L.faction)
-					continue
-				L.apply_damage(20, PALE_DAMAGE, null, L.run_armor_check(null, PALE_DAMAGE), spread_damage = TRUE)
-				to_chat(L, "<span class='userdanger'>You're hit by [src]'s fire field!</span>")
-		SLEEP_CHECK_DEATH(1.5)
+		if(istype(L, /mob/living/simple_animal/hostile/apostle) && L.stat == DEAD)
+			L.revive(full_heal = TRUE, admin_revive = FALSE)
```
dsmith328, line 110: This probably shouldn't full heal, as it kinda makes suppressing them at all entirely rewardless and just encourages quite literally putting them somewhere else. Maybe a half health revive so all the work done by agents aren't entirely instantly undone, but still enough so that they can come back from it and be a threat? Also the Apostles seem to fade on death but immediately pop back into reality when revived, can we get a downed Sprite for them?

dsmith328: Something just occured to me was that you can probably gib apostles with Smile weapon. Might want a check to count apostles and spawn a new one if they get gibbed? Uncertain.

EgorDinamit: You cannot gib apostles, at all.
And full revive will stay: You are not supposed to fight apostles, you are supposed to go for the White Night.

PR #203, opened 2022-08-25
Better Regenerators

dsmith328:
## About The Pull Request
Initial Change is that Regenerators simply heal % health instead of flat. This is nice because high level employees genuinely have to wait at least 2 entire minutes of doing NOTHING to heal to full, be it sanity or physical damage. Now it's about a minute to full heal, though I'd entirely be down to change Regenerators base heal to 2% per 2 seconds as well.
Furthermore, this adds a new Clerk-Only device they can use called the Safety Department Regenerator Augmentation Kit, or the R.A.K. for short. This device can be used to augment the Regenerator for a short period of time to provide various effects. These being:
1. Increase HP Regeneration by 3 but decrease SP Regeneration by 1. (Duration: 60 Seconds)
2. Increase SP Regeneration by 3 but decrease HP Regeneration by 1. (Duration: 60 Seconds)
3. Increase both HP and SP Regeneration by 1. (Duration: 30 Seconds)
4. Allow the Regenerator to heal Insane and Critically Injured people but lose 1 HP and SP regeneration. (Duration: 30 Seconds)
5. Set the Regenerator to Overload, healing for 7.5x times the normal amount (or 15 seconds worth) on the next use, but then turning off for 30 seconds.

All of these functions require the Clerk to work on the device for 5 seconds without being interrupted and all numbers are negotiable.

## Why It's Good For The Game
Gives more purpose to Clerks existing and helps solve the "Time to sit in this room for ages while I heal" issue that's been happening near the end game of rounds. While I do think the numbers might be a bit high (I'd be down to take the Regenerator base 3% to 2%, honesty) I feel like this will help speed things along. The halved healing rate still applies to all of these as well. Whether not the kits are made limited on the map or just standard for Clerks is up in the air as well.

I've heard some voices as well about this potentially being a decent replacement for the EMAIS, though that's entirely opinion based.

## Changelog
:cl:
add: Added the Clerk-Only R.A.K.
add: R.A.K. Sprites by "a Ghost#7213"
tweak: Changed the Regenerators to heal % health and sanity.
/:cl:


code/game/machinery/regenerator.dm
```diff
@@ -19,15 +28,80 @@
 		if(!("neutral" in L.faction)) // Enemy spotted
 			regen_amt *= 0.5
 			break
+	if(burst)
+		regen_amt *= 7.5
+		burst = FALSE
+		cooldown = TRUE
+		addtimer(CALLBACK(src, .proc/Reset_Cooldown), 30 SECONDS)
```
EgorDinamit, line 35: Instead you should use variable cooldowns, see how abnormalities do this.

code/game/machinery/regenerator.dm
```diff
@@ -19,15 +28,80 @@
...
+/obj/machinery/regenerator/attackby(obj/item/I, mob/living/user, params)
+	if(I.tool_behaviour == TOOL_SAFETYKIT)
+		if(user?.mind?.assigned_role != "Clerk")
+			to_chat(user,"<span class='warning'>You don't know how to use this.</span>")
+			return FALSE
+		if (modified)
+			to_chat(user, "<span class='notice'>The [src] is already modified.</span>")
+			return FALSE
+		to_chat(user, "<span class='notice'>You begin tinkering with the [src].</span>")
+		if(do_after(user, 5 SECONDS))
```
EgorDinamit, line 62: Add a check if someone modifies it in the middle of the do_after below.

code/__DEFINES/tools.dm
```diff
@@ -18,6 +18,7 @@
 #define TOOL_KNIFE			"knife"
 #define TOOL_BLOODFILTER	"bloodfilter"
 #define TOOL_ROLLINGPIN		"rollingpin"
+#define TOOL_SAFETYKIT		"safetykit"
```
EgorDinamit, line 21: I am fairly sure you don't need it, just do istype check in the regenerator code.
Tool types are only really used when several unrelated objects can do same thing(I.e. knives being used as screwdriver).

code/game/machinery/regenerator.dm
dsmith328: This one is SPECIFICALLY for turning off the machine for 30 seconds after the "Burst" function has been used. The others use a timer but I'll see what I can do to standardize the other timers.

PR #212, opened 2022-08-26
WN fix and buff

EgorDinamit:
## About The Pull Request

- Fixes issue with apostles being unable to move after revival.
- Gives WN a bit more health and shortens the revival cooldown.

## Why It's Good For The Game

- A fix.
- WN was a bit too easy to kill.

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/aleph/white_night.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/aleph/white_night.dm
@@ -3,8 +3,8 @@ GLOBAL_LIST_EMPTY(apostles)
 /mob/living/simple_animal/hostile/abnormality/white_night
 	name = "White night"
 	desc = "The heavens' wrath. Say your prayers, heretic, the day has come."
-	health = 12000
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/abnormality/aleph/white_night.dm
```diff
@@ -3,8 +3,8 @@ GLOBAL_LIST_EMPTY(apostles)
 /mob/living/simple_animal/hostile/abnormality/white_night
 	name = "White night"
 	desc = "The heavens' wrath. Say your prayers, heretic, the day has come."
-	health = 12000
-	maxHealth = 12000
+	health = 15000
```
Kitsunemitsu, line 6: Egor, he already had 3 times the amount of HP as every other Aleph, he doesn't need 4x.
This is too much.

EgorDinamit: This is a static target that does not move. It is very much needed.

PR #216, opened 2022-08-27
Observation Bonuses

dsmith328:
## About The Pull Request
So, we lacked the wonderful mechanic that was Observation Bonuses from Lobotomy Corporation, right? But how do we do that when we just kinda know all the Abno info as is? Simple! Make the Observation bonuses a  reward for working on the Abnormality! With this change, work on Abnormalities will become slightly easier and faster the better work is done on it. Though the more dangerous the Abnormality, the harder it is to truly understand it. Zayins and Teths get up to a +10% bonus to work speed and success rate, while HEs get up to +8%, and WAWs and Alephs get up to +6%. All of these build up at the same rate (10 Good Results, 20 Neutral Results, or any mix of the two), and at Max rating, the abnormality gains a 50% increased gift chance (So a Zayin would have 7.5% instead of 5%).

## Why It's Good For The Game
Works were overall a little more dangerous than they should've been, and this should help fix that issue. The longer an Abnormality is in the facility and actively worked with, the less dangerous it'll be to those who need to work on it... usually. Some abnormalities are just that temperamental and this doesn't really interfere with that.

## Changelog
:cl:
add: Added the "Understanding" mechanic.
add: Added "Understanding" progress to the pop-up for Abnormality Consoles.
/:cl:

### DIFF
--- a/code/datums/abnormality/datum/abnormality.dm
+++ b/code/datums/abnormality/datum/abnormality.dm
@@ -39,6 +39,9 @@
 	var/overload_chance_amount = 0
 	/// Limit on overload_chance; By default equal to amount * 10
 	var/overload_chance_limit = 100
+	/// Simulated Observation Bonuses
+	var/understanding = 0
+	var/max_understanding = 0
 
 /datum/abnormality/New(obj/effect/landmark/abnormality_spawn/new_landmark, mob/living/simple_animal/hostile/abnormality/new_type = null)
 	if(!istype(new_landmark))
@@ -65,6 +68,8 @@
 	current.toggle_ai(AI_OFF)
 	current.status_flags |= GODMODE
 	current.setDir(EAST)
+	if (understanding == max_understanding)
+		current.gift_chance *= 1.5
 	threat_level = current.threat_level
 	qliphoth_meter_max = current.start_qliphoth
 	qliphoth_meter = qliphoth_meter_max
@@ -76,10 +81,18 @@
 	neutral_boxes = round(max_boxes * 0.4)
 	available_work = current.work_chances
 	switch(threat_level)
+		if(ZAYIN_LEVEL)
+			max_understanding = 10
+		if(TETH_LEVEL)
+			max_understanding = 10
+		if(HE_LEVEL)
+			max_understanding = 8
 		if(WAW_LEVEL)
 			overload_chance_amount = -2
+			max_understanding = 6
 		if(ALEPH_LEVEL)
 			overload_chance_amount = -4
+			max_understanding = 6
 	overload_chance_limit = overload_chance_amount * 10
 
 /datum/abnormality/proc/FillEgoList()
@@ -93,6 +106,14 @@
 
 /datum/abnormality/proc/work_complete(mob/living/carbon/human/user, work_type, pe, max_pe, work_time)
 	current.work_complete(user, work_type, pe, work_time) // Cross-referencing gone wrong
+	if (understanding != max_understanding) // This should render "full_understood" not required.
+		if (pe >= success_boxes) // If they got a good result, adds 10% understanding, up to 100%
+			understanding = clamp((understanding + (max_understanding/10)), 0, max_understanding)
+		else
+			if (pe >= neutral_boxes) // Otherwise if they got a Neutral result, adds 5% understanding up to 100%
+				understanding = clamp((understanding + (max_understanding/20)), 0, max_understanding)
+		if (understanding == max_understanding) // Checks for max understanding after the fact
+			current.gift_chance *= 1.5
 	stored_boxes += pe
 	SSlobotomy_corp.WorkComplete(pe)
 	if(overload_chance > overload_chance_limit)
@@ -149,5 +170,6 @@
 			acquired_chance += user.physiology.repression_success_mod
 	acquired_chance *= user.physiology.work_success_mod
 	acquired_chance += get_attribute_level(user, TEMPERANCE_ATTRIBUTE) / 5 // For a maximum of 26 at 130 temperance
+	acquired_chance += understanding // Adds up to 6-10% [Threat Based] work chance based off works done on it. This simulates Observation Rating which we lack ENTIRELY and as such has inflated the overall failure rate of abnormalities.
 	acquired_chance += overload_chance
 	return clamp(acquired_chance, 0, 100)
--- a/code/game/machinery/computer/abnormality_work.dm
+++ b/code/game/machinery/computer/abnormality_work.dm
@@ -53,6 +53,8 @@
 	dat += "<b><span style='color: [THREAT_TO_COLOR[datum_reference.threat_level]]'>\[[THREAT_TO_NAME[datum_reference.threat_level]]\]</span> [datum_reference.name]</b><br>"
 	if(datum_reference.overload_chance != 0)
 		dat += "<span style='color: [COLOR_VERY_SOFT_YELLOW]'>Current success chance is modified by [datum_reference.overload_chance]%</span><br>"
+	if(datum_reference.understanding != 0)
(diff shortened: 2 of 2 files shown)


EgorDinamit:
Why do you think this is needed? Work chances are fine as is, at least in my opinion, and keep in mind that working here and in LC is very different, so 10 works is practically nothing.
Temperance limit here is higher than in LC as well, so I think this change is not needed.

dsmith328:
> Why do you think this is needed? Work chances are fine as is, at least in my opinion, and keep in mind that working here and in LC is very different, so 10 works is practically nothing. Temperance limit here is higher than in LC as well, so I think this change is not needed.

Sure the temperance LIMIT is higher but you can't REACH THAT at any point except ALEPHS which barely give you ANY stats (I have consistently gotten like, 4s and 5s when getting 27/30 PE from them). There's no limit to level from any class at ALL in LC but there IS here. The work chances we have are also basically copied from LC so they basically assume we're gonna get that "free" +6-10. Also while 10 works is "practically nothing" it's not just 10 works, it's 10 GOOD results, which on some Abnormalities can be genuinely hard to achieve at a reasonable pace. Also there are sometimes where Abnos AREN'T touched so those would remain scary. Outside of that I feel it encourages actively engaging with more than the standard Abnos and can help fix some of the Abnos with... really bad success rates overall. Finally, people have been asking for a way to get EGO Gifts more often and I felt this was a neat way to do it.

dsmith328:
Mind you, with the "more EGO Gifts" thing, I definitely have a plan to implement it so that manager can boost that Globally with an upgrade in the future

code/datums/abnormality/datum/abnormality.dm
```diff
@@ -65,6 +70,8 @@
 	current.toggle_ai(AI_OFF)
 	current.status_flags |= GODMODE
 	current.setDir(EAST)
+	if (fully_understood)
```
EgorDinamit, line 73: You could just check understanding against max_understanding for the same effect.

code/game/machinery/computer/abnormality_work.dm
```diff
@@ -13,6 +13,8 @@
 	var/meltdown_time = 0
 	/// Can the abnormality even meltdown?
 	var/can_meltdown = TRUE
+	/// Has the Abnormality reached max Understanding?
+	var/fully_understood = FALSE
```
EgorDinamit, line 17: Why is it here?

code/game/machinery/computer/abnormality_work.dm
```diff
@@ -53,6 +55,8 @@
 	dat += "<b><span style='color: [THREAT_TO_COLOR[datum_reference.threat_level]]'>\[[THREAT_TO_NAME[datum_reference.threat_level]]\]</span> [datum_reference.name]</b><br>"
 	if(datum_reference.overload_chance != 0)
 		dat += "<span style='color: [COLOR_VERY_SOFT_YELLOW]'>Current success chance is modified by [datum_reference.overload_chance]%</span><br>"
+	if(datum_reference.understanding != 0)
+		dat += "<span style='color: [COLOR_BLUE_LIGHT]'>Current Understanding is: [(datum_reference.understanding/datum_reference.max_understanding)*100]%, granting a [datum_reference.understanding]% Work Success and Speed bonus.</span>"
```
EgorDinamit, line 59: Round the percents so it doesn't show up as 54.33333333%

code/game/machinery/computer/abnormality_work.dm
```diff
@@ -140,6 +144,14 @@
...
 /obj/machinery/computer/abnormality/proc/finish_work(mob/living/carbon/human/user, work_type, pe = 0, max_pe = 0, work_speed = 2 SECONDS, training)
 	working = FALSE
+	if (pe >= datum_reference.success_boxes)
+		datum_reference.understanding = clamp((datum_reference.understanding + (datum_reference.max_understanding/10)), 0, datum_reference.max_understanding)
+	else
+		if (pe >= datum_reference.neutral_boxes)
+			datum_reference.understanding = clamp((datum_reference.understanding + (datum_reference.max_understanding/20)), 0, datum_reference.max_understanding)
+	if (!datum_reference.fully_understood && datum_reference.understanding == datum_reference.max_understanding)
+		datum_reference.fully_understood = TRUE
+		datum_reference.current.gift_chance *= 1.5
```
EgorDinamit, line 154: I think this whole section should be moved to the datum reference.

code/game/machinery/computer/abnormality_work.dm
dsmith328: The purpose of this was to make sure I don't hit the "gift chance multiplier" multiple times

EgorDinamit: This one is in the console variables and is unused.

dsmith328: I think I moved around it and have removed it

code/datums/abnormality/datum/abnormality.dm
dsmith328: Am doing so now

dsmith328:
Gonna test some stuff before pushing and squashing the commit

dsmith328:
image
Seems to work after moving stuff, will commit and squash

code/game/machinery/computer/abnormality_work.dm
dsmith328: moved

code/datums/abnormality/datum/abnormality.dm
```diff
@@ -65,6 +68,8 @@
 	current.toggle_ai(AI_OFF)
 	current.status_flags |= GODMODE
 	current.setDir(EAST)
+	if (understanding == max_understanding)
```
EgorDinamit, line 71: It is recommended to use >= instead of == here, just in case something does go wrong(i.e. it registers 2 works instead of 1, for any reason).

code/game/machinery/computer/abnormality_work.dm
```diff
@@ -53,6 +53,8 @@
 	dat += "<b><span style='color: [THREAT_TO_COLOR[datum_reference.threat_level]]'>\[[THREAT_TO_NAME[datum_reference.threat_level]]\]</span> [datum_reference.name]</b><br>"
 	if(datum_reference.overload_chance != 0)
 		dat += "<span style='color: [COLOR_VERY_SOFT_YELLOW]'>Current success chance is modified by [datum_reference.overload_chance]%</span><br>"
+	if(datum_reference.understanding != 0)
```
EgorDinamit, line 56: Same here, it'd be a good idea to use >

code/datums/abnormality/datum/abnormality.dm
dsmith328: The works are always Clamped though, they physically can't go above max_understanding.

code/game/machinery/computer/abnormality_work.dm
dsmith328: Again, physically can't be below 0 or above max understanding as they are clamped between 0 and max_understanding

dsmith328: `understanding = clamp((understanding + (max_understanding/20)), 0, max_understanding)`

dsmith328: This also specifically has it so that if the Abno hasn't been worked on, the menu doesn't show at all, just like Overload chance

code/datums/abnormality/datum/abnormality.dm
dsmith328: If there was any chance they could be above max_understanding I would, but the only way that's possible is direct admin-bussing to set it to above max_understanding

PR #268, opened 2022-09-04
Fear Update

dsmith328:
## About The Pull Request
Okay so I need to really get into what this is an why it is why it is.
A major fucking annoyance for players is those who want to play Clerk and be able to help out, but one of the biggest issues is stuff like QoH teleporting onto them and just instantly panicking them, so they can do NOTHING to prevent their death. Which is the entire issue, the feeling of "not being able to prevent your death." What I've done with this is set it so that people take sanity damage as they did in base LC. 10%, 30%, 60%, and then 99% (or 200% if they have a new variable set to TRUE) based off respective levels. I want to see how this works for people and how they feel about it. WhiteNight also showcases his own enhanced fear-level system where he deals more fear damage than normal. I've also made it so that when people are hit by fear, they say one of a few lines depending on the level so others know that they've been hit by fear. This is EXTREMELY useful because it could indicate to others that "OH SHIT, SOMETHING TERRIFYING IS AROUND THAT CORNER". Though another thing I've done with this reduces the minimum level so that Clerks are TRULY level 0. So a TETH or HE is gonna be SIGNIFICANTLY scarier to them than an Agent. One of the main reasons I'm making this is because of CENSORED and I feel like this is a path he could REALLY benefit from as an Abnormality, because right now we do sanity damage in intervals of 26% for some reason? This means he's doing 16% more max sanity damage than he's supposed to.
Part of this change is also giving QoH a "Teleportation Marker" that arrives before she does, so that people don't get ABSOLUTELY destroyed by something they can't control, but will instead be punished for not paying attention. This turns it into a skill issue, instead of a random damage moment.

## Why It's Good For The Game
Should ideally reduce the overall annoyance of abnos that use fear as a "BOO, SCARY!" mechanic while making it so that Abnos that use fear damage to KILL people still get their head-start. Like, TSO won't have a clerk INSTANT panic but the moment they get hit by the song they're fucked. The same occurs with Blue Star. Or, hell, maybe the Clerk gets the chance to mutter its location before being sucked into the void.

## Changelog
🆑
add: "panic_bypass" variable that lets any Abnormality be able to bypass the cap on sanity damage being 99% from a single source.
add: WhiteNight unique fear levels/responses
add: CENSORED unique fear responses
add: "Audible" (The character uses "say") lines when someone encounters a fear source for the first time.
add: An Indicator for QoH's arrival, before she shows up.
balance: Changed Fear damage to be based on the LC fear level damages, so 10%/30%/60%/100%, but 100% was replaced with 99% for most cases.
/🆑


Kitsunemitsu (approved):
Looks good, actually.
With balancing, it looks to be good enough.
Code looks to work.

code/game/machinery/computer/abnormality_work.dm
```diff
@@ -104,7 +104,32 @@
 
 /obj/machinery/computer/abnormality/proc/start_work(mob/living/carbon/human/user, work_type, training = FALSE)
 	var/sanity_result = round(datum_reference.current.fear_level - get_user_level(user))
-	var/sanity_damage = -(max(((user.maxSanity * 0.26) * (sanity_result)), 0))
+	var/sanity_damage
+	if (istype(datum_reference.current, /mob/living/simple_animal/hostile/abnormality/white_night)) // I wanna check for WN, and if so give him a custom work fear rating. Ideally I'd get a console subtype but I'm unsure how to properly pre-set it to be a WN console for WN.
```
EgorDinamit, line 108: I feel like this and CENSORED checks should be generalized.

code/game/machinery/computer/abnormality_work.dm
```diff
@@ -104,7 +104,32 @@
...
+				sanity_result = 5
+	if(istype(datum_reference.current, /mob/living/simple_animal/hostile/abnormality/censored))
+		switch(round(get_user_level(user)))
+			if(4 to INFINITY)
+				sanity_result = 3
+			else
+				sanity_result = 5
+	switch(sanity_result)
+		if(-INFINITY to 0)
+			sanity_damage = 0
```
EgorDinamit, line 124: This here is pointless.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -23,6 +23,7 @@
 	. = ..()
 
 	RegisterSignal(src, COMSIG_COMPONENT_CLEAN_FACE_ACT, .proc/clean_face)
+	RegisterSignal(src, COMSIG_HUMAN_INSANE, .proc/Spread_Panic)
```
EgorDinamit, line 26: There is no need to do this, just call that proc in death()

code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
```diff
@@ -110,21 +111,43 @@
...
 				to_chat(H, "<span class='danger'>Not that thing...")
-			if(3 to INFINITY)
+			if(3)
+				sanity_damage = -(H.maxSanity*0.6)
+				H.apply_status_effect(/datum/status_effect/panicked_lvl_3)
+				if(H.sanity_lost)
+					continue
 				to_chat(H, "<span class='userdanger'>I'm not ready for this!")
+			else
+				if(panic_bypass) // Should certain things be able to drive people insane instantly by themselves? Most definitely. Should most things? ehh... It's not exactly fun to turn a corner and lose control because of it. Now if you get hit afterwards or see multiple? Shit out of luck.
```
EgorDinamit, line 142: No. Just generalize the panic effect depending on fear_level.

code/modules/mob/living/simple_animal/abnormality/waw/hatred_queen.dm
```diff
@@ -123,6 +123,11 @@
 		return FALSE
 	return ..()
 
+/mob/living/simple_animal/hostile/abnormality/hatred_queen/CanAttack(atom/the_target)
```
EgorDinamit, line 126: This here is pointless because AttackingTarget is overriden.

EgorDinamit (changes requested):
1. It doesn't have to be a status effect, specially with duration as low as 20.
1.2. For panicked types, you could instead add it to the ai controller itself.
2. Use camel case for procs.

code/modules/mob/living/carbon/human/human.dm
dsmith328: You do realize that is for spreading panic when someone else panics, right? Which is a function of the base game and is what gets replicated here?

dsmith328:
> 1. It doesn't have to be a status effect, specially with duration as low as 20.
>    1.2. For panicked types, you could instead add it to the ai controller itself.
> 2. Use camel case for procs.

The status effect is to make sure the overlays don't stick around in case something stupid like the person gets gibbed or otherwise dies occurs. It's a safety net if anything.

code/game/machinery/computer/abnormality_work.dm
dsmith328: Best way I can think of would be setting a list of "reactions" to the abno based on a list, I.e. 1 level difference is X level fear reaction, etc. and having WN and CENSORED overwrite that

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1219,6 +1220,78 @@
...
+		var/sanity_damage
+		switch(sanity_result)
+			if(-INFINITY to 0)
+				continue
+			if(1)
+				sanity_damage = -(H.maxSanity*0.1)
+				H.apply_status_effect(/datum/status_effect/panicked_lvl_1)
+				if(H.sanity_lost)
+					continue
+				to_chat(H, "<span class='warning'>Here it comes!")
```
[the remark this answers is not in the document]
dsmith328: For when someone sees an Abno? Yeah? Unless you're referring to the EO, RO, and Manager, which that's a placeholder at the moment. Otherwise I thought have a text indication along side the visual is just nice to have.

code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
dsmith328: ...it is? This is literally the make 99% panic thing that was talked about so Clerks and Low-Level agents don't instant panic to a single source from max sanity UNLESS it's one of a few specific sources?

dsmith328: This is only even checked if a person is 4+ levels lower than the fear source.

code/modules/mob/living/carbon/human/human.dm
dsmith328: I think I understand what you mean, and that's to remove the code in death(gibbed) and just have death(gibbed) call Spread_Panic() which I've done

code/game/machinery/computer/abnormality_work.dm
dsmith328: and I've realized that would require overwriting it for each threat level... hmm. I feel like that might just be less efficient than this.

dsmith328:
This compiles on my end? So I have no idea what the issue is on the latest commit, as it seems to be a MAPPING ISSUE???

dsmith328:
> 1. It doesn't have to be a status effect, specially with duration as low as 20.
>    1.2. For panicked types, you could instead add it to the ai controller itself.
> 2. Use camel case for procs.

If has occurred to me that if I added it to the AI Controller... would QDEL_NULL properly remove the icon?

dsmith328:
Not squashing this yet because I want your thoughts on the current commit

EgorDinamit (commented):
Still not generalized... Ugh, I'll probably show what I meant later.

PR #270, opened 2022-09-05
Finally destroy Pbird

SarmentiCampbell:
## About The Pull Request
we done it

## Changelog
:cl:
balance: break punishing bird's wings when he start retaliating.
/:cl:


Kitsunemitsu:
Make it a more substantial nerf. He shouldn't full heal on kill.

dsmith328:
There's not really a reason for this. If you're actively suppressing Punishing Bird, then you're supposed to be slapped for it. Was this a good idea BEFORE he killed himself? Yes. Should we balance around Admemes? No. Sentient Punishing Bird is NOT the standard, it is the exception.

PR #281, opened 2022-09-06
Blue smocked shepherd can now lie

Caldony:
## About The Pull Request
Whenever a work on shepherd is finished, he will have a 30% chance of lying and 70% chance of saying the truth, as of now shepherd is only capable of mentioning people's death/life or if an abnormality is breaching/contained
originally I wanted to make it a 50% chance of lying, however him saying the truth _most_ of the time means the lies are more likely to be believed.

I haven't coded in a while, so do expect some shitcode and really hacky solutions. But there isn't any runtime and it accounts for most scenarios in which the lies are relevant so there shouldn't be any problems functionally.

ADDITIONALLY there is now a GLOBAL SIGNAL sent whenever an abno spawns, it sends the abno datum as an argument and stuff so other coders can use that for other shenanigans I guess. (I didn't end up using it in the PR so i'll see if I remove it or not, but i'll probably use it in another PR anyway)

## Why It's Good For The Game
A huge part of shepherd personality in WL is the fact that he's constantly mixing truth with lies,  that feature not only gives him some personality but it can also make people a little bit more paranoid as to what has and hasn't breached or who is and isn't dead, it's mostly flavour though.

## Changelog
:cl:
add: Shepherd is now capable of lying 
/:cl:


code/__DEFINES/dcs/signals.dm
```diff
@@ -31,6 +31,8 @@
 #define COMSIG_GLOB_HUMAN_INSANE "!human_insane"
 /// a lobotomy_corp subsystem meltdown occured
 #define COMSIG_GLOB_MELTDOWN_START "!meltdown_started"
+/// a new abnormality has spawned with their room (/datum/abnormality)
+#define COMSIG_GLOB_ABNORMALITY_SPAWN "!abno_datum"
```
EgorDinamit, line 35:
```suggestion
#define COMSIG_GLOB_ABNORMALITY_SPAWN "!abno_spawned"
```
abno_datum is rather confusing for a signal name.

code/modules/mob/living/simple_animal/abnormality/he/blue_shepherd.dm
```diff
@@ -52,19 +52,65 @@
...
+	var/list/abno_breach_lines = list(
+				" has breached, I could help you know?",
+				" is out, are you sure you're strong enough to take care of it by yourself?",
+				" is going on a rampage, you guys really can't do your job right huh?",
+				" has breached and you're still wasting your time on me? I'm flattered.",
+				)
+	//lines shepperd say when an abno hasn't breached (yet)
+	var/list/abno_safe_lines = list(
+				" is still stuck in their cell like me, but freedom isn't something you can just take away so easily.",
+				"'s hasn't breached yet, but I wouldn't count on it staying that way.",
```
EgorDinamit, line 89:
```suggestion
				" hasn't breached yet, but I wouldn't count on it staying that way.",
```

EgorDinamit (changes requested):
1. It'd be a better idea to not create an entirely new list that does same thing as existing ones. Just use global lists.
2. Procs should be using camel case, i.e. "OnNewCrew()" instead of "on_new_crew". It is the opposite for variables.

Caldony:
I did most of the change asked with a slight difference
1. I didn't end up using the custom abno list as I could in fact just use the global datum list at each work, that does mean there is now no use for the abno spawn global signaller, I can delete it if you want since it currently serves no purpose but I still think it's a useful signaller to have around
2. I kept the people list but slightly changed how it work, while I CAN use the global living list, I realised that people respawn very often which means there's inevitably going to be duplicate names, which means the list will probably be bloated by dead versions of someone, so whenever a new crewmember that has the same name of someone already in the list joins it gets deleted to avoid useless duplicate, while still acknowledging dead people as long as they haven't already respawned

EgorDinamit:
There is a global list of players which excludes dead people and mobs with no clients.

Caldony:
that has a different issue in that if I search this list whenever a work is complete, then it won't include dead people, which defeats the point

dsmith328:
I've seen this in practice and it's honestly great having people double check over radio what abnos have breached when Blue Smocked talks some shit just after a meltdown.

Caldony:
fuck I messed up the commit squash hold on fuck

PR #283, opened 2022-09-07
Censor EGO nerf

TemperanceTempy:
## About The Pull Request

Nerfs Censors damage resist and fixes reduction applying while not equipped in the hand.

## Why It's Good For The Game

local crackhead makes 3 censored ego and starts facetanking multiple pale apostles. That probably shouldn't happen.
## Changelog
:cl:
balance: reduced damage reduction from 40% to 10% as this applies alongside armor calculation
fix: fixed damage reduction still applying when weapon was moved from the hand to other inventory slots.
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/aleph.dm
+++ b/code/game/objects/items/ego_weapons/aleph.dm
@@ -339,8 +339,7 @@
 	name = "CENSORED"
 	desc = "(CENSORED) has the ability to (CENSORED), but this is a horrendous sight for those watching. \
 			Looking at the E.G.O for more than 3 seconds will make you sick."
-	special = "This weapon increases its user resistance to all damage sources by 40% while equipped. \
-			Using it in hand will activate its special ability. To perform this attack - click on a distant target."
+	special = "Using it in hand will activate its special ability. To perform this attack - click on a distant target."
 	icon_state = "censored"
 	force = 75
 	damtype = BLACK_DAMAGE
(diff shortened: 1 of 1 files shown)


EgorDinamit:
Nuh-uh, it is a 250 PE weapon from a highly dangerous abnormality. It deserves to be strong.
If it, for some reason, allows users to stack it more than twice(humans have two hands), then it's a bug that should be fixed instead.

Kitsunemitsu:
Egor, Censored is an issue as a weapon.
You should NOT be able to get 80% resistances across everything because you "worked on a kinda hard abnormality"

It should have it's resistances removed entirely, and the damage on the IFF special attack reduced to 120.
It should NOT be objectively better than Paradise lost and Twilight.

EgorDinamit (changes requested):
Remove armor buff. Nothing else is to be changed.

dsmith328:
Removed the "Do Not Merge" tag because we've literally talked about adding this exact change into the game and confirmed it should be added.

code/game/objects/items/ego_weapons/aleph.dm
```diff
@@ -407,21 +406,3 @@
...
-	H.physiology.damage_resistance += 40
-
-/obj/item/ego_weapon/censored/dropped(mob/user, silent = FALSE)
-	. = ..()
-	if(!ishuman(user))
-		return
-	var/mob/living/carbon/human/H = user
-	H.physiology.damage_resistance -= 40
-
-/obj/item/ego_weapon/censored/get_clamped_volume()
```
EgorDinamit, line 426: Bring it back, this makes the attack sound less loud.

EgorDinamit:
> Removed the "Do Not Merge" tag because we've literally talked about adding this exact change into the game and confirmed it should be added.

For future reference, I'd recommend against removing it since it's specifically for use by head coder(currently me).

killer7luis:
ngl instead of lowering it to 10% how about 20%?

Kitsunemitsu:
Yeah instead of a free immunity ypu get a free executive armor that stacks with other armor :clueless:. The problem is that it has this ability. Free armor just for carrying a weapon that stacks additively to the other stuff.

EgorDinamit (approved):

PR #294, opened 2022-09-09
Fixes WCCA EGO cost

killer7luis:
WCCA EGO had teth cost instead of zayin cost

changes: WCCA EGO cost from 20 to 12 like all the other zayins


EgorDinamit:
In future, make proper PR titles that represent the change being made.

PR #295, opened 2022-09-09
Agents become BatB upon killing her

Caldony:
## About The Pull Request

When someone "kills" beauty their ghost is forcefully yeeted into the newly respawned beauty that comes after.
in order to remember who was killed by beauty.

I had to add a new list system to the abnormality datum so that I could add variables to transfer later. the reason I made it a list even though only one var needs to be transfered is because it'd be nice to have some way for abnormalities to keep some data after death. which may include multiple variables. 

it's a VERY shoddy solution but I couldn't think of anything better. I also added a "PostSpawn" proc to the abnormality datum as some variables aren't accessible on Initialize like the datum_reference

also the player that's "in place" of beauty can only talk, they can't move or do anything other than nuzzle people, so it doesn't actually do anything balance wise. (actually they can pull people that get close but they can't aggro grab them so it's not that bad)

## Why It's Good For The Game

I think giving more personality to abnormalities beyond just "they kill you" is always good and make each interaction with them feel a little more special. agents becoming beauty is canon to the original story and adds a lot of flair to what is a relatively simple abno to work on.

## Changelog
:cl:
tweak: Dying to beauty now turns you into beauty
/:cl:

### DIFF
--- a/code/datums/abnormality/datum/abnormality.dm
+++ b/code/datums/abnormality/datum/abnormality.dm
@@ -44,6 +44,8 @@
 	var/max_understanding = 0
 	/// A list of performed works on it
 	var/list/work_logs = list()
+	///a list of variable the abno wants to remember after death
+	var/list/transferable_var
 
 /datum/abnormality/New(obj/effect/landmark/abnormality_spawn/new_landmark, mob/living/simple_animal/hostile/abnormality/new_type = null)
 	if(!istype(new_landmark))
@@ -96,6 +98,7 @@
 	if (understanding == max_understanding && max_understanding > 0)
 		current.gift_chance *= 1.5
 	overload_chance_limit = overload_chance_amount * 10
+	current.PostSpawn()
 
 /datum/abnormality/proc/FillEgoList()
 	if(!current || !current.ego_list)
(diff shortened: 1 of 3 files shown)


EgorDinamit:
I personally don't think it's a good idea to put players in control of abnormalities.
This will only force people to ghost.

Caldony:
I feel like the main difference is that Beauty can't really move or attack and do anything else than talk, it's also really easy to just not get killed by her.
I 100% expect people to ghost after a while, but that's why I think the mechanic is rather harmless, you don't get anything out of it other than a fun interaction that's almost like an easter egg

EgorDinamit:
Fair enough, I'll leave this one up to the maintainers.

code/datums/abnormality/datum/abnormality.dm
```diff
@@ -96,6 +98,7 @@
 	if (understanding == max_understanding && max_understanding > 0)
 		current.gift_chance *= 1.5
 	overload_chance_limit = overload_chance_amount * 10
+	current.PostSpawn()
```
dsmith328, line 101: oooo i can use this to change the direction of melting love

dsmith328:
The code looks pretty good in all honesty. In my opinion you should add a message telling the person to ghost when they want to leave Beauty and then just squash commits.

dsmith328 (approved):

Kitsunemitsu (approved):
Looks good, I love this. Lore accuracy and playable abnos in a very humerous package. 
Ship it.

dsmith328 (approved):
Lets do this

PR #300, opened 2022-09-11
Adds a bit denoting the change to Dingle-Dangle's success rate at higher levels

Auris456852:
## About The Pull Request

Literally changes the record to say High/Low for Instinct, and Low/High on Insight. As it is.

## Why It's Good For The Game

While agents should not need exact percentages, it would be best if an agent did not end up getting gotted by Dangle because they just reached level 3 through 55 in every stat, then had to handle a meltdown or something without knowing of the success rate difference. 
Such writing can also be seen on Silent Girl's record, so I figured it wouldn't hurt to have here, too.

## Changelog
:cl:
tweak: Adjusted Dingle-Dangle's record to mention the change in success rate on higher ranks.
:cl:

### DIFF
--- a/code/modules/paperwork/records/info/teth.dm
+++ b/code/modules/paperwork/records/info/teth.dm
@@ -125,8 +125,8 @@
 	Work Damage : Moderate	<br>
 	- An employee with Level 3 or higher Prudence was consumed after completing their work.	<br>
 	- When the work result was Bad, the employee was consumed with a normal probability	<br><br>
-	<h4>Instinct:</h4> High<br>
-	<h4>Insight:</h4> Low<br>
+	<h4>Instinct:</h4> High/Low<br>
+	<h4>Insight:</h4> Low/High<br>
 	<h4>Attachment:</h4> Common<br>
 	<h4>Repression:</h4> Common<br>"}
 


Kitsunemitsu:
In general, we avoid doing this, it's unclear in general.
All records info pertains to the level of the abnormality. Dangle is level 2, and it gets the level 2 chances written,
Writing `Very High/High/Low/Low/Low` is very messy, and just `High/Low` isn't clear enough.

dsmith328:
> In general, we avoid doing this, it's unclear in general. All records info pertains to the level of the abnormality. Dangle is level 2, and it gets the level 2 chances written, Writing `Very High/High/Low/Low/Low` is very messy, and just `High/Low` isn't clear enough.

Though I do feel it's a bit unfair for some abnos, as they simple fail you outright without a certain amount of a stat, see General Bee who autofails level 2 Works and lower, iirc.

Kitsunemitsu:
We already have something for that: Abnormality level.
All records denote the chances at their level. If you're outside of their level it's a crapshoot. This is honestly the best compromise between clarity and cleanliness.

Auris456852:
Admittedly, due to how work success rate is finally based off of the corresponding attribute and not the total level, the Insight bit doesn't actually matter. If you're going to do work at the level it becomes high, you're going to die anyway.

But keeping it as is for Instinct can still potentially lead to a lot of deaths. It would be especially unfair, if someone less aware of exact percentages just grinded Fortitude to get everyone White weapons, but then died because whoops, hit 60 Fortitude, now success rate's a mere 30 percent while total level is 1 or 2. Didn't know this? Too bad.

Same if someone with high Fortitude, but just below 60 Prudence went to handle it due to meltdowns, but got BTFO because of it. You might think it to be unlikely, but Scarecrow specifically makes this an option that can happen. It would be ideal if those sorts of fringe cases could be avoided without people like me code-diving.

DiaFRAME444:
Disagree with the PR because its not exactly clear that "High/Low" and "Low/High" means if you just look at the paper. Nobody will guess that it means "High at low level, then low at higher levels" unless they've seen this PR.
On the other hand,

> We already have something for that: Abnormality level. All records denote the chances at their level. If you're outside of their level it's a crapshoot. This is honestly the best compromise between clarity and cleanliness.

I very much doubt this is the best that can be done.

Kitsunemitsu:
Originally we had it as
Common/Common/Low/Low/Very Low
This is messy and made it borderline unreadable.

Auris456852:
I guessed that much on Silent Girl's record without having to look up success rates, but I guess that's just my experience with abnos talking.
What if you just made the success rates a table, though...?

Kitsunemitsu:
Honestly, I prefer this method we have right now.
It gives just the right amount of information.
I know you don't like reading the code, but I don't want to clog up the records, and want to keep it concise.
Going back and making every success a table isn't exactly fair or feasible, and ruins part of the mystique.

Just keep in mind that the records pertain to the level of the abnormality, and any irregularities aren't approved by the vatican.

PR #302, opened 2022-09-13
Silent Girl Null Removal

dsmith328:
## About The Pull Request
If someone's dead for enough time, their body is deleted (or cremated), or they are gibbed, then they can be set to "null" in the list. I did NOT account for this previously so she would be stuck being... well, harmless. This has remedied that and she WILL kill. I've also reduced her panic warning time from 10 seconds to 5.

## Why It's Good For The Game
Fixes a broken abno.

## Changelog
:cl:
add: Null-check for Silent Girl's list
/:cl:


Kitsunemitsu:
Warning wasn't slow enough as is, keep it at 10 seconds.
With 5 seconds you have time to move about 10 squares and open one door

dsmith328:
> Warning wasn't slow enough as is, keep it at 10 seconds. With 5 seconds you have time to move about 10 squares and open one door

Reverted it

PR #307, opened 2022-09-14
CENSORED tweaks

EgorDinamit:
## About The Pull Request

- Increases attribute requirements for CENSORED weapon.
- Increases damage dealt with CENSORED weapon.
- Increases damage dealt by CENSORED itself.

## Why It's Good For The Game

- Slightly more accurate to the game. Makes it a weapon that can only be used by strong agents.
- With armor gone, it's only natural it should receive a buff somewhere else.
- CENSORED doesn't have any real special attacks right now, making it really easy to fight. This should *slightly* change it.

### DIFF
--- a/code/game/objects/items/ego_weapons/aleph.dm
+++ b/code/game/objects/items/ego_weapons/aleph.dm
@@ -341,21 +341,21 @@
 			Looking at the E.G.O for more than 3 seconds will make you sick."
 	special = "Using it in hand will activate its special ability. To perform this attack - click on a distant target."
 	icon_state = "censored"
-	force = 75
+	force = 85
 	damtype = BLACK_DAMAGE
 	armortype = BLACK_DAMAGE
 	attack_verb_continuous = list("attacks")
 	attack_verb_simple = list("attack")
 	hitsound = 'sound/weapons/ego/censored1.ogg'
 	attribute_requirements = list(
-							FORTITUDE_ATTRIBUTE = 80,
-							PRUDENCE_ATTRIBUTE = 100,
-							TEMPERANCE_ATTRIBUTE = 80,
-							JUSTICE_ATTRIBUTE = 80
+							FORTITUDE_ATTRIBUTE = 110,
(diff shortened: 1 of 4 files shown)


code/game/objects/items/ego_weapons/aleph.dm
```diff
@@ -341,21 +341,21 @@
...
 	hitsound = 'sound/weapons/ego/censored1.ogg'
 	attribute_requirements = list(
-							FORTITUDE_ATTRIBUTE = 80,
-							PRUDENCE_ATTRIBUTE = 100,
-							TEMPERANCE_ATTRIBUTE = 80,
-							JUSTICE_ATTRIBUTE = 80
+							FORTITUDE_ATTRIBUTE = 110,
+							PRUDENCE_ATTRIBUTE = 90,
+							TEMPERANCE_ATTRIBUTE = 90,
+							JUSTICE_ATTRIBUTE = 90
```
dsmith328, line 354: Why the increased requirements? This is essentially on par with most of our Aleph weapons at the moment now, is it not?

Kitsunemitsu: This is a strange attribute requirement.

Kitsunemitsu:
Egor, this is honestly too much.
CENSORED is already a pretty great weapon and out-DPSes Smile before the buff, with regular attacks

dsmith328:
> Why buff the CENSORED ego even more? Was the 300 death laser + 40% resistance across the board not enough?

It doesn't have the resistances anymore

dsmith328:
Honestly this could be a normal 100 cost if we get the armor with it's current stats. I don't mind it getting more damage overall and the special. Just makes it a really offensive Aleph weapon with no other benefits. Smile is still really good for the EXECUTE so this doesn't really steal any spotlights.

EgorDinamit:
**If** or **When** we do get the armor, it can be adjusted.
It's a 250 PE weapon from one of the most difficult abnormalities to work on, it **deserves** to be special, just like paradise lost.

PR #308, opened 2022-09-15
Dreaming Current

EgorDinamit:
## About The Pull Request

Adds the funny drug shark. It kills people funnily.

## Why It's Good For The Game

New abnormality! More death!


Kitsunemitsu:
I can add Ego in an incoming update, ASAP.
I have an ego update that I want to do anyways, I'll get it done sunday.

dsmith328:
Oh right. Gonna mention this here: It breached in the farthest left containment unit in Information on Alpha Corp and just DIDNT do ANYTHING. A guy walked in and just shot it to death point blank.

dsmith328:
It needs to make sure it's not targeting allies for it's charge so it won't charge randomly and also it needs a way to make sure it's not trying to charge into itself as it got stuck on QoH earlier.

PR #330, opened 2022-09-28
Fragment + Grant Us Love Nerf

PotatoTomahto:
## About The Pull Request
Fragment posed way too much danger for a TETH. It now stands still when singing. 
Grant us love takes just a little bit longer to start pulsing, and it doesn't drop counters by 2 anymore.

## Why It's Good For The Game
Fragment is a goddamn menace, and it's all because it can sing while still chasing after you. Grant us love should hopefully not shred the entire facility with WL as quickly.

## Changelog
:cl:
balance: fragment and grant us love nerf
:cl:


Kitsunemitsu:
Fragment is the only threat early game, It's best to keep it that way.

EgorDinamit:
> Fragment is the only threat early game, It's best to keep it that way.

I was expecting damage nerf, but it just makes it so it doesn't move while singing. It's fine.

PR #337, opened 2022-09-30
Silent Girl Lowpop Compensation

monochromatism:
## About The Pull Request

What it says on the tin. If Silent Girl has 5 or more candidates for her guilt effect, she operates exactly as before, but if she has 4 or less, she'll give them an extra-long grace period between melting down and driving them insane.

The length of the lowpop grace period, the length of the highpop grace period, and the threshold of what she considers low enough population for a grace period at all, are all variables, for your administrative convenience.

## Why It's Good For The Game

On any map but Alphacorp, ten seconds is too little of a delay for small groups to respond to and coordinate saving someone from Silent Girl's guilt effect.  While you could argue that competent players won't let her melt down, the fact is that when you have low population, you can't handle every meltdown to begin with, and players spreading out to handle as much as possible makes a potential Silent Girl meltdown even more of a death sentence. This should mitigate that.

## Changelog
:cl:
balance: silent girl insanity has a longer grace period on lowpop
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/silent_girl.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/silent_girl.dm
@@ -28,6 +28,9 @@
 
 	var/mob/living/carbon/human/guilty_people = list()
 	var/mutable_appearance/guilt_icon  // Icon for the effect
+	var/insanity_long = 30 SECONDS
+	var/insanity_short = 10 SECONDS
+	var/population_threshold = 5
 
 /mob/living/simple_animal/hostile/abnormality/silent_girl/Initialize(mapload)
 	. = ..()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/abnormality/he/silent_girl.dm
```diff
@@ -89,24 +92,29 @@
...
 
 /mob/living/simple_animal/hostile/abnormality/silent_girl/zero_qliphoth(mob/living/carbon/human/user)
+	var/insanity_timer = insanity_short
 	var/dead_list = guilty_people
+	var/list/potential_guilt = list()
 	for(var/mob/living/carbon/human/dead_body in dead_list)
 		if(dead_body.stat == DEAD || isnull(dead_body))
 			guilty_people -= dead_body
 	if (!LAZYLEN(guilty_people)) // No Guilty on 0 counter? Find a random person and take them <3
-		var/list/potential_guilt = list()
```
dsmith328, line 97: This doesn't need to be moved and I'd like to avoid declaring variables when they may not be used. It's literally my only issue with this.

dsmith328 (changes requested):

code/modules/mob/living/simple_animal/abnormality/he/silent_girl.dm
monochromatism: Fixed in latest commit.

dsmith328 (approved):
Aight let's do this

PR #345, opened 2022-10-06
Bodybag Nerfed

Kitsunemitsu:
## About The Pull Request
Bodybags can only hold 2 items, and slow you down while dragging them.

## Why It's Good For The Game
I'm kinda sick of people just dragging a bodybag full of ego gear into battle.

## Changelog
:cl:
balance: Bodybags slow you down while you drag them.
balance: Bodybags can now hold 3 bodies and 2 items, to stop people from gaming with them
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/bodybag.dm
+++ b/code/game/objects/structures/crates_lockers/closets/bodybag.dm
@@ -5,7 +5,7 @@
 	icon = 'icons/obj/bodybag.dmi'
 	icon_state = "bodybag"
 	density = FALSE
-	mob_storage_capacity = 2
+	mob_storage_capacity = 3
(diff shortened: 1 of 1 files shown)


EgorDinamit:
I think adding slowdown is more than enough to solve the problem behind it.

Kitsunemitsu:
> I think adding slowdown is more than enough to solve the problem behind it.

I just want to be sure that only clerks use it.

EgorDinamit:
> I just want to be 100% sure that only clerks use it.

Clerks also use it to transport trash or meat from gibbed ordeals, so that's counter-intuitive.
Agents cannot sacrifice speed, so slowdown will handle it.

dsmith328 (approved):
Seems good on my side

PR #347, opened 2022-10-09
Blue Star & Queen Bee buff

Kitsunemitsu:
Update bee_spawn.dm

## About The Pull Request
Doubles Queen Bee Damage
Blue star has double damage but hits 1.5x speed

## Why It's Good For The Game
So, Queen Bee and Blue Star were at the time motherfuckers to deal with, but now they're kinda mid. 
Honestly, it's entirely because their numbers are too small for Queen bee, everyone can heal it without any need for armor.
Now we have armor and a better way of healing in L&H.

Blue Star just simply does not do enough damage.
It should be as scary, if not scarier than other alephs, but as it stands now it's a joke

## Changelog
:cl:
balance: Beesease does more damage.
balance: Blue Star does more damage.
/:cl:

### DIFF
--- a/code/datums/diseases/bee_spawn.dm
+++ b/code/datums/diseases/bee_spawn.dm
@@ -22,7 +22,7 @@
 		return
 
 	var/mob/living/carbon/human/H = affected_mob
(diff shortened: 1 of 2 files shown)


dsmith328 (approved):
This has been TM'd for a bit and I've seen no complaints. Let's get these guys spookier.

PR #353, opened 2022-10-14
working on a meltdown now gives pity attributes

Caldony:
## About The Pull Request

Working on an abnormality who is melting down will give you 2 of an attribute if your level is already too high to get anything from them.

## Why It's Good For The Game

meltdowns tend to get ignored by anyone of a high enough level because "they can just beat them up later" and while that can certainly work, taking care of meltdowns should be encouraged by everyone especially if we want meltdowns to take away total PE later on and the cargo system gets implemented.

this also allows people that are already maxed out with no abnormality of their level to work on an alternative to go slightly beyond what they normally can at the moment.

## Changelog
:cl:
tweak: Working on meltdowns now give "pity stats" if your level is too high
/:cl:

### DIFF
--- a/code/datums/abnormality/datum/abnormality.dm
+++ b/code/datums/abnormality/datum/abnormality.dm
@@ -116,7 +116,7 @@
 	return TRUE
 
 
-/datum/abnormality/proc/work_complete(mob/living/carbon/human/user, work_type, pe, work_time)
+/datum/abnormality/proc/work_complete(mob/living/carbon/human/user, work_type, pe, work_time, was_melting)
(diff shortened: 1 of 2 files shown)


Kitsunemitsu (approved):
Balance-wise, I approve.

dsmith328:
We do actually have an issue with some facilities just getting, say, 1 or 2 alephs and none of them liking Fort so this WOULD help fix that. I've written this pre-code review and am now checking the code.

dsmith328 (approved):
So it's not 2 additional, it's 2 if it would've been 0 and was melting down. Nice.

Kitsunemitsu:
Mergine

PR #361, opened 2022-10-17
Green Noon now has some spread on their gun

Kitsunemitsu:
Update smg.dm

## About The Pull Request
Adds 25 degrees of spread on their gun

## Why It's Good For The Game
They used to be pinpoint accurate, now this makes them less so.

## Changelog
:cl:
balance: Green noon guns are no longer pinpoint accurate
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets/smg.dm
+++ b/code/modules/projectiles/projectile/bullets/smg.dm
@@ -41,4 +41,5 @@
 /obj/projectile/bullet/c9x19mm
 	name = "9x19mm bullet"
 	damage = 10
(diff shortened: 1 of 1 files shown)


dsmith328 (approved):
I would like to test-merge this and see how it affects the difficulties of Green Noon and Dusks

EgorDinamit (changes requested):
The spread is too high, forcing them to miss at even 3-4 tiles range.
Change it to 10-15.

EgorDinamit (approved):
Squash commits and this is good to go.

PR #364, opened 2022-10-17
Dangle work chance increase

Kitsunemitsu:
## About The Pull Request
Increases work chances on dangle

## Why It's Good For The Game
For a teth, they were actually very low.

## Changelog
:cl:
balance: Dangle has increased workchances
/:cl:


dsmith328 (approved):
Brings it in line with Fragment chances. Still scary overall but not as bad as it once was.

PR #366, opened 2022-10-18
adds porccubus

Caldony:
## About The Pull Request

Adds Porccubus, with a twist.

Porccubus keeps most of its gimmick from LC intact, which is to say : immune to range on breach, same way to increase and decrease its counter ect

however, when you work on it when lower than 80 temperance on a good work, you now get drugged instead of dying.
the drug gives you a massive fuck off buff of +30 in all attributes, and regenerate your sanity over time, it can even stack if you decide to drug yourself again

the catch is that it's temporary, every X second, you will lose 1 in all buffed attributes and sanity regen
and it keeps going down, eventually going down in the negative. you can take more of the drug, but the positive effects of it lower at each use until it barely does anything, and the more of the drug you take, the faster the withdrawal will take over.

eventually it will start doing sanity damage and kill you.

one thing of note is that you can only become addicted once per ckey to avoid abuse, that does mean you can technically work on porccubus with lower than 60 temperance safely if you've died to it before. But I'll probably add another form of punishment later

the drug is very much slow acting, it's supposed to feel like a great choice at first, only to realise how terrible everything becomes much later on when the negative effects start picking up.

## Why It's Good For The Game

Porccubus will act as both a weird tool abno that people can use in emergencies, while still acting like a normal abno as his drug mechanic is very easy to ignore if you don't want to get fucked by it. It gives the abnormality different reason to pick it and it's up to the player what they want to do with it.

## Changelog
:cl:
add: Porccubus, it is fighting the war on drugs on the side of drugs
/:cl:


Caldony:
ho you're right it was on good work, yeah I'll make it trigger on good work then, it makes more sense too

Caldony:
I've changed it so that porccubus temperance cap is 80 instead of 60, it's technically more accurate to the original game but also since it only triggers on good work you can technically avoid it with lower temperance if possible

My original reasoning as to why the cap was lower was that I didn't want it to be used too late into the game, but it also made it pretty hard to get the drug at all unless you specifically kept your temperance off like you'd do with woodsman.

Caldony:
porccubus now has 2 tile melee range, while he does a lot of damage and has a decent chunk of health, the fact you can literally just ignore him means he needs to live up to his expectation of environmental threat, while it does technically make it harder to use thrown weapon to cheese him, it doesn't make it harder by much

Caldony:
just gave porccubus the ability to dash to someone every 15 seconds, it doesn't really "fit" the static abno idea I wanted for it, but teleport is way too awkward with the way xeno spawn are usually set up. I also buffed a bunch of damage value so it feels more like a proper threat up close.
it should still stay *mostly* unmoving but now it can actually jump in the middle of hallways more reliably, this does mean you can theorically bait it into a cell for a while without it moving which kind of sucks

Kitsunemitsu:
Tested this both on the server, and on a local.
I feel confident merging this.

PR #369, opened 2022-10-19
Managers console and bullets + Gadgets

InsightfulParasite:
## About The Pull Request
There used to be another PR with the same code but the code got really messed up and deleted icons so now there is a new PR.
Some gadgets for clerks and bullets for clerks to throw.
Ordeal Detector is a proximity detector ONLY for ORDEALS, and Abnormality detector is a proximity detector ONLY for Abnormalities. 
## Why It's Good For The Game
Improves life expectancy of employees. Also now we can make abnormalities more bullshit since survival rate has increased.

## Changelog
:cl:
add: Manager Console for remote shield application
add: abnormality and Ordeal detector for when there is no manager.
add: slowdown gadget and bullet.
/:cl:

image
image


ModularTegustation/tegu_items/lc13unique_items.dm
```diff
@@ -0,0 +1,499 @@
+	//Defective Manager Bullet PLACEHOLDER OR PROTOTYPE SHIELDS
+/obj/item/managerbullet
+	name = "prototype manager bullet"
+	desc = "Welfare put out a notice that they lost a bunch of manager bullets. This must be one of them."
+	icon = 'icons/obj/ammo.dmi'
+	icon_state = "sleeper-live"
+	var/bullettype = 1
```
EgorDinamit, line 7: You should've probably made them all a different item path instead of using this hack.

EgorDinamit: (You actually did...)
Then, this all is not needed. Simply add a proc responsible for bullet effect and modify it for each bullet type.

ModularTegustation/tegu_items/lc13unique_items.dm
```diff
@@ -0,0 +1,499 @@
...
+		return
+	else
+		qdel(src)
+	return
+
+/datum/status_effect/interventionshield/be_replaced()
+	var/mob/living/carbon/human/L = owner
+	switch(respectivedamage)
+		if(RED_DAMAGE)
+			L.physiology.red_mod /= 0.0001
```
EgorDinamit, line 131: As I've said long time ago, we should probably find a way to make it work normally(early return before damage is actually taken by mob) without this mess.

code/game/machinery/computer/manager_camera.dm
```diff
@@ -1,6 +1,29 @@
...
+	var/datum/action/innate/door_bolt/bolt_action
+	var/datum/action/innate/healthheal/hpbullet
+	var/datum/action/innate/sanityheal/spbullet
+	var/datum/action/innate/redshield/rshield
+	var/datum/action/innate/whiteshield/wshield
+	var/datum/action/innate/blackshield/bshield
+	var/datum/action/innate/paleshield/pshield
+	var/datum/action/innate/slowdown/abnoslowdown
+	var/ammo = 0
+	var/maxAmmo = 5
```
EgorDinamit, line 15: All of it. Yes: ALL of it can be reduced to 2 variables: Action to fire bullet, and currently loaded bullet type.
Please do so. If you don't exactly understand what I mean, I could do it myself.

ModularTegustation/tegu_items/lc13unique_items.dm
```diff
@@ -0,0 +1,499 @@
...
+	L.adjustStaminaRecovery(-10) //anything with below 10 stamina recovery will continue to lose stamina
+	L.update_stamina()
+	statuseffectvisual = icon('icons/obj/clockwork_objects.dmi', "vanguard")
+	owner.add_overlay(statuseffectvisual)
+
+/datum/status_effect/qliphothoverload/on_remove()
+	var/mob/living/simple_animal/hostile/L = owner
+//	L.remove_movespeed_modifier(QLIPHOTH_SLOWDOWN)
+	L.adjustStaminaLoss(-167, TRUE, TRUE)
+	L.adjustStaminaRecovery(0)
```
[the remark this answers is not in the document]
InsightfulParasite: No i think i did that and it resulted in 30 for some reason.

ModularTegustation/tegu_items/lc13unique_items.dm
InsightfulParasite: How would i hypothetically do this. Cancel a damage taken by mob proc?

ModularTegustation/tegu_items/lc13unique_items.dm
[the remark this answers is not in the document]
InsightfulParasite: I see why i made it like this. When i coded the new stamina recovery alteration proc i made it so stamina recovery = intial stamina recovery altered by the modifier. So without a value its set to initial + 0

code/game/machinery/computer/abnormality_work.dm
```diff
@@ -152,6 +152,10 @@
 		if(do_after(user, work_speed, src))
 			if(!CheckStatus(user))
 				break
+			user.remove_status_effect(/datum/status_effect/interventionshield)
+			user.remove_status_effect(/datum/status_effect/interventionshield/white)
+			user.remove_status_effect(/datum/status_effect/interventionshield/black)
+			user.remove_status_effect(/datum/status_effect/interventionshield/pale)
```
EgorDinamit, line 158: Might want to test it, but I think putting the parent type will handle all the subtypes.

InsightfulParasite: I do remember testing this and it only removing red type when only one is said to be removed. It doesn't seem to remove the parent.

Kitsunemitsu:
Squash commits and I'll merge.
During testing this all worked perfectly.

PR #371, opened 2022-10-19
Adds Flesh Idol

Kitsunemitsu:
## About The Pull Request
Adds Flesh Idol
```
- When work was complete, everyone in the facility was healed.
- When work was completed 4 times, The Qliphoth counter lowered.
- When the Qliphoth counter reached 0, two abnormalities breached containment.	
```

## Changelog
:cl:
add: Added Flesh Idol
/:cl:


code/modules/clothing/suits/ego_gear/waw.dm
```diff
@@ -173,3 +173,11 @@
 	attribute_requirements = list(
 							JUSTICE_ATTRIBUTE = 60
 							)
+
+/obj/item/clothing/suit/armor/ego_gear/heart
+	name = "bleeding heart"
+	desc = "And the prayer shall inevitably end with the eternal despair of its worshiper."
+	icon_state = "heart"
+	armor = list(RED_DAMAGE = 70, WHITE_DAMAGE = 10, BLACK_DAMAGE = 10, PALE_DAMAGE = 50) // 140
```
EgorDinamit, line 181: I don't think it should have such high red.
In LC, it was a tool that dealt *random* damage type, so we could probably interpret it here better?

code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
```diff
@@ -0,0 +1,73 @@
...
+	base_pixel_x = -16
+	work_chances = list(
+		ABNORMALITY_WORK_INSTINCT = 40,
+		ABNORMALITY_WORK_INSIGHT = 40,
+		ABNORMALITY_WORK_ATTACHMENT = 40,
+		ABNORMALITY_WORK_REPRESSION = 40
+			)
+	start_qliphoth = 1
+	work_damage_amount = 10
+	work_damage_type = RED_DAMAGE
```
EgorDinamit, line 19: Same here. Change the damage proc to be random type.

code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
```diff
@@ -0,0 +1,73 @@
...
+		total_abnormalities += A
+
+	while(num_breached <= 2 && run_number <=30)
+		var/mob/living/simple_animal/hostile/abnormality/processing = pick(total_abnormalities)
+		total_abnormalities -= processing
+
+		if(processing.can_breach && (processing.status_flags & GODMODE))
+			processing.breach_effect()
+			num_breached ++
+		run_number++
```
EgorDinamit, line 69: This entire proc looks terrible, seriously.
Use `for` loop instead and look for abnormality datums instead, triggering 0 qliphoth to avoid any potential issues.

code/modules/clothing/suits/ego_gear/waw.dm
Kitsunemitsu: Not that I know aside from keeping 40/30 across the board and that would mess with the weapon. I chose high red to keep synergy with the weapon.

code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
Kitsunemitsu: I converted it to red across the board to keep it consistent. Making it random would make it more difficult on top of the already brutal effect it has

code/modules/clothing/suits/ego_gear/waw.dm
EgorDinamit: Wouldn't it be funny if weapon dealt random damage too?

code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
EgorDinamit: That was basically the point, though. We do not need yet another red damage abnormality, we have a ton of these.

code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
EgorDinamit: You can randomize list order with a proc.

Kitsunemitsu: Which proc?

code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
Kitsunemitsu: I think the point was the healing on top of breaching if you went too far.... the damage type is kinda just a footnote. 
In terms of Red damage, we actually lack a lot of late game red damage, especially during work.
There's:
Nothing There
Mountain of Smiling Bodies (ONLY when it breaches and only it's normal attack)
Queen bee
General Bee
Dreaming Current (ONLY when it breaches)
King of Greed, I guess? 

That's 4 and two halves of abnormalities out of 18. The Majority of WAW+ does black.

code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
EgorDinamit: You can find it in the defines/helpers.
It is also used in work console for information suppression effect.

code/game/objects/items/ego_weapons/waw.dm
```diff
@@ -174,7 +174,9 @@
 	name = "crimson claw"
 	desc = "It's more important to deliver a decisive strike in blind hatred without hesitation than to hold on to insecure courage. "
 	icon_state = "crimsonclaw"
-	force = 32
+	special = "This weapon hits faster than usual."
```
Kitsunemitsu, line 177: I had to change this to differentiate it from Bleeding heart.

code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
Kitsunemitsu: Egor, I've spoken with the other devs, I'm NOT making this random

EgorDinamit: There's only one maintainer other than you, and that's Redacted. And as far as I can tell, they too thought of making it random as a good idea.
And development isn't democracy either.

Kitsunemitsu: It was mostly me speaking to the juniors. I dunno, I just think that would make this abnormality unworkable. Entirely removing the preparation is a bad idea.

Kitsunemitsu:
Its _just poor design_ to not know what you're getting with the ego gear too. You're working on his extremely dangerous abno with random damage type and you don't even know if you need the weapon or armor. They may have the exact same stats as a weapon or armor set you're using, for all that investment. Players BARELY stand Red Queen because of how random it is.

PR #372, opened 2022-10-20
[TM] Silence should no longer desync

Kitsunemitsu:
## About The Pull Request
So apparently the stoplag proc that I was told to use breaks if there's a lot of players, and desyncs the sound from the damage.
This should fix that.

## Changelog
:cl:
fix: fixed a bug with Silence Desyncing sound and damage
/:cl:


EgorDinamit:
Handled in #379

PR #374, opened 2022-10-21
Roguelite manager cards

EgorDinamit:
## About The Pull Request

- Adds managerial interference console(name subject to changes). It allows manager(or anyone) to select and activate "cards", which are basically small(or not) round modifiers.
- Cards are planned to be of 5 tiers, first being available at round-start and the others after each subsequent ordeal level(T2 after dawn, T3 after noon and so on).

## Why It's Good For The Game

- Adds more uncertainty in the game, which is good given its nature. Each round will(hopefully) feel different, if only just a bit. Additionally, it also increases the amount of manager gameplay functions so they don't die from boredom.


Kitsunemitsu:
This PR has not been worked on in months, I was given permission to close. 
If you want to reopen it, please do

PR #387, opened 2022-10-25
Adds Child of the Galaxy

Kitsunemitsu:
## About The Pull Request
```
	- Successful work completion will cause a healing Token of Friendship to be bestowed upon the employee. The amount of healing will vary in relation to the number of people holding a token.    <br>
	- The Qliphoth Counter increased when NAME completed their work with the Token    <br>
	- When an employee with Child of the Galaxy’s Token of Friendship died, the Qliphoth Counter lowered by 4.    <br>
	- When the Qliphoth Counter reached 0, all employees with the Token took massive black damage    <br>
```

## Changelog
:cl:
add: Added Child of the Galaxy
/:cl:


dsmith328 (approved):
Seems good to me. Biggest concern is that the debuff (the damage) for what is essentially free healing may not be enough but I feel like that's something that'll just need tweaking later and isn't major enough to hold up this PR.

PR #379, opened 2022-10-22
Abnormality balance and tweaks

EgorDinamit:
## About The Pull Request

**More important changes:**
- Mountain of smiling bodies received unique patrol targeting. It will navigate towards corpses or wounded humans first, if it can find those.
- Mountain of Smiling Bodies can no longer be cheesed with its own weapon. It will still get gibbed, but only on first stage.
- Blue Star and Silent Orchestra were buffed, as they were severely underperforming on breach.
- CENSORED and its minions are moving faster, so as to compensate for it being generic melee abnormality with no way of damaging outside of this range.
- Punishing Bird may now breach whenever someone goes insane or works on any other abnormality. Chances for that are lower if you are working on high level ones.
- Punishing Bird also received unique patrol targetting, but it is merely beneficial as it will now attempt to bring insane people back to sanity.

**Minor changes:**
- Abnormality patrolling movement speed is increased to be closer to their normal speed.
- Price of Silence sounds improved/updated. Additionally, it will only play them on its z level.
- Price of Silence also has an ambience now.
- Nothing There normal melee damage increased by 10.
- Dreaming Current will now show its path before charging. Additionally, after-move smoke is in random colors.

**Fixes/Improvements:**
- Queen of Hatred arcana beats sound will play when it's firing, not before.
- Cases of using abnormality's name for conditions replaced with abnormality's path/type.
- All-Around Helper will now try to perform the charge even in melee.

## Why It's Good For The Game

- MoSB needed this behavior for a long time. This will finally make it a proper threat deserving of ALEPH level.
- MoSB hammer issue was just a bug, it should not outright skip all phases for mobs such as MoSB.
- TSO and BS were made before regenerators and all the new content, so nowadays they are way too weak. This changes it.
- CENSORED can be effectively kited for eternity, as it has practically no way of reaching agents with even 0 justice. This is still the case, but less apparent.
- LC accurate and makes Punishing Bird more active.
- Same here, but also gives some benefit to having it in the facility other than summoning apocalypse bird.

- Abnormalities will get to their destination *slightly* faster. That's about it.
- Better sounds are always good. Also fixed the oversight with it playing the sound everywhere, despite only hitting one z-level.
- Same here,. More sounds is also good.
- Makes NT's damage less laughable, outside of abilities.
- A long-awaited feature. No more tokyo drifting out of nowhere, you now have a few seconds to dodge. Additionally, it looks fancy.

- Fix.
- Improvement.
- Improvement.

PR #390, opened 2022-10-26
Scaredy cat now properly teleport to other Oz abnos and use types instead of names

Caldony:
## About The Pull Request

fixes shitcode where scaredy cat used abno names instead of types to detect oz abno breaching, which meant any typo fixing would fuck scaredy cat's ability to discern friends forever. so now scaredy cat should properly prioritize oz abnormalities

## Why It's Good For The Game
cat bug bad

## Changelog
:cl:
fix: scaredy cat will now properly teleport to scarecrow when it breaches
/:cl:


Kitsunemitsu:
Solved in #379

PR #395, opened 2022-10-28
Auto-Apocalypse

dsmith328:
## About The Pull Request
Automates Apocalypse Bird and limits it to once per round. This is done via the new "Lobotomy Events" subsystem which is designed to be able to handle future events as well. This also implements and utilizes the "patrol_to" proc, which accepts a single Turf as a location and sends the target walking to that location.  

While the Birds are walking, they lose density, become immune to damage, and will not target people. Furthermore, the Gate will receive the birds in bulk and output their text one at a time, both giving a buffer before the big bird himself shows up and allowing the text to fully render and not overlap with other instances.

The gate IS vulnerable but is still absurdly healthy and resistant but if it dies the event WILL stop. The event will also be able to occur again later, should the situation allow for it.

Finally this also implements and utilizes a Global list for Abnormality mobs

## Why It's Good For The Game
No longer will admins be forced to spawn the gate! Also, it's frankly fucking terrifying, as this can occur BEFORE people are ready.

## Changelog
:cl:
add: Selectable Patrol
add: Lobotomy Events subsystem
add: Loot Drop to Apocalypse Bird
/:cl:

### DIFF
--- a/ModularTegustation/tegu_mobs/apocalypse_bird.dm
+++ b/ModularTegustation/tegu_mobs/apocalypse_bird.dm
@@ -25,6 +25,11 @@
 	deathmessage = "finally stops moving, falling to the ground."
 	deathsound = 'sound/abnormalities/apocalypse/dead.ogg'
 
+	loot = list(
+		/obj/item/ego_weapon/twilight,
+		/obj/item/clothing/suit/armor/ego_gear/twilight
+		)
+
 	var/list/eggs = list()
 	var/list/egg_types = list(
 						/mob/living/simple_animal/apocalypse_egg/beak,
@@ -485,7 +490,17 @@
 	icon_living = "forest_portal"
 	del_on_death = TRUE
 	/// List of birds that entered it. We don't delete/kill them for the sake of abnormality respawn mechanics.
-	var/list/stored_birds = list()
+	var/list/stored_birds = list("spoken" = list(), "unspoken" = list())
+	/// These are the birds.
+	var/list/bird_types = list(
+		/mob/living/simple_animal/hostile/abnormality/punishing_bird,
+		/mob/living/simple_animal/hostile/abnormality/big_bird,
+		/mob/living/simple_animal/hostile/abnormality/judgement_bird
+		)
+	var/force_bird_time
+	/// I wanna guarantee we get all Blurbs spaced and also a little prep delay so...
+	COOLDOWN_DECLARE(speak_bird)
+	COOLDOWN_DECLARE(summon_bird)
 
 /mob/living/simple_animal/forest_portal/CanAttack(atom/the_target)
 	return FALSE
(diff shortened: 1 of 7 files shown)


ModularTegustation/tegu_mobs/apocalypse_bird.dm
```diff
@@ -494,6 +508,22 @@
...
 			flash_color(M, flash_color = "#CCBBBB", flash_time = 50)
 	addtimer(CALLBACK(GLOBAL_PROC, .proc/show_global_blurb, 5 SECONDS, "A long time ago, in a warm and dense forest lived three happy birds.", 25))
+	COOLDOWN_START(src, take_bird, 10 SECONDS)
+	force_bird_time = world.time + 3 MINUTES
+
+/mob/living/simple_animal/forest_portal/Life()
+	if(COOLDOWN_FINISHED(src, take_bird))
+		var/list/range_area = orange(1, src)
+		if(force_bird_time < world.time)
+			range_area = GLOB.mob_list // If they take too long, we shove them in MANUALLY.
```
EgorDinamit, line 518: It would be a better idea to use the list of abnormalities instead.

EgorDinamit: If there isn't one, then use simple mob list or create a new global list for abnormalities.

InsightfulParasite:
> Why make the Birds immortal while walking to the Gate?

Its mirroring how they work in the Lobotomy Corp game. When the portal spawns they are supposed to be immortal and unable to attack since their focused on entering the gate. I think mechanically its a DPS race against the birds slowly walking to the gate. Right now they are admin controlled and the gate is admin spawned.

dsmith328:
Please. Someone review me :(

ModularTegustation/tegu_mobs/apocalypse_bird.dm
```diff
@@ -499,33 +514,101 @@
...
+	if(COOLDOWN_FINISHED(src, speak_bird))
+		SpeakBird()
+	if(COOLDOWN_FINISHED(src, summon_bird))
+		SummonBird()
+		COOLDOWN_START(src, summon_bird, 30 SECONDS) // So they keep trying to move towards the portal, even if temporarily blocked.
 
 /mob/living/simple_animal/forest_portal/Bumped(atom/movable/AM)
 	if(!isliving(AM))
 		return ..()
+	if(!ishostile(AM))
```
EgorDinamit, line 543: Any reasons for this?

ModularTegustation/tegu_mobs/apocalypse_bird.dm
```diff
@@ -499,33 +514,101 @@
...
 
-	var/mob/living/L = AM
+/mob/living/simple_animal/forest_portal/proc/SpeakBird()
+	if(!COOLDOWN_FINISHED(src, speak_bird))
+		return FALSE
 	var/blurb_text
-	if(istype(L, /mob/living/simple_animal/hostile/abnormality/punishing_bird))
+	if(!length(stored_birds["unspoken"]))
+		return FALSE
+	say("Running SpeakBird")
```
EgorDinamit, line 589: Debug leftovers.

code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
```diff
@@ -93,6 +93,14 @@
 		addtimer(CALLBACK (datum_reference, .datum/abnormality/proc/RespawnAbno), 30 SECONDS)
 	..()
 
+/mob/living/simple_animal/hostile/abnormality/add_to_mob_list()
```
EgorDinamit, line 96: It doesn't have to be its own proc.
Just add them to list on Initialize() and remove on Destroy()

dsmith328: This is how literally every mob list is added to and is *built in*. I'm not making a new one here I'm using the existing one that adds them to the global mob list and expanding it for abnormalities.

dsmith328: Also I'm pretty sure these are called at those exact times.

ModularTegustation/tegu_mobs/apocalypse_bird.dm
dsmith328: Thought it may reduce error cases of the ConsumeBird proc trying to use it on non-hostiles and forces normal interaction from non-hostiles guaranteed

code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
dsmith328: Yeah it is. This is literally the best way to do this.

EgorDinamit (approved):
I am assuming this was thoroughly tested.
Merge it yourself if you think it is ready.

dsmith328:
Yes, it's been TM'd for... at least 4 days straight without issue.

PR #397, opened 2022-10-29
Minor insanity buffs

EgorDinamit:
## About The Pull Request

- Most prob() checks moved to cooldowns instead. This includes justice insanity console smashing.
- Justice insanity type moves MUCH faster.
- Wander insanity type will move from one department to another, constantly yelling stuff. It is also much faster.
- Some random changes to fortitude insanity that I didn't exactly understand. It might actually break everything instead. _Jokes aside, it should now use weapons from its inventory._
- Suicide and wander white damage is increased.

## Why It's Good For The Game

- Less RNG where it wasn't needed.
- GAS! GAS! GAS!
- Nyooom. Now it isn't the most useless insanity type.
- Now, it _might_ use weapons for real. Good luck.
- It isn't negligible anymore.

### DIFF
--- a/code/datums/ai/sanity/_sanityloss_controller.dm
+++ b/code/datums/ai/sanity/_sanityloss_controller.dm
@@ -112,13 +112,22 @@
 	var/mob/living/living_pawn = pawn
 
 	if(!locate(/obj/item) in living_pawn.held_items)
-		blackboard[BB_MONKEY_BEST_FORCE_FOUND] = 10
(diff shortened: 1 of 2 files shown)


EgorDinamit:
> Since wander is getting a big buff by increasing speed, it may be too strong when combined with the past buffs to the damage. This may need to be tweaked

Good. Insanity is meant to be as bad as death.

SpaceJohn413:
"minor"

dsmith328 (approved):
I love this PR so much

PR #398, opened 2022-10-29
reworks shepherd and buddy's fight

Caldony:
## About The Pull Request
near completely reworks the way red buddy and shepherd's fight work, this also fixes a lot of bugs related to the two in the process.

-----------------------------------------------------buddy and shepherd fight---------------------------------------------------
-when red buddy and shepherd meet, they will now gain 3 to 4 times as much max health instead of extra resistances
-they will no longer be healed to full health but instead lose health proportionally to what health they had before they met
-shepherd's slash and red buddy's howl will now hurt the other, red buddy's howl is now both triggered by a cooldown AND losing 10% of their max HP
-if shepherd dies first, buddy dies too, if buddy dies first, shepherd gets a lot of movement speed but their damage is immensely nerfed (even lower than his normal form)

-----------------------------------------------------random fixes---------------------------------------------------------------
the fixes were basically necessary to make this fight work, and a lot of those issues came from the use of sleep()
-both slash and red buddy's howl now use timer to avoid some fuckery where they would interrupt the other
-shepherd can no longer (sometimes?) strike you twice in one attack right after using its slash move
-fixes a lot more dragging jank, this time I'm *pretty* sure I've got them all
-shepherd now has an attack sound

## Why It's Good For The Game

shepherd and buddy's fight was basically impossible not just because of the insane resistances but speed and a huge part of dragging jank, even if you could deal with them it wasn't a really interactive fight.

in this the goal is to make you juggle different damage type and make the fight more of a proper gimmick fight. you need to get up close to trigger shepherd's slash which speeds up the fight by a lot, guns are safer but take a lot longer because of their very high health. 

this gives actual options on how to deal with them without making their fight extremely punishing if not near impossible

## Changelog
:cl:
tweak: shepherd and buddy's fight has been reworked
balance: shepherd now has less extra speed and damage when with buddy
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/blue_shepherd.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/blue_shepherd.dm
@@ -58,7 +58,6 @@
 	var/datum/abnormality/buddy //the red buddy datum linked to this shepherd
 	var/mob/living/simple_animal/hostile/abnormality/red_buddy/awakened_buddy //the red buddy shepherd is currently fighting with
 	var/awakened = FALSE //if shepherd has seen red buddy or not
-	var/abuse = FALSE //if shepherd is attacking buddy
 	var/list/people_list = list() //list of people shepperd can mention
 	//lines said during combat
 	var/buddy_hit = FALSE
@@ -180,7 +179,7 @@
 		say(pick(combat_lines))
 		slashing = TRUE
 		slash()
-	if(awakened_buddy && !abuse)
+	if(awakened_buddy)
 		awakened_buddy.GiveTarget(target)
 	..()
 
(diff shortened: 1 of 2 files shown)


Kitsunemitsu:
They really shouldn't be fighting each other

Caldony:
I feel like it fits more the theme of shepherd and buddy, since especially in the WL novel shepherd clearly showed neglect and abuse towards buddy, but buddy mentions multiple time how abusive he is. I could take that feature off but I also think it takes a lot of the charm of the boss fight too as it requires juggling both their health bar too

Caldony:
> It's bad to have them fight each other because it severely neuters their difficulty when they reach that point. Just look at how QoH used to obliterate abnorms fighting them. Also a dog and person fighting is red and wolf's gimmick. Having them hurt each other could be fine, or maybe just Shephard hurting buddy.

I mean yeah the entire point is that bringing both of them at half health is essentially a win condition, it's like only hitting helper when it dashes, it's a tactic to force people to handle both of them at once instead of only focusing shepherd. I can remove it but I'd still need to nerf their health or increase the damage they do to each other.

PR #401, opened 2022-10-31
Adds Yang, the Escort mission

Kitsunemitsu:
## About The Pull Request
Adds Yang, Reworked from Lobotomy Corporation.

Instead of a tool, Yang is now a breaching fish!
It's mostly helpful, for now, but it has a few caveats.

First is that it's not very strong, at only 800 health, as much as a teth
Second is that it's slow. One of the slowest abnormalities at 7 speed/move Delay.
Third is that it blows up on death, dealing massive white AOE to everything in range.

So in short... It's an abnormality you have to protect or suffer the consequences. Something different.

## Why It's Good For The Game
Lots of bosses we have you just kill, but..... what if we made it something you have to protect?

## Changelog
:cl:
add: Added Yang
/:cl:


EgorDinamit:
`code/modules/paperwork/records/info/waw.dm:  
space indentation detected`

PR #408, opened 2022-11-02
fix porccubus head explosion not accounting for directions

Caldony:
## About The Pull Request

porccubus head overlay thing always faced south which looked weird
the fix is basically making an extra argument (again) that allows us to get every direction for the head overlay

this is believe it or not, the simpler solution as the alternative is copying the entirety of the get head code and changing 2 lines.

from what i've tested, this does not impact the regular decapitated head icon item at all

## Why It's Good For The Game

I hate visual contents I hate visual contents

## Changelog
:cl:
fix: Porccubus's head explosion now face the proper way
/:cl:


EgorDinamit (approved):
I hope you tested it locally.

Caldony:
I did test it locally and decapitated multiple heads to make sure it didn't fuck with anything, no runtimes as far as I could tell since the changes just go to the default state if you don't add the extra arguments

PR #409, opened 2022-11-03
Express Train adjustments

monochromatism:
## About The Pull Request

Gives the Express Train to Hell some requested buffs. 

- Fires at department centers rather than aiming for the average location of all humans.
- Flings hit targets away on hit, and removes some limbs when killing a human.
- Moves twice as fast across the map.
- Spawns smoke as it travels.
- Adjusts the length of the train's repeated noises and makes them play more frequently.

## Why It's Good For The Game

Makes train stronger, better looking, and a bit spicier when you get hit by it.

## Changelog
:cl:
balance: train is now harder
/:cl:


code/modules/mob/living/simple_animal/abnormality/waw/express_train.dm
```diff
@@ -171,6 +155,7 @@
 				src.segments -= seg
 			else
 				seg.forceMove(get_step(seg, seg.dir))
+				seg.forceMove(get_step(seg, seg.dir))
```
EgorDinamit, line 158: Can't you just change delay on the timer above to 0.5?

monochromatism: ... The timer is one tick, is that not the smallest possible unit of time??

PR #411, opened 2022-11-04
Walls and Floors PR 4

Kitsunemitsu:
## About The Pull Request
image
image
image

It's not accurate to wonderlabs... But It'll do.

Floor sprites from TGstation.

## Why It's Good For The Game
Hopefully gives us a new look

## Changelog
:cl:
imageadd: reworks our look.
/:cl:


code/game/turfs/open/floor/misc_floor.dm
```diff
@@ -9,6 +9,10 @@
 	icon_state = "rockvault"
 	floor_tile = /obj/item/stack/tile/plasteel
 
+/turf/open/floor/tile
+	icon_state = "floor_plates"
+	floor_tile = /obj/item/stack/tile/plasteel
```
EgorDinamit, line 14: While you are at it, might want to add a proper stack item for this.
It's not necessary, but would be nice if people could replace floors using it, if they find a crowbar.

code/game/turfs/open/lobotomy.dm
```diff
@@ -0,0 +1,23 @@
+/turf/open/floor/facility
```
EgorDinamit, line 1: Same for all floors there.

EgorDinamit:
Overall, walls look good, although I'd prefer to keep it test-merged for a while to see how others react to it.

code/game/turfs/open/lobotomy.dm
Kitsunemitsu: Just did what all the other plasteel tiles do and default to the plasteel.

Kitsunemitsu:
> Airlocks open very janky, the animation of icons may need to be checked. Animation with paper needs to be changed to match how the airlocks open vertically and the open airlock sprite could be made more raised.

Fixed the animation , paper isn't done per airlock though.

code/game/turfs/open/lobotomy.dm
EgorDinamit: That's a bad idea: Removing floor will result in default tiles being dropped.

code/game/turfs/open/floor/fancy_floor.dm
```diff
@@ -18,7 +18,7 @@
 	tiled_dirt = FALSE
 
 /turf/open/floor/wood/setup_broken_states()
-	return list("wood-broken", "wood-broken2", "wood-broken3", "wood-broken4", "wood-broken5", "wood-broken6", "wood-broken7")
+	return list("wood")
```
EgorDinamit, line 21: Why?

EgorDinamit: Remove that proc altogether, and what it is used for as well.

EgorDinamit: Actually, forget it.

EgorDinamit (approved):

PR #392, opened 2022-10-27
EGO Dump 5 - October 2022

Kitsunemitsu:
### New Weapons
**Assonance** (Yang) - Our first, and likely only hitscan laser rifle
**Unrequited Love** (Piscine Mermaid) - A burst fire white rifle. 
**Feather of Honor** (Firebird) - A sword that fires a magic beam

### Weapon reworks
**Wrist Cutter** - Use in hand to "dodgeroll" (code by para)
This flings you forwards a couple spaces and deals some stamina damage.

**Blind Fury** - Kill something to gain massive damage on your next attack. You must deal the finishing blow
Damage reduced to 41 to keep in line with the other ego.

**Remorse** - Gains two modes. hammer mode and nail mode.
Hit something in nail mode to mark it, hit something in hammer mode to activate it....
Soon we'll make another ego using the old system

**Prank and Hornet** - Melee is removed

### Weapon Tweaks
Oppression gains meter exponentially (up to 60). Damage reduced to 13
Sword Sharpened with Tears can now turn off the combo mode. Combo damage is now comparable, 5x damage for 5x cooldown.

### Weapon Number Changes
Twilight has been nerfed from 160 to 140 damage. 
Smile healing has been reduced to 15% 

## Why It's Good For The Game
Ego Diversification, and a few critical number changes. Twilight was kinda disgusting and so was smile.
Remorse was a nothing weapon, but the ability will be used again in the future.

This PR holds all the EGO for a bunch of upcoming abnormalities so we can TM more than one at the same time.

## Changelog
:cl:
tweak: ego reworks
/:cl:

PR #412, opened 2022-11-04
Edits EGO sounds and updates crimson claw

EgorDinamit:
## About The Pull Request

- Edits sounds for some EGO guns.
- Adds special attack to crimson claw, which you can see below.

## Why It's Good For The Game

- More variety of ***sounds***. I simply didn't like how almost every weapon is using same sound, so here's that.
- So, I was looking through LC sounds and remembered that red-riding hood mercenary has throwing attack. I decided to add something similar to it, because it adds some flavor to it(it's a WAW weapon, after all). To prevent it being death beam 2.0, its damage is reduced per each mob hit, so it can't be really used to clear 999 worms with ease. It also lacks IFF, so be careful.

https://user-images.githubusercontent.com/53223414/199995977-0550f2d5-0116-40b0-85a2-e8568bf3cd7d.mp4


Kitsunemitsu:
Could have notified me while I was working on #392 and I could have made it easier for TM.

EgorDinamit:
> Could have notified me while I was working on #392 and I could have made it easier for TM.

We'll merge 392 first and then work from there.

Kitsunemitsu:
Also, Keep in mind crimson claw is a small weapon, maybe tone the damage back to 80?

EgorDinamit:
> Also, Keep in mind crimson claw is a small weapon, maybe tone the damage back to 80?

Damage is reduced on every hit, and initial is meant to be somewhere around normal melee at max justice.

Kitsunemitsu:
Gotcha, should be good then.

PR #420, opened 2022-11-08
Scaredy Cat loses its target when teleporting

Caldony:
## About The Pull Request

A very simple fix that makes Scaredy cat loses its target when teleporting

## Why It's Good For The Game

Scaredy cat instantly leaving his friend after teleporting looked weird, especially when that target was way too far for him to reach.  so we reset his aggro whenever he comes back to his friend.

## Changelog
:cl:
fix: Scaredy Cat will now lose aggro when he teleports back to his friend
/:cl:


Kitsunemitsu:
Was this tested on a local?

PR #423, opened 2022-11-09
Abnormality queue, qliphoth and ordeal timing changes

EgorDinamit:
## About The Pull Request

- Refactored the way abnormality queue works to allow dynamically changing the time for next abno spawn.
- Additionally, the queue will only fire when round has started(?). This provides a 3 minute window of peace and calm just like in the past, for people to have a roll call and anything the manager/captains might have to say.
- Amount of players has no role in the speed of abnormality spawns, instead it will now slow down the further you are into the game(when you begin getting WAW and ALEPH abnormalities) and speed up at the beginning(to get all the ZAYIN, TETH and HE).
- Qliphoth counter was changed in the opposite direction: It now only cares for the amount of players, instead of a combination of players and amount of abnormalities.
- Ordeals will not show up on the monitor if they are still timelocked(they are simply getting pushed further away).
- Added some cleanup code for when abnormality datum gets destroyed so admins don't have to do it manually in case they want to replace it with something else.

## Why It's Good For The Game

- The game was too slow at the beginning and too fast at the end. This should solve the problem and slightly update the code in the better direction.
- Something that was requested by a lot of players. You don't have to speedrun the nearest abnormality cell 0.001 seconds into the round to keep up with the others. Take it slow, talk to people, *relax*.
- See above. Game will slightly slow down while you are adjusting for the higher level abnormalities.
- I see no reason to keep abnormality count as a factor for qliphoth counter required for meltdown/ordeal since most people will be working on the maximum threat abnormality, instead of working on all of them. The amount of players, on the other hand, is very important.
- ";VIOLET NOON NEXT MELTDOWN!!", and then it doesn't happen. Sounds familiar? Not anymore.
- It's just a QOL change for admins(and me).

### DIFF
--- a/code/controllers/subsystem/abnormality_queue.dm
+++ b/code/controllers/subsystem/abnormality_queue.dm
@@ -1,7 +1,8 @@
 SUBSYSTEM_DEF(abnormality_queue)
 	name = "Abnormality Queue"
 	flags = SS_KEEP_TIMING | SS_BACKGROUND
-	wait = 5 MINUTES
+	runlevels = RUNLEVEL_GAME
+	wait = 10 SECONDS
 
 	/// List of(preferably) 3 abnormalities available for manager to choose from.
 	var/list/picking_abnormalities = list()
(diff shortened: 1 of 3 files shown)


dsmith328 (commented):
Looks good beyond what I've commented on in the discord

PR #424, opened 2022-11-09
Gunblade ego dump

scionalu:
## About The Pull Request

I love PSO I love PSO I love PSO
Anyways, just a random ego idea, it's fine if it doesn't go through
The weapon has what I'd call, a tag system, using marks
(Although I didn't give overlays, because having more than one person with the weapon will make it confusing when I force remove the marks. Currently it just puts the targets in a list, still works though)

Melee gives blade mark, projectile gives gun mark
Hitting a gun mark with melee increases your melee attack speed for 5 sec
Hitting a blade mark with projectile increases your projectile damage for 8 sec
You can't mark when you're buffed
Upon hitting a mark, it goes into a cooldown, forcing the person to use the other mark
These make the player need to constantly switch between melee combat and ranged

Without using the buffs, it's currently the weakest Aleph
Using the gimmick will easily net you 80-90 dps
MinMaxers will get around 110 dps or less (Basically during gun buff, you also hit melee in between shots)

Everything's locally coded in Aleph.dm, so even if this doesn't go through, it's still a nice thing to salvage

## Why It's Good For The Game

To appease the voices in my head
Also neat weapon I guess

## Changelog
:cl:
add: Gunblade
/:cl:


Kitsunemitsu:
Based?

SpaceJohn413:
luster in lobcorp?????

code/game/objects/items/ego_weapons/aleph.dm
```diff
@@ -404,3 +404,104 @@
 
 /obj/item/ego_weapon/censored/get_clamped_volume()
 	return 50
+
+/obj/item/ego_weapon/gunblade
```
Kitsunemitsu, line 408: Wouldn't it be better to refactor this as a gun? Just asking.

scionalu: Honestly, not really sure

Kitsunemitsu (dismissed):
Approved. Code is good.

Kitsunemitsu (approved):
Re-approving after squash.

PR #430, opened 2022-11-10
Some Claw buffs

EgorDinamit:
## About The Pull Request

- Buffs some of the Claw attacks, namely the dash(it will no longer be going in just one direction).
- Decreases cooldowns for some abilities.
- Adds continuous dash attack: It just calls the swift dash 5 times in a row.
- Changes proc names to CamelCase since that's what we're doing nowadays.
- Additionally made all abilities be called by invoke async.

## Why It's Good For The Game

- The easiest to dodge attack got buffed, that's about it. Claw was rather weak for super-late game boss.
- Ditto.
- Makes it easier for the Claw to chase someone who's running too fast.
- Code improvement/standartization.
- Same here.

### DIFF
--- a/ModularTegustation/tegu_mobs/the_claw.dm
+++ b/ModularTegustation/tegu_mobs/the_claw.dm
@@ -1,8 +1,8 @@
 /mob/living/simple_animal/hostile/megafauna/claw
 	name = "Claw"
 	desc = "A strange humanoid creature with several gadgets attached to it."
-	health = 7000
-	maxHealth = 7000
+	health = 8000
+	maxHealth = 8000
 	damage_coeff = list(BRUTE = 1, RED_DAMAGE = 0.4, WHITE_DAMAGE = 0.4, BLACK_DAMAGE = 0.4, PALE_DAMAGE = 0.4)
 	attack_verb_continuous = "slices"
 	attack_verb_simple = "slice"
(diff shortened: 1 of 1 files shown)


ModularTegustation/tegu_mobs/the_claw.dm
```diff
@@ -11,7 +11,7 @@
 	icon_living = "claw"
 	icon_dead = "claw_dead"
 	icon = 'ModularTegustation/Teguicons/48x48.dmi'
-	faction = list("Head")
+	faction = list("hostile", "Head")
```
dsmith328, line 14: Why? Not necessarily a negatively charged question just wondering the intent here

dsmith328:
`blink` causes runtime errors should it's target no longer exist.

ModularTegustation/tegu_mobs/the_claw.dm
EgorDinamit: I don't want Claw to be fighting random abnormalities. It is here to fight the players, not random things.

EgorDinamit:
Updated.

Kitsunemitsu (approved):
Should be fine from a code and balance perspective. Code looks good and all LAZYLEN runtimes should be caught.
Mostly just number and proc name change.

PR #431, opened 2022-11-10
WCCA Damage refactor

Kitsunemitsu:
Update we_can_change_anything.dm

## About The Pull Request
We Can Change Anything now deals damage every work tick.

## Why It's Good For The Game
More accurate to the original, less wounds.

## Changelog
:cl:
tweak: WCCA now does damage every worktick
/:cl:

### DIFF
--- a/code/modules/paperwork/records/info/zayin.dm
+++ b/code/modules/paperwork/records/info/zayin.dm
@@ -99,8 +99,9 @@
 	Max PE Boxes : 10	<br>
 	Qliphoth Counter : X	<br>
 	Work Damage Type : Red	<br>
-	Work Damage : Low	<br>
-	- Once a work is done on we can change anything, the agent working will get hurt in exchange for a temporary buff.	<br>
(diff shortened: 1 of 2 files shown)


code/modules/paperwork/records/info/zayin.dm
```diff
@@ -99,8 +99,9 @@
 	Max PE Boxes : 10	<br>
 	Qliphoth Counter : X	<br>
 	Work Damage Type : Red	<br>
-	Work Damage : Low	<br>
-	- Once a work is done on we can change anything, the agent working will get hurt in exchange for a temporary buff.	<br>
+	Work Damage : See Below	<br>
+	- No matter if a worktick fails or succeeds, the employee working on T-09-85 took damage.	<br>
```
EgorDinamit, line 103: Just keep work damage low and add note below saying that it deals work damage continuously.
"Worktick" doesn't sound very fitting for something that is meant to be used in-character.

PR #432, opened 2022-11-10
Daredevil Parry Update

dsmith328:
## About The Pull Request
Parry now sets your armor instead of just tanking your physiology modifiers. Messing up the parry still uses them to increase damage taken though. Parrying can no longer be done while working (or adjacent to a console undergoing work) nor while you are already parrying. Parrying reduces damage by 90% for R/W/B but only 30% for Pale. If the user has no armor on (including vest) then it instead blocks 95% for R/W/B and 100% Pale. (No armor means that it essentially gives you the equivalent of having all 5/5/5/10 armor with normal parrry). Reduced parry window by 0.1 seconds to 0.5 seconds. If the user has no chest armor and successfully parries an attack, they gain +6 force (non-stacking) to the weapon until it comes off cooldown, for 18 force.

## Why It's Good For The Game
Parry no longer reduces damage to 0, with the except of "naked" Pale (but that's YOUR risk).  Parry window reduced to hopefully make it more reasonable. Gaining damage on parry while "naked" gives the incentive to stay in the fight. 

## Changelog
:cl:
code: Added checks for existing parries so that they can't stack with two daredevils or any other parry item.
code: Improved the method at which Parry applies damage reduction.
code: Improved how different messages are displayed by parry success.
tweak: Changed how damage mitigation is decided by parry.
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/he.dm
+++ b/code/game/objects/items/ego_weapons/he.dm
@@ -171,6 +171,7 @@
 	var/parry = 0
 	var/parry_success
 	var/naked_parry
+	var/list/reductions = list(90, 90, 90, 30, 1)
 
 /obj/item/ego_weapon/daredevil/melee_attack_chain(mob/user, atom/target, params)
 	..()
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (dismissed):

EgorDinamit (changes requested):
Didn't we plan to nerf pale defenses?
I am also not a big fan of giving HE weapon ability to 100% negate any damage, could you change armor to 90 instead?

EgorDinamit (approved):

PR #442, opened 2022-11-13
Mermaid gift nerf

Caldony:
## About The Pull Request

Nerfs the mermaid gift to lower her counter every 1 minute instead. This also indirectly nerfs its success rate as the counter is directly tied to its success rate

## Why It's Good For The Game
Admittedly, 15% was a _little_ over tuned, while there is only one gift, and there is the very high chance of you getting fucked over. The success rate is the equivalent of 75 temperance and just requires you to check up on her every 7 or so minutes.

Now you'll have 3 minute max before she gets pissed off. So she will need constant attention and will feel a lot more needy during ordeals and other long fights.

## Changelog
:cl:
balance: Mermaid's gift now lowers mermaid's counter every minute
/:cl:


Kitsunemitsu (approved):
Well-needed change.
If you want the good Gift you deal with bullshit.

PR #443, opened 2022-11-13
Rabbit IFF

EgorDinamit:
## About The Pull Request

Laser beams used by rabbits now have IFF built in.

## Why It's Good For The Game

No more friendly fire.

https://user-images.githubusercontent.com/53223414/201524769-00a3c4cb-6021-4844-b150-a61d07e864d9.mp4

### DIFF
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -28,6 +28,13 @@
 	wound_bonus = -100
 	bare_wound_bonus = -100
 
+/obj/projectile/beam/laser/prehit_pierce(atom/A)
+	if(isliving(A))
+		var/mob/living/L = A
+		if("rabbit" in L.faction) // RABBIT
+			return PROJECTILE_PIERCE_PHASE
+	return ..()
+
 /obj/projectile/beam/laser/red
 	damage_type = RED_DAMAGE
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/projectile/beams.dm
```diff
@@ -28,6 +28,13 @@
 	wound_bonus = -100
 	bare_wound_bonus = -100
 
+/obj/projectile/beam/laser/prehit_pierce(atom/A)
```
Kitsunemitsu, line 31: Isn't it more processor efficient to do it the way Loyalty does?

EgorDinamit: The way loyalty does it is questionable.

Kitsunemitsu: Not really, this always checks 1 variable, sets one variable, then if it hits something it checks one flag, and if it pierces it checks another flag and returns Loyalty just checks one flag whether it hits or misses, then sets one flag if it hits. It's less lines of code and less processor intensive.

EgorDinamit: You'd like to think so, but I really don't see anything good about the way you did it.

dsmith328 (approved):
Looks real good, no more piles of 5 rabbits

PR #448, opened 2022-11-15
Robots Improved?

dsmith328:
## About The Pull Request
Bots now use Abnormality containment cells, Xeno Spawns, and Main Rooms as viable locations to roam to, should they find no nav beacons. Cleaning Bots are untargettable and otherwise invincible. Clerks now have access to Behaviour panels on Bots, allowing them to do the classic "turn it off and back on again" or simply make them stop patrolling. EGO weapons can now be strapped to Cleaning bots, given they possess no requirement higher than 40. Doing so will make them vulnerable but attack Hostile creatures. They are NOT very tanky but could serve as a brief distraction or even finish off a very wounded Teth. Any weapon with a requirement above 40 will simply fall off upon attempting to attach it to the Cleaning bot. Medbots will now IGNORE someone working on an abnormality console. Armed Cleaning bots can now steal the Lobotomy Corporation job titles of people they trip, where they previously could not.

Unarmed Cleaning Bots are given GODMODE by default. I know this is a very delicate toggle but it does serve a few very specific purposes. The first is that enemies will ignore the bot and vice versa; They won't be infinitely hitting it or just stuck focusing one with lots of health. This one makes sense because from what I recall, Abnormalities target people in particular, as this also generates Enkaphalin, though they won't discriminate against anything going after them. The second is that it protects the bot from any AoE that may come after it. 

## Why It's Good For The Game
Cleaning bots can't patrol as is and, while beloved, typically end up as buckets roughly 20 minutes into the game. They also have a violently untapped potential that was equipping them with weapons to be shit-mobs (and tripping hazards). The equipable weapons are almost always Zayin and TETH unless someone has made a weapon that specifically lacks requirements. Guns are also excluded so that shouldn't be an issue.

Medbots were also being exploited a bit and it seemed a good idea to at least reduce that.

## Changelog
:cl:
code: Changed the CanUseEgo code for `ego_weapon` to check for if it's a bot and if so, check if the weapon has any requirements. 
code: Changed Deputize to allow for bots to equip more than just a kitchen knife.
code/add: Added lines to check for an in-use abnormality console within a tile of potential patients for the Medbot.
code/add: Added godmode checking for potential mobs to be chosen by the Bot as a target.
code/add: Gave bots the ability to repeatedly hit a target based on the `strike` cooldown, modified by the individual weapons attack speed, should it be an EGO Weapon.
add: Gave assistants/clerks Robotics access, so they can operate bots.
mapping: Adds and Removes cleanbots so there's ~4 cleanbots on every map. Also fixes some double turfs on Beta and Delta (they had Space as well as a basic floor turf, so I removed the space)
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/_ego_weapon.dm
+++ b/code/game/objects/items/ego_weapons/_ego_weapon.dm
@@ -12,7 +12,7 @@
 	var/attack_speed
 	var/special
 
-/obj/item/ego_weapon/attack(mob/living/target, mob/living/carbon/human/user)
+/obj/item/ego_weapon/attack(mob/living/target, mob/living/user)
 	if(!CanUseEgo(user))
 		return FALSE
 	. = ..()
@@ -61,7 +61,12 @@
 		display_text += SpecialGearRequirements()
 		to_chat(usr, display_text)
 
-/obj/item/ego_weapon/proc/CanUseEgo(mob/living/carbon/human/user)
+/obj/item/ego_weapon/proc/CanUseEgo(mob/living/user)
+	if(istype(user, /mob/living/simple_animal/bot/cleanbot))
+		for(var/atr in attribute_requirements)
+			if(attribute_requirements[atr] > 40)
+				return FALSE
+		return TRUE
 	if(!ishuman(user))
 		return FALSE
 
--- a/code/modules/mob/living/simple_animal/bot/cleanbot.dm
+++ b/code/modules/mob/living/simple_animal/bot/cleanbot.dm
@@ -17,6 +17,8 @@
 	window_name = "Automatic Station Cleaner v1.4"
 	pass_flags = PASSMOB | PASSFLAPS
 	path_image_color = "#993299"
+	auto_patrol = TRUE // Let's get going, naturally!
+	path_hud = null // No patrol visual. Literally all that was required.
 
 	var/blood = 1
 	var/trash = 0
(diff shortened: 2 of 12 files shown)


dsmith328:
We're in improvement Week so this can definitely wait. Was already working on this so I wanted to push it out before I went and started Paper

code/game/objects/items/ego_weapons/_ego_weapon.dm
```diff
@@ -38,17 +38,21 @@
...
 
-/obj/item/ego_weapon/proc/CanUseEgo(mob/living/carbon/human/user)
-	if(!ishuman(user))
-		return FALSE
-	var/mob/living/carbon/human/H = user
-	for(var/atr in attribute_requirements)
-		if(attribute_requirements[atr] > get_attribute_level(H, atr))
-			to_chat(H, "<span class='notice'>You cannot use [src]!</span>")
+/obj/item/ego_weapon/proc/CanUseEgo(mob/living/user)
+	if(ishuman(user))
```
EgorDinamit, line 42: Why would you switch the way it was done?
First we check for all things that could go wrong, not the other way around.

code/game/objects/items/ego_weapons/_ego_weapon.dm
```diff
@@ -38,17 +38,21 @@
...
+			if(attribute_requirements[atr] > get_attribute_level(H, atr))
+				to_chat(H, "<span class='notice'>You cannot use [src]!</span>")
+				return FALSE
+		if(!SpecialEgoCheck(H))
 			return FALSE
-	if(!SpecialEgoCheck(H))
-		return FALSE
-	return TRUE
+		return TRUE
+	if(isbot(user))
```
EgorDinamit, line 51: If that's your concern, put it at the top instead.

code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -110,9 +128,11 @@
 	prev_access = access_card.access
 	stolen_valor = list()
 
-	prefixes = list(command, security, engineering)
+	prefixes = list(command, security, engineering, LC_highStaff, LC_lowStaff, LC_clerks)
 	suffixes = list(research, medical, legal)
 
+	status_flags ^= GODMODE // There's no real reason for them to be targetted normally, so they're ignored!
```
EgorDinamit, line 134: I think that's excessive.
How about just giving them the hostile faction or making them otherwise untargettable?

EgorDinamit: This will especially be an issue since you allowed cleanbots to equip EGO weapons.

dsmith328: Giving them hostile does not account for various other factions that would make the faction list MASSIVE and not future-proofed for ANY new addition, as every ordeal to my knowledge is JUST the ordeal faction and not the hostile faction. Also when a bot gets a weapon it is no longer God-moded, allowing it to be targeted and fought normally.

code/game/objects/items/ego_weapons/_ego_weapon.dm
dsmith328: Understood

code/modules/mob/living/simple_animal/bot/cleanbot.dm
EgorDinamit: I don't see removal of god mode when "deputized".

Kitsunemitsu (approved):
LGTM. 
If it breaks then it's no skin off my back, and a couple very small lads are now useless

dsmith328:
I may want to make a separate mapping PR to specifically remove the pure quantity of cleanbots we have, because they previously were not as effective.

dsmith328:
Made the mapping changes

Kitsunemitsu (approved):
Reapproving.

PR #449, opened 2022-11-15
Snow White deletes vines on death

Caldony:
## About The Pull Request

Snow-white will now make every vine fade out on death. All the vines will fade out at different interval from 4 to 10 seconds.
The vines will also stop replicating on snow's death.

## Why It's Good For The Game

I'm tired of nature healing, I want nature hurting.

## Changelog
:cl:
tweak: Snow White's vine will now destroy themselves on the abnormality's death

### DIFF
--- a/code/game/objects/items/ego_weapons/waw.dm
+++ b/code/game/objects/items/ego_weapons/waw.dm
@@ -371,14 +371,14 @@
 		return
 	if(vine_cooldown <= world.time)
 		user.visible_message("<span class='notice'>[user] stabs [src] into the ground.</span>", "<span class='nicegreen'>You stab your [src] into the ground.</span>")
-		for(var/obj/structure/alien/weeds/F0442/F in range(0, get_turf(user)))
+		for(var/obj/structure/alien/weeds/apple_vine/F in range(0, get_turf(user)))
 			if(F)
 				playsound(src, 'sound/creatures/venus_trap_hurt.ogg', 10, FALSE, 5)
 				qdel(F)
 				return
 		var/mob/living/carbon/human/L = user
 		L.visible_message("<span class='notice'>Wilted stems grow from [src].</span>")
-		new /obj/structure/alien/weeds/F0442(get_turf(user))
+		new /obj/structure/alien/weeds/apple_vine(get_turf(user))
 		L.adjust_nutrition(-10)
 		vine_cooldown = world.time + vine_delay
 
--- a/code/modules/mob/living/simple_animal/abnormality/waw/snow_whites_apple.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/waw/snow_whites_apple.dm
@@ -1,3 +1,5 @@
+GLOBAL_LIST_EMPTY(vine_list)
+
 /mob/living/simple_animal/hostile/abnormality/snow_whites_apple
 	name = "Snow Whites Apple"
 	desc = "An abnormality taking form of a tall humanoid with an apple for a head."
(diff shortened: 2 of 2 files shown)


code/modules/mob/living/simple_animal/abnormality/waw/snow_whites_apple.dm
```diff
@@ -256,6 +272,16 @@
 	if(!ishuman(L))
 		L.adjustStaminaLoss(10, TRUE, TRUE) //nonhumans are hindered more
 
+/obj/structure/alien/weeds/F0442/proc/get_abno(mob/living/simple_animal/hostile/abnormality/snow_whites_apple/SWA)
```
EgorDinamit, line 275: Why do we even need it?

Caldony: the issue is setting the variable BEFORE the vines are created is, as far as I can tell, impossible. I could make the vine check on Initialize for every datum to find snow, but it's a lot more clean to have it be attributed to a proc instead. If the proc is (somehow?) happening AFTER its death, it'll delete itself anyway as a failsafe.

In other words it's mostly for convenience.

code/modules/mob/living/simple_animal/abnormality/waw/snow_whites_apple.dm
```diff
@@ -77,6 +78,13 @@
 	..()
 
 /mob/living/simple_animal/hostile/abnormality/snow_whites_apple/death(gibbed)
+	for(var/obj/structure/alien/weeds/F0442/vine in vine_list)
+		vine.can_expand = FALSE
+		var/del_time = rand(5,30) //all the vines dissapear at different interval so it looks more organic.
```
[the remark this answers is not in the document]
Caldony: done

code/game/objects/items/ego_weapons/waw.dm
```diff
@@ -371,14 +371,14 @@
 		return
 	if(vine_cooldown <= world.time)
 		user.visible_message("<span class='notice'>[user] stabs [src] into the ground.</span>", "<span class='nicegreen'>You stab your [src] into the ground.</span>")
-		for(var/obj/structure/alien/weeds/F0442/F in range(0, get_turf(user)))
+		for(var/obj/structure/alien/weeds/apple_vine/F in range(0, get_turf(user)))
```
EgorDinamit, line 374: range 0 is its own turf...

code/modules/mob/living/simple_animal/abnormality/waw/snow_whites_apple.dm
```diff
@@ -77,6 +79,12 @@
 	..()
 
 /mob/living/simple_animal/hostile/abnormality/snow_whites_apple/death(gibbed)
```
EgorDinamit, line 81: You might want to move some actions to Destroy() instead, since it *could* get deleted without dying.

code/game/objects/items/ego_weapons/waw.dm
Caldony: I mean yeah, it checks if the person is on top of a vine. Should I make it so it gets the turf of the holder and then checks the contents of that?

code/modules/mob/living/simple_animal/abnormality/waw/snow_whites_apple.dm
Caldony: done

code/game/objects/items/ego_weapons/waw.dm
EgorDinamit: The result would be the same either way, I suppose.

EgorDinamit (approved):

PR #450, opened 2022-11-15
fix mermaid's gift keeping buff after destruction

Caldony:
## About The Pull Request

If mermaid breached by herself it sometimes caused the gifted to keep the buff if they survived mermaid. Fixes this by simply removing the stats on the gift's destruction, so it's more reliable.

## Why It's Good For The Game

bug bad

## Changelog
:cl:
fix: mermaid's gift will no longer sometimes keep its buff after being destroyed.
/:cl:


EgorDinamit (approved):
I will assume you tested it.

PR #453, opened 2022-11-16
Big Bird tweaks

EgorDinamit:
## About The Pull Request

- Big Bird doesn't put you to sleep anymore: Instead it will blind(4s) and stun(2s) you.
- Affected people will have a visible overlay over them to warn others about your impending doom, so that they may try to save you.
- One death will reduce BB's qliphoth counter. This reduces total requirement from 10 to 5, given its counter is at 5.

## Why It's Good For The Game

- Sleep was rather annoying and boring, as it didn't let you scream for help :)
- Just what it says: Allows others to save you from death.
- 10 deaths with our normal pop was TOO much, especially since you can increase its qliphoth counter easily.


Kitsunemitsu (approved):
Looks good, mostly small things.

PR #454, opened 2022-11-16
Judgement Bird range increase

EgorDinamit:
## About The Pull Request

- Changes Judgement Bird's ability range from 8 to 10(one tile off-screen).

## Why It's Good For The Game

- Players were able to kite JBird way too easily. This should change it so running away from JBird after attacking it in melee is a bit harder.


Kitsunemitsu (approved):
Number Change

PR #458, opened 2022-11-17
Buffs clerk pistol

EgorDinamit:
## About The Pull Request

- Using clerk pistols for suicide is now much more effective.

## Why It's Good For The Game

- Now clerks can use their pistol for its true purpose!


dsmith328 (approved):
We do a bit of *head explosions*

PR #459, opened 2022-11-17
Frequency and Telecomms tweaks

EgorDinamit:
## About The Pull Request

- Long range link functionality changed: When it's off, machinery will auto-link with things on its z level, instead of looking in a range of 20.
- Old frequency names replaced with LC stuff.
- Moved some code around.

## Why It's Good For The Game

- Weird function that ruined our communications with placement of telecomms machinery on some maps.
- 'Control' is no longer 'Supply' when you check the telecomms machinery.
- Looks better.


dsmith328 (approved):
Small tweaks that I don't see any issue with as it currently is.

PR #462, opened 2022-11-17
MoSB Upgrade

EgorDinamit:
## About The Pull Request

- Improves MoSB pathfinding to ignore being attacked if it is currently moving towards a corpse.
- While being attacked in melee during patrol, MoSB will retaliate(RIP fast attacking weapons).
- Stage 1 MoSB will now pass through people blocking its path.
- MoSB will prioritize corpses and wounded people not only at stage 1, but also when on low HP.
- Only 2 deaths are required for qliphoth counter to go down, instead of 3.
- Only 2 kills are required for MoSB to grow.
- Melee damage is slightly increased by 5.
- Starting health increased from 1000 to 1500.
- Health increase when growing increased from 500 to 1000; This means stage 3 MoSB will have 3500 HP.
- Most cooldowns are reduced.
- The spit attack deals more damage and hits in range, but fires less projectiles in total.

## Why It's Good For The Game

- No longer can you just punch MoSB to stop its path towards a corpse. Now abnormalities will kite YOU.
- See above.
- You can't cheese MoSB by standing still at a door frame with corpses behind you now.
- It should prioritize healing itself when in a fight.
- 6 deaths for an ALEPH to breach was too much.
- Same as above.
- Fighting tier 1 MoSB, in case it does not find any corpses, shouldn't be *too* easy.
- Killing MoSB is a tiny bit harder now.
- Same as above. Most ALEPHs have 3000 or 4000 HP.
- Old cooldowns were slightly too long, this fixes it.
- This is mostly done to reduce the lag.

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/aleph/mountain.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/aleph/mountain.dm
@@ -5,13 +5,13 @@
 	icon_state = "mosb"
 	icon_living = "mosb"
 	icon_dead = "mosb_dead"
-	maxHealth = 1000
-	health = 1000
+	maxHealth = 1500
+	health = 1500
 	pixel_x = -16
 	base_pixel_x = -16
 	damage_coeff = list(BRUTE = 1, RED_DAMAGE = 1.2, WHITE_DAMAGE = 0.8, BLACK_DAMAGE = 0.8, PALE_DAMAGE = 0.5)
-	melee_damage_lower = 20
-	melee_damage_upper = 30
+	melee_damage_lower = 25
+	melee_damage_upper = 35
 	melee_damage_type = RED_DAMAGE
 	armortype = RED_DAMAGE
 	rapid_melee = 2
@@ -19,6 +19,7 @@
 	ranged = TRUE
 	speed = 2
 	move_to_delay = 2
+	generic_canpass = FALSE
 	attack_sound = 'sound/abnormalities/mountain/bite.ogg'
 	attack_verb_continuous = "bites"
 	attack_verb_simple = "bite"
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/simple_animal/abnormality/aleph/mountain.dm
```diff
@@ -5,20 +5,21 @@
...
+	melee_damage_lower = 25
+	melee_damage_upper = 35
 	melee_damage_type = RED_DAMAGE
 	armortype = RED_DAMAGE
 	rapid_melee = 2
 	stat_attack = DEAD
 	ranged = TRUE
 	speed = 2
 	move_to_delay = 2
+	generic_canpass = FALSE
```
dsmith328, line 22: I'm so mad I didn't know this was a variable

EgorDinamit: It's a cryptic variable that causes normal proc to call another one.

dsmith328 (commented):
So far so good but you've said this can sit around for a while so I'll leave this neutral

code/modules/mob/living/simple_animal/abnormality/aleph/mountain.dm
```diff
@@ -154,8 +137,40 @@
 				belly += 1
 				StageChange(TRUE)
 
+/mob/living/simple_animal/hostile/abnormality/mountain/attacked_by(obj/item/I, mob/living/user)
+	if(LAZYLEN(patrol_path) && CanAttack(user)) // Basically, it will retaliate while patrolling without stopping or chasing after them
+		user.attack_animal(src)
```
dsmith328, line 142: My main concern here is that this would retaliate at the same speed as the attacker hits, so Torn Off Wings would be hit VERY quickly and disproportionally punish fast attackers

DiaFRAME444:
> While being attacked in melee during patrol, MoSB will retaliate(RIP fast attacking weapons).

Can you make it so MoSB only retaliates on a timer? Ex, only if it hasn't retaliated within the last second, or something.

EgorDinamit:
> MoSB is now consistently wiping the server every time it breaches. Maybe you could tone back some of the buffs?

No. The only real change done was making its AI and movement better.
If you don't want MoSB to grow - you must not leave corpses around.

code/modules/mob/living/simple_animal/abnormality/aleph/mountain.dm
Kitsunemitsu: During testing this never came up.

code/modules/projectiles/projectile/magic/abnormality.dm
```diff
@@ -78,12 +78,18 @@
 	icon_state = "mountain"
 	damage_type = BLACK_DAMAGE
 	flag = BLACK_DAMAGE
-	damage = 10 // Launches 32(96) of those, for a whooping 320(960) black damage
+	damage = 15 // Launches 16(48) of those, for a whooping 240(720) black damage
```
Kitsunemitsu, line 81: This is extremely laggy.

Kitsunemitsu (changes requested):
We've been testing this on the server for a while and have no issues with any of the changes from a balance perspective.
Change the amount of projectiles to reduce lag, and other than that the code looks great.

Kitsunemitsu (approved):
Dismissing old review. I was going off TMed and couldn't tell if the projectile lag was TMed or not

PR #466, opened 2022-11-18
Alriune and KoD kill effects

EgorDinamit:
## About The Pull Request

- Alriune and Knight of Despair will now apply an overlay to humans that died from their attacks.

## Why It's Good For The Game

- More visually pleasant things and a good way to tell the cause of death.


dsmith328 (approved):
Very shrimple, VERY GOOD

PR #467, opened 2022-11-18
Death & Insanity tweaks and fixes

EgorDinamit:
## About The Pull Request

- Spreading panic on death will not occur twice(i.e. if a corpse got gibbed or for some reason death() was called again).
- Spreading panic will not affect dead people around the source.
- When someone goes insane, there will be an alert in dead-chat, similar to deaths. It will also show the panic type.

## Why It's Good For The Game

- Fix.
- Fix.
- Observers now can see who's about to breach WN. This is also a nice things for admins.


dsmith328 (approved):
Simple fixes, very nice

PR #469, opened 2022-11-18
Sal acid & mental stabilizator tweaks; Changes examine message for damage.

EgorDinamit:
## About The Pull Request

- Both mental stabilizator and salicylic acid will simply stop healing on overdose, instead of damaging the user/victim. High speed healing of sal. acid will apply when brute damage is above 25%, instead of flat 25.
- Examine messages for damage(severe bruising, etc) will use percentages of max health instead of flat values.

## Why It's Good For The Game

- Sometimes clueless clerks would overdose people on either of these chems, usually resulting in a death of agent. This keeps overdose something you'd like to avoid while not being outright destructive.
- 25 brute damage is moderate in normal SS13 where max health is 100, sure, but here people go up to 230(or more) max health, so it should stay consistent, instead of showing "They have severe bruising!" when it's not even quarter of their max health.

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -187,27 +187,27 @@
 		else
 			temp = getBruteLoss()
 		if(temp)
-			if(temp < 25)
+			if(temp < (maxHealth*0.25))
 				msg += "[t_He] [t_has] minor bruising.\n"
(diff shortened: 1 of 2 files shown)


Kitsunemitsu (approved):
As much as I miss para's flavor text I believe this PR is probably for the best.

PR #476, opened 2022-11-19
TGS chat commands updated

EgorDinamit:
## About The Pull Request

TGS check and status commands will display more useful information, such as:
- Round time.
- Alert level.
- Amount of abnormalities currently in the facility.
- Next ordeal type(only the level, i.e. Dawn or Noon).
- Core suppression status(if one is happening).

## Why It's Good For The Game

More information for the players(and admins) who use Discord to check information about the current round before joining.


Kitsunemitsu (approved):
Code looks good, it's testmerged on the server and works flawlessly

PR #479, opened 2022-11-20
Simple Smile always makes you drop your weapon

Kitsunemitsu:
## Why It's Good For The Game
Smile cat is like hardcore countered by just laying down. This removes that

## Changelog
:cl:
balance: Buff to Gone with a Simple Smile.
/:cl:


EgorDinamit:
You can just make it drop the weapons instead of using stun.

Kitsunemitsu:
It's 1/10th a second of stun.
It's also less lines of code.

EgorDinamit:
You are using a hack: Calling a proc1 that calls proc2, instead of using proc2.
In the end, you are executing more code for less reasons.

monochromatism (dismissed):
Looks good to me. Make the graytider tide harder.

EgorDinamit (approved):

PR #482, opened 2022-11-21
Possession Alter for Ghosts on our Sentient Abnormality Rounds

InsightfulParasite:


Kitsunemitsu (dismissed):
Admin only item that makes things way easier for us.

EgorDinamit:
Why does it have to be a new structure..?

Kitsunemitsu:
So admins can just place it down and forget about it

PR #487, opened 2022-11-23
Punishing bird attack fix

EgorDinamit:
## About The Pull Request

- PBird will not kill people that did not attack it.

## Why It's Good For The Game

- Due to the way rapid melee works, punishing bird could get to hit you with killing move after someone else attacked it while it was targetting you. This fixes it and some more.

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/teth/punishing_bird.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/teth/punishing_bird.dm
@@ -118,7 +118,7 @@
 
 /mob/living/simple_animal/hostile/abnormality/punishing_bird/Life()
 	if(..())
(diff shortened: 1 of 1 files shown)


dsmith328 (commented):
Thing im concerned about is P-bird looking at an insane employee while enraged by something else and wanting to hit them because they're insane, but then not being able to because of being transformed, resulting in P-bird just following them endlessly.

EgorDinamit:
> Thing im concerned about is P-bird looking at an insane employee while enraged by something else and wanting to hit them because they're insane, but then not being able to because of being transformed, resulting in P-bird just following them endlessly.

ListTargets() already handles it. Transformed PBird will only target enemies(those that attacked it).

dsmith328 (approved):
This should work then, here's hoping people stop exploding

PR #488, opened 2022-11-23
Alriune kill effect curtains

EgorDinamit:
## About The Pull Request

Adds the "curtains" that appear for a moment when someone dies to Alriune. Sprites by Lammar#4878

## Why It's Good For The Game

More fancy effects.


code/modules/mob/living/simple_animal/abnormality/waw/alriune.dm
```diff
@@ -78,7 +78,9 @@
 				if(ishuman(L))
 					var/mob/living/carbon/human/H = L
 					if(H.sanity_lost)
-						H.add_overlay(icon('ModularTegustation/Teguicons/tegu_effects.dmi', "alriune_kill"))
+						new /obj/effect/temp_visual/alriune_curtain(get_turf(H))
```
dsmith328, line 81: Doesn't Alriune does this effect regardless of kill, it being her "you were hit by my attack" indicator?

dsmith328: Also does this fail to check for the target being dead, resulting in potentially infinite overlays stacking on a dead insane person?

EgorDinamit: The death check is above;
The curtains only appear on killed: For normal damage there's generic effect.

dsmith328 (approved):
Looks cool, sounds good 👍

PR #507, opened 2022-12-03
Possessing Ghosts now always allows Ghosts to possess.

InsightfulParasite:
## About The Pull Request
Ghosts now always get possession but cannot possess unless a subsystem variable is enabled by the event. There is now a 10 seconds delay on possession whenever a ghost is initialized.
There is also now a banned abnormality list within the Lobotomy Corp subsystem that will not show up on the possession list.
Yes i did test this.
## Why It's Good For The Game
I'm tired of being on spamming the event button duty for Possessing G-G-ghosts.

## Changelog
:cl:
tweak: possession
/:cl:

### DIFF
--- a/code/controllers/subsystem/lobotomy_corp.dm
+++ b/code/controllers/subsystem/lobotomy_corp.dm
@@ -53,6 +53,8 @@ SUBSYSTEM_DEF(lobotomy_corp)
 	var/current_box = 0
 	var/box_goal = INFINITY // Initialized later
 	var/goal_reached = FALSE
+	//possession conditions
+	var/enable_possession = FALSE
 
 /datum/controller/subsystem/lobotomy_corp/Initialize(timeofday)
 	. = ..()
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -3,6 +3,10 @@ GLOBAL_LIST_EMPTY(ghost_images_simple) //this is a list of all ghost images as t
 
 GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 
+GLOBAL_LIST_INIT(unpossessable_mobs, typecacheof(list(
+	/mob/living/simple_animal/hostile/abnormality/woodsman,
+	/mob/living/simple_animal/hostile/abnormality/punishing_bird)))
+
 /mob/dead/observer
 	name = "ghost"
 	desc = "It's a g-g-g-g-ghooooost!" //jinkies!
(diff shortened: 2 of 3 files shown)


InsightfulParasite:
Currently Humans can see it but since i know about invisibility i can change it so only ghosts can see it.

EgorDinamit:
I think it would be better to add a variable/verb that enables posession of abnormalities.

InsightfulParasite:
> I think it would be better to add a variable/verb that enables posession of abnormalities.

I just thought it would be more visually intresting for players than just making it a variable that you can turn on in SSlobotomy_corp and editing possession and ghost initialization to have fun verbs enabled but i guess i can do that instead.

Kitsunemitsu (dismissed):
Approved, this looks fine.

Kitsunemitsu:
Egor, why does this PR need you specifically to look at it?

EgorDinamit (changes requested):
There's too much left-over code.
Is there any reason for adding a typecache list to ghosts when you are using a list from subsystem anyway?

InsightfulParasite:
> There's too much left-over code. Is there any reason for adding a typecache list to ghosts when you are using a list from subsystem anyway?

It was for quick edits during a round since a ghosts initialization cannot be edited during a round ever.

InsightfulParasite:
> There's too much left-over code. Is there any reason for adding a typecache list to ghosts when you are using a list from subsystem anyway?

Oh you also meant the alter structure code. Yeah I did forget about that.

code/controllers/subsystem/lobotomy_corp.dm
```diff
@@ -53,6 +53,8 @@ SUBSYSTEM_DEF(lobotomy_corp)
 	var/current_box = 0
 	var/box_goal = INFINITY // Initialized later
 	var/goal_reached = FALSE
+	//possession conditions
+	var/enable_possession
```
EgorDinamit, line 57:
```suggestion
	var/enable_possession = FALSE
```

code/modules/mob/dead/observer/observer.dm
```diff
@@ -28,7 +28,7 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 							//If you died in the game and are a ghost - this will remain as null.
 							//Note that this is not a reliable way to determine if admins started as observers, since they change mobs a lot.
 	var/atom/movable/following = null
-	var/fun_verbs = 0
+	var/fun_verbs = 1
```
EgorDinamit, line 31: Possession is not the only thing affected by it, as far as I am aware.

InsightfulParasite: Fixed it.

Kitsunemitsu:
Woodsman and Pbird do still need to be blocked.
Neither Woodsman nor pbird are ready to be playable.

EgorDinamit (approved):
I will have to remove that list of banned abnos later, it seems.

PR #547, opened 2022-12-30
Nest Fixes

InsightfulParasite:
## Changelog
:cl:
add: naked nests that have remaining nested serpents when they die will drop a single serpent and command it to hide.
fix: naked nested inheriting damage coefficents if there is no armor to read.
tweak: status effect Rend Armor duration to 6 seconds
tweak: naked nest now drops the hosts armor when it dies.
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -968,3 +968,26 @@
 /datum/status_effect/cloudstruck/Destroy()
 	. = ..()
 	QDEL_NULL(mob_overlay)
+
+
+//~~~LC13 General Debuffs~~~
+
+/datum/status_effect/sunder_red
+	id = "sunder red armor"
+	status_type = STATUS_EFFECT_UNIQUE
+	duration = 60 //3 seconds
+	alert_type = null
+
+/datum/status_effect/sunder_red/on_apply()
+	. = ..()
+	if(isanimal(owner))
+		var/mob/living/simple_animal/M = owner
+		if(M.damage_coeff[RED_DAMAGE] >= 0)
+			M.damage_coeff[RED_DAMAGE] += 0.2 //20% damage increase
+
+/datum/status_effect/sunder_red/on_remove()
+	. = ..()
+	if(isanimal(owner))
+		var/mob/living/simple_animal/M = owner
+		if(M.damage_coeff[RED_DAMAGE] >= 0)
+			M.damage_coeff[RED_DAMAGE] -= 0.2
--- a/code/modules/mob/living/simple_animal/abnormality/waw/naked_nest.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/waw/naked_nest.dm
@@ -76,20 +76,34 @@
 		serpentsnested = serpentsnested + 1
 	return
 
+/mob/living/simple_animal/hostile/abnormality/naked_nest/death(gibbed)
+	for(var/atom/movable/AM in src)
+		AM.forceMove(get_turf(src))
+	if(serpentsnested > 0)
+		var/mob/living/simple_animal/hostile/naked_nest_serpent/S = new(get_turf(src))
+		S.hide()
+	..()
+
 /mob/living/simple_animal/hostile/abnormality/naked_nest/Move()
 	return FALSE
 
 /mob/living/simple_animal/hostile/abnormality/naked_nest/AttackingTarget(atom/attacked_target)
 	return OpenFire()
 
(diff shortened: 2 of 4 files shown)


code/datums/status_effects/debuffs.dm
```diff
@@ -968,3 +968,26 @@
...
+	id = "rend red armor"
+	status_type = STATUS_EFFECT_UNIQUE
+	duration = 60 //3 seconds
+	alert_type = null
+	var/original_armor
+
+/datum/status_effect/rendRedArmor/on_apply()
+	. = ..()
+	var/mob/living/simple_animal/M = owner
+	original_armor = M.damage_coeff[RED_DAMAGE]
```
EgorDinamit, line 985: Not the best way of doing it - if something temporarily debuffs red armor along the way it will basically reset it to some new value and then increase it even higher.
Just add 0.2 on apply and remove 0.2 on remove.

InsightfulParasite: Egor the 0.2 add and 0.2 remove is what im doing right now. Thats the issue. 
1 / 1000 = 0.001
0.001 + 0.2 is 0.201
0.201 * 1000 = 201000
So there does need to be a fix. I just have to use divide and multiplication rather than raw adding

InsightfulParasite: or maybe i do raw add and subtract it? im not sure what your saying egor.

code/modules/mob/living/simple_animal/abnormality/waw/naked_nest.dm
```diff
@@ -76,6 +76,14 @@
 		serpentsnested = serpentsnested + 1
 	return
 
+/mob/living/simple_animal/hostile/abnormality/naked_nest/death(gibbed)
+	for(var/atom/movable/AM in src)
+		AM.forceMove(loc)
```
EgorDinamit, line 81: It's prefered that you use get_turf(src) instead of loc, since loc can be other object at times.

code/modules/mob/living/simple_animal/abnormality/waw/naked_nest.dm
```diff
@@ -220,6 +228,7 @@
 /mob/living/simple_animal/hostile/naked_nest_serpent/proc/hide()
 	for(var/obj/structure/table/S in oview(get_turf(src), 9))
 		Goto(S, move_to_delay, 0)
+		wander = 0
```
EgorDinamit, line 231:
```suggestion
		wander = FALSE
```
It's prefered you use defines instead of numbers.

code/datums/status_effects/debuffs.dm
```diff
@@ -968,3 +968,25 @@
 /datum/status_effect/cloudstruck/Destroy()
 	. = ..()
 	QDEL_NULL(mob_overlay)
+
+
+//~~~LC13 General Debuffs~~~
+
+/datum/status_effect/rendRedArmor
```
EgorDinamit, line 975: Datum names(just like variables) should be in lower_case.
CamelCase is reserved for procs.

code/datums/status_effects/debuffs.dm
```diff
@@ -968,3 +968,25 @@
...
+	if(isanimal(owner))
+		var/mob/living/simple_animal/M = owner
+		if(M.damage_coeff[RED_DAMAGE] >= 0)
+			M.damage_coeff[RED_DAMAGE] += 0.2
+
+/datum/status_effect/sunder_red/on_remove()
+	. = ..()
+	if(isanimal(owner))
+		var/mob/living/simple_animal/M = owner
+		M.damage_coeff[RED_DAMAGE] -= 0.2
```
EgorDinamit, line 992: You never check if damage_coeff is 0 here, meaning you could make immune abnormalities absorb damage instead.

code/modules/mob/living/simple_animal/abnormality/waw/naked_nest.dm
```diff
@@ -210,17 +225,16 @@
 		to_chat(host, "<span class='notice'>You feel something cold touch the back of your leg.</span>")
 	to_chat(src, "<span class='nicegreen'>You've found a new nest!</span>")
 	host.apply_status_effect(/datum/status_effect/serpents_host)
-	qdel(src)
+	addtimer(CALLBACK(src, .proc/death), 1)
```
EgorDinamit, line 228: You might want to use QDEL_IN instead.

EgorDinamit (approved):

PR #446, opened 2022-11-14
Gun Complexity update and Ego fixes

Kitsunemitsu:
## About The Pull Request
Reworks:
Adoration now has 3 attack modes.
DOT, AOE and Shotgun.
Maybe it will finally be used.

Loyalty can turn off it's IFF for bonus damage

Solemn Vow & Solemn Lament get 1.5x damage when used together.

Throwing EVERYTHING scales with justice

Fixes:
Fixed Executive not justice scaling properly
Fixed Jank regarding Harvest's AOE attack

## Why It's Good For The Game
Throwing weapons are a little underpowered, so this will buff those *just* a little.
Loyalty was a Point and click gun, this should make it so that it now has more than just point and click.
Adoration STILL sucks after my buff. Let's do it again.
Solemns are actually more often used as 2 sets of one. I want people to use one of each.

And some Ego fixes

## Changelog
:cl:
tweak: Ego guns get alternate firing modes.
tweak: Solemns now encourage dual wielding
tweak: throwing weapons get justice scaling
fix: Executive and Harvest now work as intended
/:cl:

PR #568, opened 2023-01-22
Harvest ability re-coded and fixed

EgorDinamit:
## About The Pull Request

- Harvest EGO ability re-coded to fit our standards; Additionally, this will fix all the issues that players seem to be having with it.

## Why It's Good For The Game

- Proper code standards should be in place all over our codebase. Also a fix.


Kitsunemitsu:
Fixed in #446

PR #573, opened 2023-01-25
Ports vote panel fix

EgorDinamit:
## About The Pull Request

- Ports #tgstation/tgstation/pull/58489

## Why It's Good For The Game

- Fixes issues with vote panel, because I forgot to port it back then during Tegu times.


dsmith328 (approved):
After checking the comments on the original and this, it seems fine.

PR #575, opened 2023-01-26
Fixes jukebox and lowers volume

EgorDinamit:
## About The Pull Request

- Jukebox will now properly stop the sound for all mobs that have it playing when it stops or gets destroyed.
- Jukebox will play at half the volume.
- dance_machine.dm renamed to jukebox.dm

## Why It's Good For The Game

- Fix.
- Less loud noises out of nowhere.
- Easier to find.


dsmith328 (approved):
Potato commented about it in the Discord but otherwise seems good; Leaving merging to your discretion. If needs a re-review ping me

PR #579, opened 2023-01-28
Removes questionable text

Coxswain-Navigator:
Removes a line of examine text that may have been just funny once, eons ago. Hopefully.

## About The Pull Request
Sends a certain conditional examine text to the backstreets of district 23, where it belongs.

## Why It's Good For The Game
Players who do not enjoy being eaten or eating others will no longer wonder if they are in the wrong server.

## Changelog
:cl:
del: Removed fat examine text
/:cl:


InsightfulParasite:
I get what it means but its just kinda weird even if it only appears when the examiner is hungry.

EgorDinamit (approved):
We should improve preferences and add the home district of characters and then re-add it :)

PR #581, opened 2023-01-28
Fix woodsman

EgorDinamit:
## About The Pull Request

- Woodsman won't be able to buckle mobs to itself when breaching or if he's the one attempting it.
- Replaced ineffective code with a single proc.

## Why It's Good For The Game

- Mostly an issue with player controlled woodsman, but it also prevents accidents where people drag themselves to it mid-fight.
- Code.


dsmith328 (approved):
Cleaner code that prevents a big issue, very nice!

PR #582, opened 2023-01-28
Fixes on_mob_death procs

EgorDinamit:
## About The Pull Request

- MoSB, Big Bird and QoH will now react to insane/afk/otherwise ghosted people dying, if they were previously player controlled.
- The abnormalities will ignore deaths that occured on another z-level.
- Uses IsContained() instead of directly checking for godmode.

## Why It's Good For The Game

- Previously insane people would be ignored as they technically do not have ckey. This fixes it by checking for mind instead.
- They will no longer get triggered by deaths in manager's office or centcom/thunderdome.
- Just code stuff.


dsmith328 (approved):
Code cleanup is nice

PR #584, opened 2023-01-29
Manager can now trigger core suppression

EgorDinamit:
## About The Pull Request

- Adds "auxiliary managerial console" which is used for triggering core suppressions. It is located in manager's office.
- Adds an admin verb to begin core suppression.

## Why It's Good For The Game

- Managers no longer need to rely on admins for the epic post-game content. Although be warned, game allows you to only pick from two randomly chosen types.
- Admin QOL.

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -1110,6 +1110,34 @@ Traitors and the like can also be revived with the previous role mostly intact.
 			continue
 		A.current.AbnoRadio()
 
+// Selects and starts a core suppression
+/client/proc/InitCoreSuppression()
+	set category = "Admin.Fun"
+	set name = "Start Core Suppression"
+	if(!check_rights(R_ADMIN) || !check_rights(R_FUN))
+		return
+
+	if(istype(SSlobotomy_corp.core_suppression))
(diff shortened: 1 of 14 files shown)


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -1110,6 +1110,34 @@ Traitors and the like can also be revived with the previous role mostly intact.
...
+	var/datum/suppression/core_type = input("Select core suppression type","Set Core Type") as null|anything in subtypesof(/datum/suppression)
+	if(!ispath(core_type))
+		return
+	var/run_white = alert(src, "Do you wish to queue white ordeals?", "White Ordeals", "Yes", "No", "Cancel")
+	var/white_ordeals = TRUE
+	if(!(run_white in list("Yes", "No")))
+		return
+	if(run_white == "No")
+		white_ordeals = FALSE
+	SSlobotomy_corp.core_suppression = new core_type
```
dsmith328, line 1133: Even if it's admemes, if something like Training Suppression is replaced by this, employees would be stuck with a -40 to all stats, no? The Ordeal `End` proc should probably contain a bool parameter that when set to `True` simply ends it without providing rewards, so it can be properly replaced.

dsmith328 (changes requested):

dsmith328 (approved):
Approving under the idea that I'm going to implement a proper way for admins to end an on-going suppression, with or without rewards.

PR #585, opened 2023-01-31
Fix murder insanity moving while dead

EgorDinamit:
## About The Pull Request

Murder insanity will not walk toward targets when dead or incapacitated.

## Why It's Good For The Game

Fix.


dsmith328 (approved):
Simple and important conditional, merging asap

PR #592, opened 2023-02-04
Bare Metal and Alleyway EGO abilities

Kitsunemitsu:
## About The Pull Request
Adds abilities to Bare Metal and Alleyway.

Bare metal now has a small pushback effect on use.
Alleyway now has a teleport slash.

## Why It's Good For The Game
Just increases some ego's fun factor.

## Changelog
:cl:
add: Added abilities to Alleyway and Bare metal
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/he.dm
+++ b/code/game/objects/items/ego_weapons/he.dm
@@ -526,6 +526,13 @@
 							FORTITUDE_ATTRIBUTE = 40
 							)
 
+/obj/item/ego_weapon/mini/metal/attack_self(mob/user)
+	if(!CanUseEgo(user) || !do_after(user, 6, src))
+		return
+
+	goonchem_vortex(get_turf(src), 1, 2)
+	to_chat(user,"<span class='warning'>You spin the fanblade, pushing everything back.</span>")
+
 /obj/item/ego_weapon/mini/alleyway
 	name = "alleyway"
 	desc = "It's a small knife forged of black metal."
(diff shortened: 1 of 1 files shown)


EgorDinamit (changes requested):
Only WAW and above deserve abilities. Free teleport on a miniature EGO weapon is a bit too much as well.

On the other note - make use of proper cooldown variables instead of constantly using booleans. As is - your weapons have no cooldowns on use and can be effectively spammed.

Kitsunemitsu:
> Only WAW and above deserve abilities. Free teleport on a miniature EGO weapon is a bit too much as well.
> 
> On the other note - make use of proper cooldown variables instead of constantly using booleans. As is - your weapons have no cooldowns on use and can be effectively spammed.

They have a do-after startup, and the boolean is for turning on and off the teleport ability.

As an aside, almost all EGO right now has an ability of some sort, you just haven't really noticed.
The Teleport also has to hit something, so you are always teleporting into danger, or harming an ally.

Kitsunemitsu:
Thats what the booleans are for. When you start the spin the bool is set to true. You cannot start another do after if the bool is true

code/game/objects/items/ego_weapons/he.dm
```diff
@@ -525,6 +525,19 @@
 	attribute_requirements = list(
 							FORTITUDE_ATTRIBUTE = 40
 							)
+	var/in_use
+
+/obj/item/ego_weapon/mini/metal/attack_self(mob/user)
+	if(!CanUseEgo(user) || in_use)
+		return
+	in_use = TRUE
+	if(!do_after(user, 6))
```
dsmith328, line 534:
```suggestion
	if(!do_after(user, 6, src))
```
I believe this would fix the issue of duplicate and remove the requirement for "in_use"

code/game/objects/items/ego_weapons/he.dm
```diff
@@ -541,6 +554,58 @@
 	attribute_requirements = list(
 							PRUDENCE_ATTRIBUTE = 40
 							)
+	var/teleporting
+
+/obj/item/ego_weapon/mini/alleyway/attack_self(mob/user)
+	if(!CanUseEgo(user))
+		return
+
+	if(!do_after(user, 10))
```
dsmith328, line 563:
```suggestion
	if(!do_after(user, 10, src))
```
So they can't spam a million bars because it would just look jank

dsmith328 (changes requested):

PR #613, opened 2023-02-09
Runtime Fixes

dsmith328:
## About The Pull Request
Should fix the following Runtimes
Runtime in projectile.dm, line 478: bad index
Runtime in _sanityloss_controller.dm, line 328: Cannot read null.status_flags

Simply checks for nulls/types first so less runtimes are thrown.

## Why It's Good For The Game
Should be better :)


code/modules/projectiles/projectile.dm
```diff
@@ -475,6 +475,8 @@
 //Returns true if the target atom is on our current turf and above the right layer
 //If direct target is true it's the originally clicked target.
 /obj/projectile/proc/can_hit_target(atom/target, direct_target = FALSE, ignore_loc = FALSE, cross_failed = FALSE)
+	if(!istype(target)) // Functional Null check as this was causing runtimes looking for "impacted[null]"
```
EgorDinamit, line 478: QDELETED below was already doing this.

PR #614, opened 2023-02-09
Torn Off Wings buff

monochromatism:
## About The Pull Request

Boosts Torn Off Wings to WAW class. Does not boost its straight DPS, but gives it additional utility for both AOE and mobility. 
New combo system:

- Hit enemies to gain combo (up to 16), which will decay after a few seconds out of combat
- Use weapon in hand to do a spin attack (costs 4 combo), hitting all enemies near you - this has a long cooldown and will end your special combo; use it as a finisher or desperation option
- Attack outside weapon range to do a dash attack (costs 4 combo), hitting all enemies near the path of your dash
- Specials also have their own combo system with damage ramp-up
- If used correctly, DPS is just below WAW standard, but with AOE _and_ mobility
- Correct use is demanding

Projected DPS:
Just spam clicking - 27
Using new mechanics to their fullest - 37 dps (10 aoe)

## Why It's Good For The Game

Nobody picks White Lake. Hopefully by making her EGO weapon notably fun, we can change this.

## Changelog
:cl:
balance: rebalanced Torn Off Wings.
/:cl:

### DIFF
--- a/code/datums/abnormality/_ego_datum/he.dm
+++ b/code/datums/abnormality/_ego_datum/he.dm
@@ -35,10 +35,6 @@
 	cost = 35
 
 // White Lake - Torn Off Wings
-/datum/ego_datum/weapon/wings
-	item_path = /obj/item/ego_weapon/wings
-	cost = 35
-
 /datum/ego_datum/armor/wings
 	item_path = /obj/item/clothing/suit/armor/ego_gear/wings
 	cost = 35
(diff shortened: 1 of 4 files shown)


Kitsunemitsu (approved):
Read the code. Looks good.
Whitelake is a cocksucker and likely no one will notice this buff. Needs a slight DPS increase to around 42 later.

PR #617, opened 2023-02-09
Green Midnight

EgorDinamit:
## About The Pull Request

- Adds Green Midnight.
- Fixes a tiny issue with ordeal monitor displaying dawn icon for white ordeals.

## Why It's Good For The Game

- Woe, laser be upon you. Unlike in LC, it doesn't exactly spin. It just fires 6 lasers in random directions trying to destroy everything. Might need some rebalancing to be an actual midnight as it feels too easy right now, if we're not looking at its health(50.000).


EgorDinamit:
Beam disco is almost ready. Also this should be the first message in #repo

Kitsunemitsu (approved):
Was thoroughly tested, Code looks good to me.

PR #622, opened 2023-02-10
Records Officer Watches

Kitsunemitsu:
## About The Pull Request
Gives the RO the ability to use 3 new abilities, The most important of which were talked about before, and all 3 are locked to the RO specifically.

The First item is a timestop watch. It stops time for 11 seconds for everyone but the RO. It's fitting for the records officer just as much as the manager. Timestop is a powerful ability and therefore should require you to move a physical person from A to B to use it. It can be used for 11 seconds every 10 minutes.

The Second item is a watch that gives information on the station, including a little bit of information hidden to the manager. It's not that important.

The Third item is a watch that can delay the next abnormality by 5 minutes. This advances the ordeal timer if people keep working but also can give you a breather during an ordeal. This can be used once every 30 minutes

## Changelog
:cl:
add: Added RO specific abilities
/:cl:


ModularTegustation/tegu_items/records/ordealcheck.dm
```diff
@@ -0,0 +1,43 @@
...
+	if(user.mind.assigned_role != "Records Officer")
+		to_chat(user, "<span class='warning'>You cannot use this!")
+		return
+
+	var/playercount
+	//Count the players
+	for(var/mob/living/carbon/human/L in GLOB.player_list)
+		if(L.stat >= HARD_CRIT || L.sanity_lost || z != L.z) // Dead or in hard crit, insane, or on a different Z level.
+			continue
+		playercount+=1
```
dsmith328, line 18: Incremented but never used.

dsmith328 (changes requested):

dsmith328 (approved):
From what I've been told, this should just be extra RO gaming.

PR #628, opened 2023-02-12
Amber midnight tweaks

EgorDinamit:
## About The Pull Request

- Multiple amber midnights can no longer burrow to the position of another one.
- Before burrowing out, there is now a tiny warning period(smoke + sound).

## Why It's Good For The Game

- Fix.
- More time to notice it before dying.


EgorDinamit:
> Amber midnight needs to spawn dusks that have a cap for maximum dawns. It lags the server way too much.

Not the reason behind lag, at least it hasn't been for a long time.

Kitsunemitsu (approved):
Been Tested, code looks good.

PR #638, opened 2023-02-15
Patrol fix attempts

EgorDinamit:
## About The Pull Request

Attempts to fix patrol issues with MoSB and potentially some other abnos and mobs.

- patrol_reset() is called if mob has target/client/godmode while trying to move by patrol path.
- MoSB target is set to null if high priority turf is found while on phase 1.

## Why It's Good For The Game

Probably fixes something..?


Kitsunemitsu (approved):
LGTM

PR #639, opened 2023-02-15
"Work Forbidden" trait

EgorDinamit:
## About The Pull Request

- Adds a trait that prevents user from using abnormality consoles. This trait is given to: ERTs(Rabbits), Manager, Officers and filthy clerks.
- Moves parent call for ERT on_gain() at the top.

## Why It's Good For The Game

- All of these roles are not meant to use consoles under any circumstances. If clerk or EO catches them on a thought "Damn, I want to work on Nothing There" - then they should spawn as agent.
- This should let them wear stat-locked gear if we add some.


Kitsunemitsu:
Looks good.
I will change the EO/RO cloak to require said trait.

Kitsunemitsu (approved):

EgorDinamit:
> Looks good. I will change the EO/RO cloak to require said trait. And now we can do the stat increases for them.

This would allow ERTs, clerks and manager to wear it as well. I don't think we need to job-lock cloaks.

PR #640, opened 2023-02-15
Sanity rework/improvement

EgorDinamit:
## About The Pull Request

- adjustSanityLoss is no longer reverse. Doing 50 sanity damage reduces sanityhealth by 50, not the other way around.
- Sanity health is now more in line with other damage types. Added sanityloss variable.
- Sanity loss check now takes stat buffs into account when choosing insanity type.

## Why It's Good For The Game

- No more weird code.
- Solves issues with prudence increase putting your current sanity below maximum, easier to get the exact damage numbers.
- Should've been done a long time ago. This mostly affects end-game 130 stats, where +5 buff will decide if you should murder everyone or yourself.


Kitsunemitsu (approved):
Everything looks good.
Recommend pinging developers about this change as it is important.

PR #646, opened 2023-02-16
WN's good effect calls async instead of normal proc

EgorDinamit:
## About The Pull Request

- WN's good effect is called by INVOKE_ASYNC instead of directly.

## Why It's Good For The Game

- Post-work effect(gift-giving, result reporting, etc) isn't delayed if WN does the pulse now.


dsmith328 (approved):
Shouldn't get hung up on healing people then, right?

PR #650, opened 2023-02-17
Overload chance is reset to 0 on meltdown

EgorDinamit:
## About The Pull Request

Instead of resetting to 0 on qliphoth effect specifically, overload chance does that on any meltdown.

## Why It's Good For The Game

Solves issues with Arbiter and Apocalypse Bird causing meltdowns and inevitably leading to -50% overload.


dsmith328 (approved):
Should fix the issue without breaking anything else.

PR #651, opened 2023-02-17
Fortitude insanity tweaks

EgorDinamit:
## About The Pull Request

- Applies to all insanity types - they will now instantly resist out of grabs.
- Fortitude insanity will stop trying to pick up weapons when they are attacked.
- Fortitude insanity will not target godmoded mobs.
- Fortitude insanity will not try to pick up items that it cannot get to.
- Fortitude insanity now has slightly less movement speed.

## Why It's Good For The Game

- No more cheesing with aggro-grabs.
- Avoids silly situations where AI can't pick up item and is just taking a beating.
- They won't hit contained abnos now.
- Fix.
- Mostly to give agents a bit more time to run from scary mimicry agent.


dsmith328 (approved):
Overall important changes.

PR #654, opened 2023-02-18
road home nerf

Caldony:
## About The Pull Request

Nerfs road home to be a little a slower and only hits people with her AOE if she can see them.
I also changed her records from 02 to 01 since I mixed animal and human.

## Why It's Good For The Game

Road home has an instant insanity and insane CC. Despite the fact she's rather easy to kill, it should at least be a lot easier to dodge her status effect considering how big the effect actually is, especially since she's a HE.

## Changelog
:cl:
balance: road home is now slower and only hits people in view with her flip.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/road_home.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/road_home.dm
@@ -7,7 +7,7 @@
 	maxHealth = 1000
 	health = 1000
 	move_resist = MOVE_FORCE_STRONG //So she can't be yeeted away and delayed indefinitely
-	move_to_delay = 10 //She needs to be slow so she doesn't reach home too fast
+	move_to_delay = 13 //She needs to be slow so she doesn't reach home too fast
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/abnormality/he/road_home.dm
```diff
@@ -140,7 +140,7 @@
 		return
 
 	playsound(src, 'sound/abnormalities/roadhome/House_NormalAtk.ogg', 100, FALSE, 8)
-	for(var/mob/living/L in livinginrange(15, src))
+	for(var/mob/living/L in livinginview(15, src))
```
Kitsunemitsu, line 143: Please use In View, Living in view takes all the mobs on the Z level gets their distance to target, and should only really be used about 50

Kitsunemitsu (changes requested):

code/modules/mob/living/simple_animal/abnormality/he/road_home.dm
Caldony: We should probably change rudolta then, fixed.

Kitsunemitsu (approved):
Changes are good.

Kitsunemitsu (approved):
LGTM

PR #667, opened 2023-02-20
Fixes the KHz paper

Auris456852:
image

## About The Pull Request

The radio call paper does not match the numbers you actually need to put in for the last line at the moment. Which, has made me very confused while practicing this. So this should fix that.

## Why It's Good For The Game

Lets people that don't dive through the Git actually use the abnormality.

## Changelog
:cl:
fix: Fixed the KHz radio sheet
/:cl:


Kitsunemitsu:
Fuuuck

Kitsunemitsu (approved):
I had a retard moment. Approving

PR #685, opened 2023-02-26
Removes bolting action from manager's console

EgorDinamit:
## About The Pull Request

Manager console no longer can bolt airlocks.

## Why It's Good For The Game

The only thing it was used for is abusing abnormality's AI.


dsmith328 (approved):
We should revert at some point, once abnos can bypass doors themselves.

PR #687, opened 2023-02-26
Treesap fix

Kitsunemitsu:
## About The Pull Request
Fixes treesap always rolling to kill you

## Changelog
:cl:
fix: fixed tree sap
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_tools/he/treesap.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_tools/he/treesap.dm
@@ -24,7 +24,6 @@
 	if(!do_after(user, 6))
 		return
 	to_chat(user, "<span class='danger'>You sip of the sap.</span>")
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/abnormality/_tools/he/treesap.dm
```diff
@@ -24,14 +24,14 @@
 	if(!do_after(user, 6))
```
dsmith328, line 24:
```suggestion
	if(!do_after(user, 6, src))
```
This will prevent extremely comical event that is multiple sips at once. This event also allows the person to potentially accidentally roll saps multiple times potentially resulting in the fabled "Boom Sap"

dsmith328 (changes requested):

dsmith328 (approved):
Beaut

PR #693, opened 2023-03-03
Pbird no longer can solo ordeals

Kitsunemitsu:
## About The Pull Request
Pbird now does only 50-60 damage against non-humans, even while enraged

## Why It's Good For The Game
Some teth insta-killing ordeals is stupid; same with instakilling random abnos

## Changelog
:cl:
balance: Pbird doesn't obliterate everything now, just humans.
/:cl:


dsmith328 (approved):
Good for balance, honestly.

PR #694, opened 2023-03-03
Real pale damage

EgorDinamit:
## About The Pull Request

Pale damage now works like in Beta: 50 pale damage means you have 50% chance to instantly die.

## Why It's Good For The Game

It is not good by any means.


EgorDinamit:
The meme is over. If I ever want to run a silly joke round again I'll test-merge it again.

PR #702, opened 2023-03-05
Makes Galaxy EGO fire faster without homing

Coxswain-Navigator:
image

## About The Pull Request
Reduces the fire delay when homing is disabled on child of the galaxy's EGO weapon from 8 to 6

## Why It's Good For The Game
There is currently no benefit to disabling homing apart from its unreliability at times. While homing is sometimes useful, this weapon is below par when fired normally. This brings the manual firing mode up to the ballpark of most regular semi's. It is still a projectile weapon that can be difficult to aim at times.

## Changelog
:cl:
tweak: galaxy EGO fires faster without homing
/:cl:


Kitsunemitsu (approved):
Looks Good To me

PR #705, opened 2023-03-06
Crumbling Armor Now only does pale if you are wearing armor

Kitsunemitsu:
## About The Pull Request
Crumbling armor now does 5 red if you aren't wearing armor and 8 pale if you are.

## Why It's Good For The Game
Makes early game pale armor actually useful for one specific instance.

## Changelog
:cl:
balance: Crumbling armor only does pale when you're not wearing armor.
/:cl:


dsmith328:
I'll be honest. I don't see much point in this PR as "early game pale" is LITERALLY non-existance outside of Pale Horse and HE's and if you're using HE's to work a Teth reasonably then I feel like something is wrong.

PR #707, opened 2023-03-06
Shield code cleanup

Coxswain-Navigator:
## About The Pull Request
Removes some unnecessary lines of code and vars

## Why It's Good For The Game
I should have done this a long time ago

## Changelog
:cl:
refactor: refactored some code
/:cl:

PR #710, opened 2023-03-06
Limbus sweep

Coxswain-Navigator:
## About The Pull Request
Adds 12 new TETH E.G.O. weapons, one for each limbus company sinner. All based on the weapons that the sinners usually carry. Currently only obtainable through admins or wishing well.

## Why It's Good For The Game
Cool things you can get sometimes. At least once they have sprites.

## Changelog
:cl:
add: Added new limbus company E.G.O.
/:cl:


Coxswain-Navigator:
By the way, do not merge. #707 will render this code non-functional but it should be fine for test merge.

Kitsunemitsu (changes requested):
Code is good, Will need spriters to come in before merge.

dsmith328 (commented):
Genuine question, we have official names for these EGO, why are they renamed here?

Coxswain-Navigator:
> Genuine question, we have official names for these EGO, why are they renamed here?

These are not the EGO. They are the weapons the sinners use, I'm fairly certain another dev is going to be doing the EGO and they will be HE instead of teth.

dsmith328:
> > Genuine question, we have official names for these EGO, why are they renamed here?
> 
> These are not the EGO. They are the weapons the sinners use, I'm fairly certain another dev is going to be doing the EGO and they will be HE instead of teth.

That was very unclear to me given the PR at the moment, could you change the description to specify it's the Sinners weapons, NOT their ego?

EgorDinamit (approved):
Still not a fan of them being only on TETH level, but we can change that later.

PR #719, opened 2023-03-11
[DNM until 03/25] Removes botany

Kitsunemitsu:
## About The Pull Request
Removes botany from all maps.

## Why It's Good For The Game
I am so sick of botany shenanigans

## Changelog
:cl:
del: Removed botany from all maps.
/:cl:

### DIFF
--- a/_maps/map_files/Alpha/alphacorp.dmm
+++ b/_maps/map_files/Alpha/alphacorp.dmm
@@ -289,15 +289,6 @@
 /obj/item/storage/box/masks,
 /turf/open/floor/carpet/green,
 /area/department_main/safety)
-"cr" = (
-/obj/effect/turf_decal/tile/brown{
-	dir = 1
-	},
-/obj/effect/turf_decal/tile/brown,
-/obj/effect/turf_decal/siding/brown,
-/obj/machinery/hydroponics/constructable,
-/turf/open/floor/plasteel/white,
-/area/department_main/training)
 "cu" = (
 /turf/closed/indestructible/reinforced,
 /area/department_main/safety)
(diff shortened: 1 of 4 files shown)


Kitsunemitsu:
> ...There're much better solutions here, you know.
> You could remove secateurs, remove plant mutation, remove specific mutant plants, or just tweak fertilizers to not give any Instability.

Frankly, Botany was added by a single contributor at the beginning of our server's conception, who is no longer contributing. 
It was added under the pretense that stuff like this would never happen.
I do not care about updating botany, in all honesty. Most rounds in the last 2 or 3 months have one or two people trying some botany shenanigans to ruin the round for everyone in a variety of ways. I have better things to spend dev time on than upkeeping a feature that I never wished for.

"Why dont you ban them?" We have an extremely small admin team and 90% of rounds lack one through the entire thing. Admins aren't needed as much because the coders and community pick up the slack

There is another dev that will fix botany, but if it's not done at the deadline I will get this PR merged.

Kitsunemitsu:
I'll give you more time, go ahead.

Doctor-Derp:
"Not having enough admin power" and letting exploiters get away with things like this, while ALSO hacking away at the codebase to "fix" the problem instead of actually doing it the right way is just plain bad. 

Have you added more basic ingredients to the kitchen? A basic way of fixing this problem would to just be removing the botany vendors and adding select equipment.

You shouldn't remove something just because you don't care enough to fix it, and you shouldn't let people off the hook because you don't care enough to enforce exploit rules.

Kitsunemitsu:
> "Not having enough admin power" and letting exploiters get away with things like this, while ALSO hacking away at the codebase to "fix" the problem instead of actually doing it the right way is just plain bad. 
> 
> Have you added more basic ingredients to the kitchen? A basic way of fixing this problem would to just be removing the botany vendors and adding select equipment.
> 
> You shouldn't remove something just because you don't care enough to fix it, and you shouldn't let people off the hook because you don't care enough to enforce exploit rules.

You are missing the fact that I literally do not care about botany, and if you ban 2 exploit abusers (this isn't technically an exploit btw) more will come crawling out of the walls.

**I literally actually do not give a single fuck about this feature being in the game** after it has caused us problems our entire server's lifespan. If someone else cares, they can fix it.

WolfInfinity:
How about just removing the mutations that are causing the problems? I mean stuff like meatwheat probably isn't causing mass chaos.

Kitsunemitsu:
I am leaving botany to potato to fix, until then this PR will be testmerged.

PR #722, opened 2023-03-12
Disables EO and RO

Kitsunemitsu:
## About The Pull Request
Sets EO and RO job slots to 0

## Why It's Good For The Game
Headcoder doesn't like 'em, weren't supposed to exist.
Clerks can do their jobs going forwards.
image
image
image

## Changelog
:cl:
del: Removed EO and RO.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/command.dm
+++ b/code/modules/jobs/job_types/command.dm
@@ -17,8 +17,8 @@
 	title = "Extraction Officer"
 	department_head = list("Manager")
 	faction = "Station"
-	total_positions = 1
-	spawn_positions = 1
+	total_positions = 0
+	spawn_positions = 0
 	supervisors = "the manager"
 	selection_color = "#ccccff"
 	display_order = JOB_DISPLAY_ORDER_HEAD_OF_PERSONNEL


WolfInfinity:
the code changes only remove one of the two, no? or am I reading it wrong

JackFrost2424:
> the code changes only remove one of the two, no? or am I reading it wrong

Egor dislikes both, Kirie said he's giving him a loaded gun and asking him to fire, so it's for both.

Kitsunemitsu:
> the code changes only remove one of the two, no? or am I reading it wrong

All Officers are a subtype of Extraction so we don't have to copypaste

civilCornball:
RO's ability to slow down Abnormality adding is invaluable as a manager. I've often found myself asking the RO if they could slow down a dangerous abno... and then realizing that we don't have a RO.

EO's only useful for preventing stupid purchases - like that 'random clerk' buying the weapon instead of the armor you need to deal with a more difficult weapon.

Doctor-Derp:
Is this a joke?

Ben10083:
eh RO is fine, EO I can understand but it is moreso as a person to prevent dumb PE purchases (think QM for cargo)

RO is a good role though, most clerks dont even do records but having someone who's job it is to track that stuff is good

TemperanceTempy:
Kirie has enough and distorts, the PR.

EgorDinamit:
Just because I joke about EO/RO(and clerks) being trash doesn't mean I want them gone.

PR #726, opened 2023-03-13
Third fate's loom

Coxswain-Navigator:
## About The Pull Request
Adds a new tool abnormality. When used, you gain 3 stacks of thread. When at 50% health, a thread is consumed and you are healed by the same amount. In addition, you take a small amount of sanity damage. If you run out of threads and reach 50% health, you die. You can re-thread yourself at the tool, but doing so will cause you to permanently lose 15 fortitude.

## Why It's Good For The Game
Is it? Only the players can determine if it's a good idea or not.

## Changelog
:cl:
add: Added third fate's loom
/:cl:


Ben10083:
Eh, this sounds like something with not that many drawbacks. Too forgiving IMO. At the very least it should be quite alot of sanity damage and no rethreading. 

Agents are already hard enough to die if they have high level EGO, why give them at least two 'freebies'?

Coxswain-Navigator:
> Eh, this sounds like something with not that many drawbacks. Too forgiving IMO. At the very least it should be quite alot of sanity damage and no rethreading.
> 
> Agents are already hard enough to die if they have high level EGO, why give them at least two 'freebies'?

I can raise the sanity damage to a percentage, but rethreading must stay an option. I do not think most players would take a 50% health increase if it meant potentially being stuck with only 50% max health forever after taking some damage.

Ben10083:
> I can raise the sanity damage to a percentage, but rethreading must stay an option. I do not think most players would take a 50% health increase if it meant potentially being stuck with only 50% max health forever after taking some damage.

It isn't a 50% increase it is **100%**. I already see how this would play out:

* Agent goes to records, checks for Loom
* Sees loom, takes thread
* Does normal stuff, can survive anomalies it would otherwise not due to 100% increase in resilience (heals back to 100% twice, while 3rd is the death)
* Loses a thread or two during work or suppression, goes back to records to regain all the buffs
* Repeat

Overall, this allows agents to punch much higher above their weight than is otherwise desirable, and even if we account for mental damage the drawbacks are not really a concern (Die instantly after 3rd thread? If you are in a scenario where you get down to 50% two times, an agent would likely just run back to records).

Coxswain-Navigator:
> > I can raise the sanity damage to a percentage, but rethreading must stay an option. I do not think most players would take a 50% health increase if it meant potentially being stuck with only 50% max health forever after taking some damage.
> 
> It isn't a 50% increase it is **100%**. I already see how this would play out:
> 
> * Agent goes to records, checks for Loom
> * Sees loom, takes thread
> * Does normal stuff, can survive anomalies it would otherwise not due to 100% increase in resilience (heals back to 100% twice, while 3rd is the death)
> * Loses a thread or two during work or suppression, goes back to records to regain all the buffs
> * Repeat
> 
> Overall, this allows agents to punch much higher above their weight than is otherwise desirable, and even if we account for mental damage the drawbacks are not really a concern (Die instantly after 3rd thread? If you are in a scenario where you get down to 50% two times, an agent would likely just run back to records).
> 
> EDIT: While yes your cap in fortitude would go down, I still standby that this isn't too much of a problem until late late game.

After discussing it with egor, I decided that it should take an equal amount of sanity to the health you gain. Whether or not this is a good idea needs to be tested.

code/modules/mob/living/simple_animal/abnormality/_tools/he/fateloom.dm
```diff
@@ -0,0 +1,60 @@
...
+	id = "stacking_red_string"
+	status_type = STATUS_EFFECT_UNIQUE
+	duration = -1
+	alert_type = null
+	stack_decay = 0
+	stacks = 4 //was 3, until I figured out that 0 stacks causes stacking status effects to forcibly qdel
+	max_stacks = 4
+	consumed_on_threshold = FALSE
+
+/datum/status_effect/stacking/red_string/tick()
```
EgorDinamit, line 34: You'll have to rewrite it with the use of signals.

code/modules/mob/living/simple_animal/abnormality/_tools/he/fateloom.dm
```diff
@@ -0,0 +1,68 @@
...
+			to_chat(H, "<span class='userdanger'>Your entire body falls apart!</span>")
+			for(var/X in H.bodyparts)
+				var/obj/item/bodypart/BP = X
+				if(BP.body_part && BP.body_part != CHEST)
+					if(BP.dismemberable)
+						BP.dismember()
+			playsound(H, 'sound/abnormalities/fateloom/garrote_bloody.ogg', 80, TRUE, -3)
+			new /obj/effect/gibspawner/generic/silent(get_turf(H))
+			H.regenerate_icons()
+		stacks -= 1
```
EgorDinamit, line 66: It is preferred that you use early returns in such cases.

I assume we should be deleting the status effect after it dismembers the user, isn't that so? Otherwise it could result in a loop of trying to dismember limbs that aren't here anymore.

Also, you might want to kill the user after dismemberment. As funny as it is, I don't think anyone will enjoy being a nugget that can't even move on their own(isn't included in the suggestion).

Also added a (potentially obsolete) stat and QDELETED check just so it doesn't run when it shouldn't
```suggestion
/datum/status_effect/stacking/red_string/proc/heal()
	SIGNAL_HANDLER
	var/mob/living/carbon/human/H = owner
	if(QDELETED(H) || H.stat == DEAD)
		return
	if(H.health > H.maxHealth * 0.5)
		return

	if(stacks > 1)
		playsound(H, 'sound/abnormalities/fateloom/garrote.ogg', 80, TRUE, -3)
		H.adjustBruteLoss(-(H.maxHealth * 0.5))
		H.adjustSanityLoss(H.maxHealth * 0.5) // lose sanity by how much health you gain
		if(stacks > 2)
			to_chat(H, "<span class='userdanger'>You lose some of your threads!</span>")
		else if(stacks == 2)
			to_chat(H, "<span class='userdanger'>You are running low on threads!</span>")
		stacks -= 1
		return

	to_chat(H, "<span class='userdanger'>Your entire body falls apart!</span>")
	for(var/X in H.bodyparts)
		var/obj/item/bodypart/BP = X
		if(BP.body_part && BP.body_part != CHEST)
			if(BP.dismemberable)
				BP.dismember()
	playsound(H, 'sound/abnormalities/fateloom/garrote_bloody.ogg', 80, TRUE, -3)
	new /obj/effect/gibspawner/generic/silent(get_turf(H))
	H.regenerate_icons()
	QDEL_NULL(src)
```

code/modules/paperwork/records/info/tools.dm
```diff
@@ -94,3 +94,14 @@
...
+
+//Loom
+/obj/item/paper/fluff/info/tool/fateloom
+	name = "T-09-142"
+	info = {"<h1><center>T-09-142</center></h1>	<br>
+	Name : Third Fate's Loom <br>
+	Risk Class: HE <br>
+	- Those who interact with T-09-142 are sewn together with red thread. <br>
+	- If the user’s HP drops below 50, they will be healed. The user loses thread and experiences significant mental corrosion when this happens. <br>
+	- If the user’s HP drops below 50 when threads are too thin, they will collapse into pieces.
```
EgorDinamit, line 106: You should clarify that it is below 50%, not *just* 50.

code/modules/mob/living/simple_animal/abnormality/_tools/he/fateloom.dm
Coxswain-Navigator: > It is preferred that you use early returns in such cases.
> 
> I assume we should be deleting the status effect after it dismembers the user, isn't that so? Otherwise it could result in a loop of trying to dismember limbs that aren't here anymore.
> 
> Also added a (potentially obsolete) stat and QDELETED check just so it doesn't run when it shouldn't
1. Ah, forgot to implement this after removing signals. Good catch.

2. Reducing stacks to 0 removes the status effect automatically, handled in stacking status_effects.

3. Check is probably obsolete but doesn't hurt not to see runtimes in the future. 

Tested the suggestion and everything works fine, so I'm implementing it.

EgorDinamit (approved):

PR #727, opened 2023-03-14
Attempts to remove botany exploits

PotatoTomahto:
## About The Pull Request
- Uranium and plasma mats no longer do their things
- Bamboo recipes removed
- Gaia no more
- Cotton can no longer become durathread
- Slip and bluespace traits removed
- All pyrotechnic recipes no longer exist

## Why It's Good For The Game
Look, botany is fun. I would rather not have to remove every reagent from plants and make mutations impossible. This should solve the biggest issues we've had with botany while also allowing for some hijinks.

## Changelog
:cl:
del: botany and chemistry and mats
:cl:

### DIFF
--- a/code/modules/hydroponics/grown/random.dm
+++ b/code/modules/hydroponics/grown/random.dm
@@ -16,8 +16,6 @@
 /obj/item/seeds/random/Initialize()
 	. = ..()
 	randomize_stats()
-	if(prob(60))
-		add_random_reagents(1, 3)
 	if(prob(50))
 		add_random_traits(1, 2)
 	add_random_plant_type(35)
--- a/code/modules/hydroponics/plant_genes.dm
+++ b/code/modules/hydroponics/plant_genes.dm
@@ -243,29 +243,6 @@
 	// Squash the plant on slip.
 	G.squash(C)
 
-/datum/plant_gene/trait/slip
-	// Makes plant slippery, unless it has a grown-type trash. Then the trash gets slippery.
-	// Applies other trait effects (teleporting, etc) to the target by on_slip.
-	name = "Slippery Skin"
-	rate = 1.6
-	examine_line = "<span class='info'>It has a very slippery skin.</span>"
-
-/datum/plant_gene/trait/slip/on_new(obj/item/food/grown/G, newloc)
-	..()
-	if(istype(G) && ispath(G.trash_type, /obj/item/grown))
-		return
-	var/obj/item/seeds/seed = G.seed
-	var/stun_len = seed.potency * rate
-
-	if(!istype(G, /obj/item/grown/bananapeel) && (!G.reagents || !G.reagents.has_reagent(/datum/reagent/lube)))
-		stun_len /= 3
-
-	G.AddComponent(/datum/component/slippery, min(stun_len,140), NONE, CALLBACK(src, .proc/handle_slip, G))
-
-/datum/plant_gene/trait/slip/proc/handle_slip(obj/item/food/grown/G, mob/M)
-	for(var/datum/plant_gene/trait/T in G.seed.genes)
-		T.on_slip(G, M)
-
 /datum/plant_gene/trait/cell_charge
 	// Cell recharging trait. Charges all mob's power cells to (potency*rate)% mark when eaten.
 	// Generates sparks on squash.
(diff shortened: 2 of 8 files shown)


dsmith328 (commented):
Good notation, TMing first.

code/modules/hydroponics/grown/random.dm
```diff
@@ -16,8 +16,6 @@
 /obj/item/seeds/random/Initialize()
 	. = ..()
 	randomize_stats()
-	if(prob(60))
-		add_random_reagents(1, 3)
```
EgorDinamit, line 20: We might keep it if reagents were pulled from a list and not entirely random.

code/modules/hydroponics/plant_genes.dm
```diff
@@ -243,28 +243,7 @@
...
-
-	if(!istype(G, /obj/item/grown/bananapeel) && (!G.reagents || !G.reagents.has_reagent(/datum/reagent/lube)))
-		stun_len /= 3
-
-	G.AddComponent(/datum/component/slippery, min(stun_len,140), NONE, CALLBACK(src, .proc/handle_slip, G))
-
-/datum/plant_gene/trait/slip/proc/handle_slip(obj/item/food/grown/G, mob/M)
-	for(var/datum/plant_gene/trait/T in G.seed.genes)
-		T.on_slip(G, M)
+//SLIP TRAIT REMOVED
```
EgorDinamit, line 246: No need to put comments like that.

code/modules/hydroponics/plant_genes.dm
```diff
@@ -367,31 +346,7 @@
...
-	var/teleport_radius = max(round(G.seed.potency / 10), 1)
-	var/turf/T = get_turf(C)
-	to_chat(C, "<span class='warning'>You slip through spacetime!</span>")
-	do_teleport(C, T, teleport_radius, channel = TELEPORT_CHANNEL_BLUESPACE)
-	if(prob(50))
-		do_teleport(G, T, teleport_radius, channel = TELEPORT_CHANNEL_BLUESPACE)
-	else
-		new /obj/effect/decal/cleanable/molten_object(T) //Leave a pile of goo behind for dramatic effect...
-		qdel(G)
+//BLUESPACE TRAIT REMOVED
```
EgorDinamit, line 349: Ditto.

lobotomy-corp13.dme
```diff
@@ -3161,7 +3161,6 @@
 #include "code\modules\reagents\chemistry\recipes\drugs.dm"
 #include "code\modules\reagents\chemistry\recipes\medicine.dm"
 #include "code\modules\reagents\chemistry\recipes\others.dm"
-#include "code\modules\reagents\chemistry\recipes\pyrotechnics.dm"
```
EgorDinamit, line 3164: Instead of only removing it from .dme, you can delete the file entirely.

code/modules/hydroponics/hydroponics_chemreact.dm
```diff
@@ -27,6 +27,6 @@
 		if(11 to 20)
 			mutateweed()
 			return
-		if(1 to 10)
+		if(0 to 10)
```
out-of-phaze, line 30: This line probably has a merge conflict now, since it was changed in the linter PR.

out-of-phaze (changes requested):
Why not blacklist uranium and plasma from metalgen instead, if that's the issue with plasma/uranium plants? Lobotomizing the game by removing banana slipping, material effects, random reagents (instead of blacklisting problem chems, which *already exists*), and *every single pyrotechnic reaction* to fix issues with a fundamentally broken system with a *lot* of griefing potential beyond what you've already removed here is a bad idea. The argument that some of these things don't belong anyway doesn't pass muster, because the same applies to a lot of botany itself, like mutations, and if you were removing this stuff for not making sense in the setting you went about it the wrong way anyway.

A much better solution, I think, would be to keep the parts of botany that people like—from asking around, that seems to mostly just be growing food—and remove the parts that cause issues, like mutations and splicing. For plants available via mutation that are still desirable, like meatwheat, alternate methods of acquiring them could be added.

Gboster-0:
> Blue-tomatoes still have space lube, and can therefore be used in slips. Liquid contents still exists, and can therefore be combined (via secateurs) with any plant containing a hazardous chemical to devastating results (against humans). Death-nettles still exist, and are rather nasty weapons (they do damage comparable to energy swords on other servers, though it's halved every time you hit someone) you can carry a hundred of in a plant bag. I do not see why Gaia was removed.

blue tomatoes and nettles were removed several months ago, before botany was seen.
About liquid contents... you can just use a normal EGO weapon to much more devastating effects if you are going to grief. believe me nobody wants to bother doing something for an hour to grief with medium efficiency if you can do the same in 10 minutes with high efficiency

dsmith328 (approved):
This should be re-reviewed and replaced with better changes later. Merging because of actual issues this solves.

PR #733, opened 2023-03-14
LONG BIRD BASEBALL BAT

dsmith328:
## About The Pull Request
No more shall he be as a Clerk with a Skateboard, his knees are now shattered so he can't both use his ability and move at the same time.

## Why It's Good For The Game
He doesn't do it in base game and it kinda feels scummy as it currently is, especially as it does not target sight but an objective radius.


EgorDinamit (changes requested):
It was something that jbird had in the past.
Issue with it is the fact it can be easily dodged when it does not follow you.

Either reconsider the idea entirely, or increase radius/damage to compensate for it.

EgorDinamit (approved):

EgorDinamit:
I still feel like it will nerf JBird too hard, however.

PR #734, opened 2023-03-14
Apostle Buff

dsmith328:
## About The Pull Request
Makes non-Guardian apostles patrol.

## Why It's Good For The Game
They're currently a non-issue due to the fact most of them spawn out of the way and can be ignored, this makes WN breach a lot more scary.


EgorDinamit (approved):
I was already doing it manually by var-editing.
It works just fine and shakes up WN breach a bit.

PR #752, opened 2023-03-18
Fix manager camera examination

Caldony:
## About The Pull Request
The camera examination showed abnormalities that were immune to a specific type of damage as endured, this fixes that.
The camera is also now capable of showing the resistances of most mobs including apocalypse bird eggs, and anything that could be considered a "mob structure". Codewise this just means the check extends to simple animals instead of just hostile mobs.
This also fixes a problem with the abnormalities helper code that would show absorbed instead of immune

## Why It's Good For The Game

The camera showing false information is kind of bad I think.

## Changelog
:cl:
fix: The camera examination now properly shows when something is "immune" to a certain damage type.
fix: The camera examination now shows the damage resistances of most non-human mobs, including apocalypse bird eggs.
/:cl:

### DIFF
--- a/code/__HELPERS/abnormalities.dm
+++ b/code/__HELPERS/abnormalities.dm
@@ -1,29 +1,29 @@
 /// This proc returns text explanation of simple mob's resistances
 /proc/SimpleResistanceToText(resist = 1)
 	switch(resist)
+		if(0 to 0) //Just putting 0 doesn't work.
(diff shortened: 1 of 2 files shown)


dsmith328 (approved):
Optimizes manager code to use new Procs and I've seen the discussion regarding this fixing the displays.

PR #754, opened 2023-03-18
Dreaming current fix(?) and tweak

EgorDinamit:
## About The Pull Request

- Added max distance variable to dreaming current and therefore increase it from default 30 to 40.
- Added a simple loop to finding the path to next node, so it doesn't fail all that often.
- DC won't play sound and animation if there's nowhere to move.

## Why It's Good For The Game

- Fixes an issue DC had on big maps where it would circle through same department forever.
- A fail-safe, nothing else.
- Fix. DC doesn't look dumb anymore.

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/waw/dreaming_current.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/waw/dreaming_current.dm
@@ -44,7 +44,7 @@
 	/// Delay between each subsequent move when charging
 	var/dash_speed = 0.8
 	/// How many paths do we create between several landmarks?
-	var/dash_nodes = 5
+	var/dash_nodes = 6
 	/// Maximum AStar pathfinding distances from one point to another
 	var/dash_max_distance = 40
 	var/datum/looping_sound/dreamingcurrent/soundloop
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/abnormality/waw/dreaming_current.dm
```diff
@@ -103,10 +114,20 @@
 		var/turf/T = get_closest_atom(/turf/open, potential_turfs, picking_from)
 		if(!T)
 			break
-		movement_path += get_path_to(path_start, T, /turf/proc/Distance_cardinal)
+		var/list/our_path = list()
+		for(var/o = 1 to 3) // Grand total of 3 retries
+			our_path = get_path_to(path_start, T, /turf/proc/Distance_cardinal, dash_max_distance)
+			if(islist(our_path) && LAZYLEN(our_path))
+				break
+			potential_turfs -= T // Couldn't find path to it, don't try again
```
dsmith328, line 122: Wouldn't this result in a duplicate removal if it initially could not get a path, but later managed to on, say loop 2 or 3?

dsmith328 (commented):
Seems good, would like comment addressed before strict approval is given.

code/modules/mob/living/simple_animal/abnormality/waw/dreaming_current.dm
EgorDinamit: Right, I will have to pick a new T variable here.

dsmith328 (approved):
More range and work exclusion are both very good changes for him

PR #755, opened 2023-03-19
Punishing Bird tweaks

EgorDinamit:
## About The Pull Request

- Slightly increased PBird's "angry damage" against abnos and added a variable for it.
- PBird can now be possessed again.

## Why It's Good For The Game

- It was doing almost no damage at all. Made it a variable for playable abnos rounds.
- PBird has no reasons to be unpossessable anymore.
hould generally represent how a player might be affected by the changes rather than a summary of the PR's contents. -->


dsmith328 (approved):
Simple and easy change

PR #757, opened 2023-03-19
Changes green midnight's behavior

EgorDinamit:
## About The Pull Request

Reworked switch statement to be ever so slightly better and less hardcoded.
Now it will continuously speed up and increase amount of lasers.

## Why It's Good For The Game

Both a fix, and improvement/balance of green midnight.


dsmith328 (approved):
Nothing's changed since last reviewed so I'm approving as it makes the end very much an important rush

PR #763, opened 2023-03-20
SpawnAbno but as a verb

InsightfulParasite:
## About The Pull Request
Works on code similar to debug "spawn object-mob" where you enter approximately the name.
image
I tested it for runtimes and fixed said runtimes so it should be safe... for now.
image
You can technically spam the verb and the Enter key to spawn random abnormalities fast.

## Why It's Good For The Game
Its a quality of life update for admins and coders.

## Changelog
:cl:
add: spawn abnormality cell Fun Verb
/:cl:


EgorDinamit (changes requested):
There's no need for it to be in Admin.Fun category.
You should also rename the proc name(remove Init part) and query(it spawns abno, not simply replaces queued one).

EgorDinamit (approved):

PR #771, opened 2023-03-22
QoH beam anywhere

EgorDinamit:
## About The Pull Request

QoH can now fire beam in any angle, that's about it.

## Why It's Good For The Game

I am not yet sure if it will make QoH more or less dangerous, but we can work it out later.
This allows QoH to be ever so slightly more precise and unpredictable.

image

image


dsmith328 (commented):
Issue seems resolved, waiting on squash

dsmith328 (approved):
"Fricken Laser beams" - Dr Evil

PR #773, opened 2023-03-22
Porccubus buff

Caldony:
## About The Pull Request

Porccubus had this issue of being the only HE requiring a level 4 stat, with very little to give back for it.

- The drug now lasts around 25% longer, and can be taken at any attribute level with a special work, the threat of death is usually enough of a deterrent to make people not want to take the drug, and I don't think the temperance barrier was ultimately necessary.

- Temperance requirement is now level 3, which means his high attachment rate actually has a reason to exist now.

I also changed a part of the code that would runtime when special works were done, functionally it's the same thing, I've tested it.

## Why It's Good For The Game

This should make porccubus a more viable pick instead of a meme pick, while still keeping his trademark temperance check and giving his drugs a little bit more versatility and use in the endgame.

## Changelog
:cl:
tweak: Porccubus now has a special work to drug yourself at any time.
balance: Porccubus now requires temperance level 3 instead of 4 and his drug lasts longer.
fix: fixes a runtime that triggered on special works.
/:cl:

### DIFF
--- a/code/datums/abnormality/datum/abnormality.dm
+++ b/code/datums/abnormality/datum/abnormality.dm
@@ -166,6 +166,8 @@
 		if(ALEPH_LEVEL)
 			maximum_attribute_level = 130
 	var/datum/attribute/user_attribute = user.attributes[attribute_type]
+	if(!user_attribute) //To avoid runtime if it's a custom work type like "Release".
+		return
 	var/user_attribute_level = max(1, user_attribute.level)
 	var/attribute_given = clamp(((maximum_attribute_level / (user_attribute_level * 0.25)) * (0.25 + (pe / max_boxes))), 0, 16)
 	if((user_attribute_level + attribute_given) >= maximum_attribute_level) // Already/Will be at maximum.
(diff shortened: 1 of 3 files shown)


code/datums/abnormality/datum/abnormality.dm
```diff
@@ -166,16 +166,18 @@
...
-	if((user_attribute_level + attribute_given) >= maximum_attribute_level) // Already/Will be at maximum.
-		attribute_given = max(0, maximum_attribute_level - user_attribute_level)
-	if(attribute_given == 0)
-		if(was_melting)
-			attribute_given = 2 //pity stats on meltdowns
-		else
-			to_chat(user, "<span class='warning'>You don't feel like you've learned anything from this!</span>")
-	user.adjust_attribute_level(attribute_type, attribute_given)
+
+	if(user_attribute) //To avoid runtime if it's a custom work type like "Release".
```
EgorDinamit, line 170: Use early returns.

code/modules/mob/living/simple_animal/abnormality/he/porccubus.dm
```diff
@@ -50,12 +51,14 @@
 
 
 /mob/living/simple_animal/hostile/abnormality/porccubus/SuccessEffect(mob/living/carbon/human/user, work_type, pe)
+
+
```
EgorDinamit, line 55: Remove empty space.

code/datums/abnormality/datum/abnormality.dm
Caldony: done

EgorDinamit (approved):

PR #774, opened 2023-03-22
Void Dream update

EgorDinamit:
## About The Pull Request

- Added all the sounds for Void Dream. This includes: Ambient sounds during second form, firing/shouting sounds, impact/ability sound for sleeping bolt and firing sounds for EGO.
- Moved transformation copy-pasta to a separate proc.
- Reduced sleep time from 60 to 30 seconds.
- Sleep can now be stopped upon taking damage(above 5).
- Void Dream's projectile won't target already sleeping people.
- Void Dream's shouting ability was considerably buffed in damage. Now attacking it is probably not the best course of actions.

## Why It's Good For The Game

Overall improvements for this abnormality; Always felt like it's lacking something.


dsmith328 (changes requested):
Overall improvements, should include something checking for apply_damage and attack_hand for weapon abilities and punching respectively.

EgorDinamit:
> Overall improvements, should include something checking for apply_damage and attack_hand for weapon abilities and punching respectively.

Extremely edge-case scenarios. If someone wants to punch Void Dream to death - be my guest.
If I were to use apply damage it would get triggered by any source of damage, including ordeals and other abnormalities, so let's keep it simple.

dsmith328 (approved):
Please squash then ping me for reapproval

dsmith328 (approved):

PR #775, opened 2023-03-22
Abnormality queue rank tweaks

EgorDinamit:
## About The Pull Request

- HE abnormalities will not appear as soon as ALEPHs become available.
- At the later point of the round(last 2 abno slots), only ALEPHs can be picked.

## Why It's Good For The Game

- Should solve the problem with RNG being too extreme at times, giving you HE abnormalities when everyone is a level 4-5 agent.
- Similar to the above, but also prevents certain kinds of managers from sitting in a nice corner, only picking WAW abnormalities without a single ALEPH.


dsmith328 (approved):
No longer will I be building extra rooms because the agents "weren't ready" when they had 2 rooms remaining, and then inevitably watch them die because they weren't ready for ADDITIONAL abnos past 24.

PR #779, opened 2023-03-26
Most abnormalities now share faction

EgorDinamit:
## About The Pull Request

- CENSORED, Melting Love, White Night and Warden now have "hostile" faction.
- Removes obsolete lines from most abnormalities.

## Why It's Good For The Game

- People are relying on breaching other abnormalities to handle their fuck-ups way too much. This should not be possible, outside of abnos like QoH or Shepherd.
- Hostile mobs have "hostile" faction by default, no need to define it everywhere.

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/aleph/censored.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/aleph/censored.dm
@@ -6,7 +6,6 @@
 	icon_living = "censored"
 	pixel_x = -16
 	base_pixel_x = -16
-	faction = list("censored")
 	speak_emote = list("screeches")
 	attack_verb_continuous = "attacks"
 	attack_verb_simple = "attack"
@@ -160,7 +159,6 @@
 	icon = 'ModularTegustation/Teguicons/32x32.dmi'
 	icon_state = "censored_mini"
 	icon_living = "censored_mini"
-	faction = list("censored")
 	speak_emote = list("screeches")
 	attack_verb_continuous = "attacks"
 	attack_verb_simple = "attack"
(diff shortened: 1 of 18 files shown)


UnokiAs:
Might break "code/modules/projectiles/projectile/magic/abnormality.dm"
`/obj/projectile/melting_blob/on_hit(target)
	if(ismob(target))
		var/mob/living/H = target
		if("slime" in H.faction)
			H.visible_message("<span class='warning'>[src] vanishes on contact with [target]!</span>")
			qdel(src)
			return BULLET_ACT_BLOCK
		if(H.stat == DEAD && ishuman(H))
			var/turf/T = get_turf(H)
			visible_message("<span class='danger'>[src] submerge in slime \the [H] and another Slime Pawn appears!</span>")
			H.gib()
			new /mob/living/simple_animal/hostile/slime(T)
			return BULLET_ACT_HIT
		H.apply_damage(rand(mindmg,maxdmg), BLACK_DAMAGE, null, H.run_armor_check(null, BLACK_DAMAGE))
		if(!isbot(H) && isliving(H))
			H.visible_message("<span class='warning'>[target] is hit by [src], they seem to wither away!</span>")
			for(var/i = 1 to 10)
				addtimer(CALLBACK(H, /mob/living/proc/apply_damage, rand(4,6), BLACK_DAMAGE, null, H.run_armor_check(null, BLACK_DAMAGE)), 2 SECONDS * i)
			return BULLET_ACT_HIT
	. = ..()`

EgorDinamit:
Alright, will update projectiles as well.

dsmith328:
There's a lot of code cleanup in here, that's great. But I am going to miss the abno brawls.

dsmith328 (approved):

PR #783, opened 2023-03-27
Torn Off Wings Update

dsmith328:
## About The Pull Request
Torn Off Wings, despite it's sick combo system, was under performing. This should resolve this.
The changes to the weapon are as following:
Force decreased from 15 to 10.
Attacks twice per click.
Special Move damage increased from 15 to 20.
Special Move damage multiplier increased from 0.1 to 0.2. (More specials means more damage)
Special Move cooldown removed for spin, and decreased from 1 Second to 0.3 Seconds.
Dash and Spin effects specified in Description.
Special Moves will now affect insane humans but not sane humans.
Special Moves say who they hit. 

## Why It's Good For The Game
The weapon was supposed to be a WAW but the mechanics, though cool, felt clunky to use and not overall great. The changes made here make it feel better without purely changing numerical values.


Kitsunemitsu (approved):
Approved; Code looks fine and Concept is good.

PR #786, opened 2023-03-27
Legacy LC

EgorDinamit:
## About The Pull Request

- Adds legacy abnormality datum subtype. Instead of relying on qliphoth meter, it is using mood that constantly decays.

## Why It's Good For The Game

It is not good, it's just hilarious.


Kitsunemitsu:
This was for april's fools

PR #787, opened 2023-03-27
New injectors

WolfInfinity:
New injectors to the extraction console. Mainly for more versatility and in case of emergencies. 
Example of an emergency: 1. No one has the stats to work X abno, pop a booster. 2. Big abno broke out. Need an extra oomph for combat(Look to the Ideas section for how to prevent this from being abused in combat).

Currently Coded:
-Big Booster Injector - Big temporary stat boost, afterwards temporary stat loss.
-Small Booster Injector - Small temporary stat boost.
-Small Reductor Injector - Small temporary stat reduction.

Planning to do:
-C-Fear Reduction - Alternate to C-Fear Immunity that was commented out of the New Gadgets PR(Still discussing the plans for this one as it needs bigger bois to mess with the bigger files)

Ideas:
The numbers can be balanced either by changing them or by adding two mechanics:
-Addiction and Withdrawal. Take too many boosters over the character's lifetime(or within a certain period of time) and the withdrawals will be worse and worse.
-Overdose. Too many boosters at one time have a chance to overdose the more they are taken on top of one another.


WolfInfinity:
Note this is branching off of the New Gifts PR.

EgorDinamit (changes requested):
There should be no freely available items that boost agent attributes, even if temporary.
That would interfere with normal gameplay way too much, allowing people to commit more mistakes and still come out fine.

The "big boost" is not getting through under any circumstances.

Coxswain-Navigator:
I might be mistaken but aren't the injectors are meant only for staff positions such as EO and RO? They should be fine if so.

WolfInfinity:
> There should be no freely available items that boost agent attributes, even if temporary. That would interfere with normal gameplay way too much, allowing people to commit more mistakes and still come out fine.
> 
> The "big boost" is not getting through under any circumstances.

Would it still be a problem even if the specific numbers were adjusted? Eg length of booster shorter, the length of the downside longer, the boost itself smaller, etc etc?
Or the price higher for them?
Plus I did suggest the two mechanics for making them worse to use as well.

WolfInfinity:
> I might be mistaken but aren't the injectors are meant only for staff positions such as EO and RO? They should be fine if so.

No, that was only for a specific permanent stat booster injector in the New-Gadgets PR. Even then EO and RO have a max cap for their stats at 60 if I remember correctly. This is a different PR for different injectors.

Kitsunemitsu:
Contrib ain't around anymore. If a new dev wants to take over, just ping me to reopen

PR #790, opened 2023-03-28
Happy teddy bear fix

Coxswain-Navigator:
## About The Pull Request
There seems to be a rare bug where happy teddy bear will kill the player regardless of whether or not they have previously worked on it. This is my attempt at fixing it. I've also decided to clean up the code in a few places, which may have been the cause of its instability. I will note that without the runtimes and only word of mouth to go off of I really don't know what was actually causing this to happen.

## Why It's Good For The Game
Dying for no reason at all can be frustrating for players.

## Changelog
:cl:
fix: fixed happy teddy bear
/:cl:


Coxswain-Navigator:
I thought so too until I couldn't replicate it at all in testing.

PR #792, opened 2023-03-29
Removes Armortype Variable

LanceSmites328:
## About The Pull Request
It was useful for if we wanted to do stuff like have Red damage hit White Armor but we've never once considered doing this intentionally and it's a gimmick that would be a pain to play against.

Removes most usage of `damage_flag` and converts it to `damage_type` checks. Most are ugly *yet functional*. 
Converts a lot of `if(damage_type == BRUTE || damage_type == MELEE || ...)` with `if(damage_type in list(BRUTE, MELEE, ...))` instead.
Removes all usage of `flag` for projectiles and `armortype` for items. We don't need it.

## Why It's Good For The Game
Redundant Code Removal makes coding things easier.

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1489,7 +1489,12 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 	hit_area = affecting.name
 	var/def_zone = affecting.body_zone
 
-	var/armor_block = H.run_armor_check(affecting, I.armortype, "<span class='notice'>Your armor has protected your [hit_area]!</span>", "<span class='warning'>Your armor has softened a hit to your [hit_area]!</span>",I.armour_penetration)
+	var/armor_block
+	switch(I.damtype)
+		if(RED_DAMAGE, WHITE_DAMAGE, BLACK_DAMAGE, PALE_DAMAGE, MELEE, BULLET, LASER, ENERGY, BOMB, BIO, RAD, FIRE, ACID)
+			armor_block = H.run_armor_check(affecting, I.damtype, "<span class='notice'>Your armor has protected your [hit_area]!</span>", "<span class='warning'>Your armor has softened a hit to your [hit_area]!</span>",I.armour_penetration)
+		else
+			armor_block = 0
 	var/Iwound_bonus = I.wound_bonus
 
 	// this way, you can't wound with a surgical tool on help intent if they have a surgery active and are lying down, so a misclick with a circular saw on the wrong limb doesn't bleed them dry (they still get hit tho)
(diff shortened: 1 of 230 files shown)


Coxswain-Navigator:
Would it be possible to simply set it equal to damage type but keep the ability to manually overwrite it?

dsmith328:
> Would it be possible to simply set it equal to damage type but keep the ability to manually overwrite it?

Every single thing that has this in it and changes damage type literally does this and has to do this every single time it changes. The only reason you WOULD is if you wanted to change what it checks against but that ONLY MATTERS for Humans as Humans are the only ones with 2 health bars (Health and Sanity) so it makes 0 difference for abnormalities and is a pain in the ass for players which is why we've NEVER done it.

Coxswain-Navigator:
> Every single thing that has this in it and changes damage type literally does this and has to do this every single time it changes. The only reason you WOULD is if you wanted to change what it checks against but that ONLY MATTERS for Humans as Humans are the only ones with 2 health bars (Health and Sanity) so it makes 0 difference for abnormalities and is a pain in the ass for players which is why we've NEVER done it.

While the argument is indeed valid, future proofing is important. I know that this won't happen, but if damage types were a significant part of fixer vs syndicate combat, just as an example this sort of thing could come into play. That being said, if it ultimately doesn't matter than I can't see any other issues here.

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1492,7 +1492,12 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 	hit_area = affecting.name
 	var/def_zone = affecting.body_zone
 
-	var/armor_block = H.run_armor_check(affecting, I.armortype, "<span class='notice'>Your armor has protected your [hit_area]!</span>", "<span class='warning'>Your armor has softened a hit to your [hit_area]!</span>",I.armour_penetration)
+	var/armor_block
+	switch(I.damtype)
+		if(RED_DAMAGE, WHITE_DAMAGE, BLACK_DAMAGE, PALE_DAMAGE, MELEE, BULLET, LASER, ENERGY, BOMB, BIO, RAD, FIRE, ACID)
+			armor_block = H.run_armor_check(affecting, I.damtype, "<span class='notice'>Your armor has protected your [hit_area]!</span>", "<span class='warning'>Your armor has softened a hit to your [hit_area]!</span>",I.armour_penetration)
+		else
+			armor_block = 0
```
EgorDinamit, line 1500: Is this needed..?
Also, this reminds me, but that's why armortype was kept - BULLET is armor, but there's no BULLET damage, so if it's every somehow used - it will break.

dsmith328: The changes I've made recently should resolve this issue via directly converting the damages to what should be a fairly relevant type, or I can convert them to BRUTE by default otherwise.

EgorDinamit (approved):
I didn't notice any big issues, but you might want to test-merge it for a while.

EgorDinamit:
Wow, we really forgot about that PR, huh? Fix conflicts please.

EgorDinamit (approved):
Let's go.

PR #800, opened 2023-04-01
Eyeball Scooper Rework

monochromatism:
## About The Pull Request

Alters the Eyeball Scooper EGO weapon to allow it to fulfill its original niche: A unique EGO that, at its best, is as powerful as its user, enabling its use for an entire round.

It now scales off of fortitude by default, with a mostly predictable, linear growth. It also has an additional scaling factor, only active at 80 fortitude and above, that helps it bridge the gap between Waw and Aleph.

Its DPS almost exactly matches that of a regular EGO one could wield at each stat threshold: It has 22 force at 40 fortitude, 33 force at 60 fortitude, 44 force at 80 fortitude, and (if the extra scaling factor is active) 80 force at 100 fortitude.

To compensate for its additional strength, the extra scaling factor only applies when Eyeball Scooper is the only EGO weapon on your person. To make this slightly less punishing within the context of LC13, the weapon will detect enemies that are black immune or that absorb black damage, and instead deal brute damage (with significantly reduced force).

Beyond 100 fortitude, Eyeball Scooper continues gaining force, but as there is only one copy of it and the weapon still requires its wielder to use no other EGO for it to exceed Aleph-tier base force, this is hopefully balanced.

The conditions under which Eyeball Scooper is acquired have also been changed; it is easier to acquire nonlethally in the early half of a round, and near impossible in the latter half, assuming you're keeping up.

## Why It's Good For The Game

The protagonist weapon deserves to actually be usable.

## Changelog
:cl:
tweak: Eyeball Scooper is worthwhile for an entire round
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/teth.dm
+++ b/code/game/objects/items/ego_weapons/teth.dm
@@ -82,6 +82,7 @@
 /obj/item/ego_weapon/eyeball
 	name = "eyeball scooper"
 	desc = "Mind if I take them?"
+	special = "This weapon grows more powerful as you do, but its potential is limited if you possess any other EGO weapons."
 	icon_state = "eyeball1"
 	force = 20
 	damtype = BLACK_DAMAGE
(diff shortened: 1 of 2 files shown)


code/game/objects/items/ego_weapons/teth.dm
```diff
@@ -82,6 +82,7 @@
 /obj/item/ego_weapon/eyeball
 	name = "eyeball scooper"
 	desc = "Mind if I take them?"
+	special = "This weapon grows more powerful as you do, but its potential is limited if you aren't loyal to it."
```
dsmith328, line 85: While this is a very nice description, it's not a good one for saying how it actually works.
```suggestion
	special = "This weapon grows more powerful as you do, but its potential is limited if you possess any other EGO weapons."
```

dsmith328: Optionally,
```suggestion
	special = "This weapon grows more powerful as you do, but its potential is limited if you aren't loyal to it, and are possessing other EGO weapons."
```

code/modules/mob/living/simple_animal/abnormality/zayin/bottle.dm
```diff
@@ -61,19 +61,56 @@
...
 
+			var/fortitude = get_attribute_level(user, FORTITUDE_ATTRIBUTE)
+			var/prudence = get_attribute_level(user, PRUDENCE_ATTRIBUTE)
+			var/temperance = get_attribute_level(user, TEMPERANCE_ATTRIBUTE)
+			var/justice = get_attribute_level(user, JUSTICE_ATTRIBUTE)
+			var/goal_damage = 0
+			if(temperance >= (fortitude + prudence + justice) / 1.5) // If your temperance is at least twice your average stat, you aren't hurt, but lose temperance.
+				to_chat(user, "<span class='userdanger'>The room is filling with water... but you feel oddly unconcerned.</span>")
+				user.adjust_attribute_level(TEMPERANCE_ATTRIBUTE, 20 - temperance)
+				// This is a PERMANENT stat change, VERY significant. But it can happen only once per round. You're The Protagonist, after all.
```
dsmith328, line 83: Permanent stat reduction that happens EARLY and it's the stat that can be regained. Not as bad as you're making it sound but also not an issue I think.

monochromatism: I know, but I wanted to make clear that I know it's far, far beyond the norm for an abnormality to affect/reduce someone's base stats rather than applying a bonus, and that I feel the degree of which is reasonable.

dsmith328 (changes requested):
Squash Commits then ping me

dsmith328 (approved):

PR #818, opened 2023-04-07
You Strong part fix

dsmith328:
## About The Pull Request
Fixes the machine breaking if you switch selected part after the do-after.


Kitsunemitsu (approved):
Looks good to me, also a needed fix.

PR #819, opened 2023-04-07
Double Tap PR

dsmith328:
## About The Pull Request
Makes weapons with attack speeds less than 0.8 attack twice per click without affecting DPS (Conflicts with Torn-Off Wings PR) then doubles the normally click cooldown.

## Why It's Good For The Game
Lets fast weapons actually use more of their DPS, still checks for the normal requirements of the weapon. Reduces arthritis build-up in users.


Kitsunemitsu (approved):
Sure, why not.

PR #824, opened 2023-04-08
Cinderella Ego Dump

ScarSkit:
## About The Pull Request

Adds an ego suit: Carriage

## Why It's Good For The Game

An ego dump for Cinderella's pumpkin Carriage


Kitsunemitsu:
You should add the changes at the end, but aside from that this is good to go!
Great first pr

ScarSkit:
ah Cox told me I did not need to add cl because the pr was too small

Coxswain-Navigator:
She is right that it is a good practice to add cl generally, but this is also such an extremely minor addition that players don't really need to be notified of it. A player might get this armor in a week or so as a nice surprise.

Kitsunemitsu (approved):
Approved.

PR #855, opened 2023-04-16
Green Ordeals Tweaks and Rebalancing

PotatoTomahto:
## About The Pull Request
Green Dawn:
- Crit finisher damage shifted to each stab (35 white total)
- Crit finisher also checks if you are out of crit
- Dawn Ordeal spawn increased from 4 to 6 (in comparison, violet dawn spawns 8)

Green Noon:
- MAJOR NERF: Ranged volley changed to 6 damage x 5 shots
- BUFF: Each shot in a volley comes out faster
- BUFF: Time between each volley reduced (doubled reload amount),
- BUFF: Bullet velocity increased
~~- BUFF: More melee damage shifted from tile effect to initial attacks~~
~~- BUFF: Recharge state reduced from 6.4 seconds to 4.8 seconds~~
- Green Noons check FF
- Noon Ordeal spawn increased from 4 to 6

Green Dusk:
- BUFF: Max health buffed from 2500 to 3000
- Visual changes, animation when producing robots

More sounds added from LobCorp for that authentic robot experience. 

Ordeal spawning was refactored to limit ordeals from spawning on top of each other. _ordealtypes.dm also rearranged for better code readability.

## Why It's Good For The Game
Green dawn's crit finisher is changed to be more consistent, with each visual attack dealing white damage.

Green noons having the potential to do up to 80 damage with their guns is absurd, and this is only magnified with multiple of them. This high-damage gun has only encouraged people to either camp doorways or corners, because trying to deal with an 80-damage ranged attack on top of a 50-damage melee is near impossible.
What this PR does is change their guns into a low-damage but quick attack that's good for taking down low-health people. This better resembles their LC behavior and means taking a step back doesn't result in your instant death. **Basically, it shifts the focus of Green Noon's damage output from their ranged to their melee**.

Due to green noons being slightly weaker and animations slowing down spawning, Green Dusk will get a health buff so it has more time to pump out robots.

Do note this is all subject to change based on player feedback. However, the purpose is to make green noons and dusk reasonably fightable without either crazy red armor or black weapons.

## Changelog
:cl:
tweak: green dawn
balance: green noon and green dusk
:cl:

PR #857, opened 2023-04-17
Green Noon Rebalance (Alternate of 855)

Kitsunemitsu:
## About The Pull Request
- It reloads more often, every 3 volleys.
- Volley reduced to 6 from 8
- Damage has been changed to be many fast attacks (with proportionally less damage) - This change isn't really necessary, I just like it better as a chainsaw.
- Green Noons from the Dusk have now patrol to not clog up halls around the factory
- Green Noons now check friendly fire.
- Green Noons no longer shoot at you when you are within spear range
- Green Dusks spawn Noons at 1/3rd rate rather than 50%

## Why It's Good For The Game
I am really not a fan of #855 and would rather it not be merged; This is my alternative PR that should make it a more engaging fight.

## Changelog
:cl:
balance: Green Noon rebalance
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/green.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/green.dm
@@ -92,18 +92,20 @@
 	base_pixel_x = -8
 	gender = NEUTER
 	mob_biotypes = MOB_ROBOTIC
-	maxHealth = 900
-	health = 900
+	maxHealth = 1100
+	health = 1100
 	speed = 3
 	move_to_delay = 6
-	melee_damage_lower = 22 // Full damage is done on the entire turf of target
-	melee_damage_upper = 26
+	rapid_melee = 4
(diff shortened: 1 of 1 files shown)


EgorDinamit:
Aside from being a huge and unwarranted nerf, rapid melee will turn the attack sounds into disgusting mess.
The "spear range" is silly and only buffs specific weapons that were already good.

EgorDinamit:
For the time being, we'll keep the DNM lables on both PRs until there's either a clear "winner" or a compromise.

Kitsunemitsu:
> Aside from being a huge and unwarranted nerf, rapid melee will turn the attack sounds into disgusting mess.
> The "spear range" is silly and only buffs specific weapons that were already good.

I can remove rapid melee. 
I added spear range for a reason, moreso that you can move a little more freely without it unloading it's mag into you than for actually buffing spears.

PR #864, opened 2023-04-19
Magical Girl Realized Abilities, Also Hats

dsmith328:
## About The Pull Request
Adds realized abilities and armors for:
KoG (already had armor)
KoD
QoH
SoW

Gold Experience:
You teleport to the nearest unseen enemy and strike them, after a short channel.

Quenched with Blood:
Fire two swords that do Pale damage equal to 2% of the target's max hp, minimum of 40 Pale, have IFF, and slow the target on hit.

Love and Justice:
Charge up a customizable ARCANA SLAVE that does 10*50 damage, scaling with justice and increasing by 50% should you deal 150+ damage with it. Has a VERY long cooldown. Heals those in critical condition: Downed (but not dead) and Insane.

Wounded Courage:
Has 3 charges before going on cooldown, 3rd charge deals +50% damage.
Smash in a 7x7 around you, dealing 30 Black damage (scales with Justice). Each hit grants you a +1 bonus to Justice, which decays over time and decays faster the more you have. Hitting the same enemies multiple times grants an additional +1 bonus for each time you've hit them.
First Hit: +1 Justice
Second Hit: +2 Justice
Third Hit: +3 Justice

Allows EGO Gear to declare hats that are only equippable via the suit, and are destroyed when the armor or hat is unequipped. These hats will drop normal hats on the ground when equipped, and are equipped solely through an ability of the EGO gear. They provide no bonuses outside of being cool.

Current EGO with Hats: Prophet, Universe, Wounded Courage

## Why It's Good For The Game
Realization good, these are cool as fuck.


code/modules/spells/ability_types/realized_aimed.dm
```diff
@@ -88,3 +88,229 @@
 		var/distance_decrease = get_dist(T, L) * 10
 		L.apply_damage((damage_amount - distance_decrease), WHITE_DAMAGE, null, L.run_armor_check(null, WHITE_DAMAGE))
 		new /obj/effect/temp_visual/revenant(get_turf(L))
+
+/obj/effect/proc_holder/ability/aimed/despair_swords
+	name = "Blades Whetted with Tears"
+	desc = "An ability that summons 2 swords to attack and slow nearby enemies. \
+		Each sword deals damage equal to 2% of the target's max HP as Pale, to a minimum of 40."
```
Coxswain-Navigator, line 95: Is this supposed to say maximum?

code/modules/clothing/head/misc.dm
```diff
@@ -489,3 +489,21 @@
 	flags_inv = 0
 	armor = list(MELEE = 30, BULLET = 15, LASER = 30, ENERGY = 40, BOMB = 25, BIO = 0, RAD = 0, FIRE = 50, ACID = 50)
 	strip_delay = 140
+
```
Coxswain-Navigator, line 492: Why not put this under ego_gear.dm? It is mostly relevant to ego gear anyways, and it could be difficult to find in its current state.

Coxswain-Navigator (changes requested):
Looks good, although I am a bit worried about how stripping might interact with the hats. Definitely in a test merge state though.

code/modules/spells/ability_types/realized_aimed.dm
dsmith328: No, it has a min damage of 40 so things with 800 hp take 40 pale instead of 16, but something with more than 2000 hp takes more. It's a realized ability and honestly doesn't tank TTK that much due to it's cooldown? iirc solely using it against an enemy with 1.0 Pale gives a TTK of about 5 minutes? Im down to tweak this if you find this egregious.

code/modules/clothing/head/misc.dm
dsmith328: Primarily because it's a hat and not directly EGO gear, I could leave a comment towards this file to point there or move it. I thought this spot fit better.

code/modules/spells/ability_types/realized_aimed.dm
Coxswain-Navigator: No, you're right. I didn't read the code all that well when I made this review. It would be interesting if the swords spawned as mobs that fight on your behalf, though.

Kitsunemitsu (approved):
Approved. was tested on the server

PR #884, opened 2023-04-27
Fourth match flame realization

EgorDinamit:
## About The Pull Request

Adds a fairly simple realization for fourth match flame EGO.

## Why It's Good For The Game

More realized armor and abilities.


code/modules/spells/ability_types/realized.dm
```diff
@@ -93,3 +93,43 @@
...
 	icon = 'icons/mob/actions/actions_ability.dmi'
 	icon_state = "judgement"
+
+/obj/effect/proc_holder/ability/fire_explosion
+	name = "Match flame"
+	desc = "An ability that deals high amount of RED damage to EVERYONE around the user after short delay."
+	action_icon_state = "fire0"
+	base_icon_state = "fire"
+	cooldown_time = 30 SECONDS
+	var/explosion_damage = 1000 // Humans receive half of it.
```
dsmith328, line 103: Only concern is that this may be a lot? I'd like to discuss how we should balance Realization EGOs and their damages at a later point

EgorDinamit: I too, thought that it may be a lot, but unlike other realizations, this one doesn't have any "auxiliary" effects.
1000 damage every 30 seconds after a formidably long delay that can also damage your team is balanced, in my opinion.

dsmith328 (approved):
I'm gonna leave this approved in case you have anything else you'd like to do with it immediately.

PR #888, opened 2023-04-29
Abnos Can Return Home

LanceSmites328:
## About The Pull Request
Allows certain abnos (P-Bird, Ppodae, Void Dream) to return to containment instead of QDEL-ing. Does this by turning their AI off, turning GODMODE on, and telling them to go the heck home. Also adds a variable that tracks the location they spawn in so they can patrol there consistently. 

After completing whatever object they aim to do (Damage, kill, or just wandering), they'll godmode, have their AI turned off, lose density, and try and patrol home. If they *can't*, they'll QDEL like they used to. After arriving, they'll become Dense, reset their Qliphoth Counter, face East, and no longer be "returning". 

Code-wise this adds:
A proc that's called when a patrol stops, either by reaching it's destination or being stopped. By default it does nothing.
A variable to _abnormality.dm that tracks starting location for the abno.
Code to abnormality.dm to set home to the abno landmark location.
Adds a conditional to IsContained() to check if they are *on* their spawn location. If not, they aren't contained (Clearly). I'm down to edit this somehow if it's shitcode.

## Why It's Good For The Game
It looks cool. Might be more trouble than it's worth but I thought it'd be a nice addition, it's something patrol code HAS given us the power to do. Biggest issue would be scaredy cat but Im making a separate PR for that issue.

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
@@ -59,6 +59,9 @@
 	var/datum/ego_gifts/gift_type = null
 	var/gift_chance = null
 	var/gift_message = null
+
+	/// Abno Landmark Location
+	var/turf/home = null
 	var/abnormality_origin = ABNORMALITY_ORIGIN_ORIGINAL
 
 /mob/living/simple_animal/hostile/abnormality/Initialize(mapload)
@@ -317,6 +320,8 @@
 
 /mob/living/simple_animal/hostile/abnormality/proc/IsContained() //Are you in a cell and currently contained?? If so stop.
 //Contained checks for: If the abnorm is godmoded AND one of the following: It does not have a qliphoth meter OR It has qliphoth remaining
+	if(!(src in home) && !isnull(home))
+		return FALSE
 	if((status_flags & GODMODE) && (!datum_reference.qliphoth_meter_max || datum_reference.qliphoth_meter))
 		return TRUE
 	return FALSE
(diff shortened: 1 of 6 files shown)


Kitsunemitsu:
Make it either not turn on god mode or be very clear if it is returning.

code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
```diff
@@ -59,6 +59,9 @@
 	var/datum/ego_gifts/gift_type = null
 	var/gift_chance = null
 	var/gift_message = null
+
+	/// Abno Landmark Location
+	var/turf/home = null
```
EgorDinamit, line 64: Obsolete. You can get that location from abno's datum reference landmark.

code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
```diff
@@ -317,6 +320,8 @@
 
 /mob/living/simple_animal/hostile/abnormality/proc/IsContained() //Are you in a cell and currently contained?? If so stop.
 //Contained checks for: If the abnorm is godmoded AND one of the following: It does not have a qliphoth meter OR It has qliphoth remaining
+	if(!(src in home) && !isnull(home))
```
EgorDinamit, line 323: Some abnos are offset from the spawning location, so this won't work.

Kitsunemitsu:
Old PR, still in draft, being closed. We have the technology I believe from a PR, it might be coming soon

PR #890, opened 2023-04-29
Cat Checks for Godmode

dsmith328:
## About The Pull Request
Cat loses courage when it's friend is godmode'd.

## Why It's Good For The Game
Allows things to become godmoded without cat being **ETERNAL**


Kitsunemitsu (approved):
Approved.

PR #891, opened 2023-04-30
Servant of Wrath and console reference retooling

dsmith328:
## About The Pull Request
Retools console reference to be in the datum so that it no longer requires a new selection every spawn.
Changes the meltdown effect to check for the abnormality's IsContained() proc instead of basically writing the same thing from scratch and not letting it be overwritten.
Makes SoW unable to be requested during meltdown.

## Why It's Good For The Game
Puts code in better places, reduces copypasta and allows meltdowns to not affect friendly breaches, and stops SoW from buggin the fuck out.


Kitsunemitsu (approved):
Looks good to me. If it causes serious issues we can revert easily.

PR #892, opened 2023-04-30
Servant of Wrath targetting fix and normal breach on random spawn

dsmith328:
## About The Pull Request
What an Embarassing Fix!
So, SoW apparently left their target search loop the moment they found even one person! Not great! They also faction checked with their faction before becoming neutral! Also not great! Both have been resolved.
SoW also now has normal breach behaviors if they're spawned and BreachEffect is called, so that's nice!

## Why It's Good For The Game
Prep for Contract change as well as fixes bugs I didn't even know existed!


dsmith328:
Honestly this fix has to be one of my biggest clown moments

Kitsunemitsu (approved):
Looks Good to me.

PR #904, opened 2023-05-03
Adds All-natural snake oil

Coxswain-Navigator:
## About The Pull Request
This pull request is two things; it replaces an old tool's old functions with a new tool, and revamps the old tool. The old tool in question is Dr. Jekyll's tincture, which players seem to be dissatisfied with according to feedback.

New Abnormality - All-Natural snake oil
Each use reduces RED defense and increases FORTITUDE. After a single use, reaching the critical state will instead cause instant death.

Revamped Abnormality - Dr. Jekyll's Tincture
All old effects removed. Now provides a 5-50 increase or decrease in stats depending on agent level and stat priority when the user's sanity falls below 50%
High stats are assigned a negative value while lower stats are assigned a positive value.
Panicking will always lead to a murder panic.
Risk level raised to HE.

## Why It's Good For The Game
These are honestly way more interesting and useful than the old effects.

## Changelog
:cl:
add: Added all-natural snake oil
tweak: revamped Dr. Jekyll's Tincture
/:cl:


dsmith328 (commented):
I merely request that Dr Jekyll's applies the Murder Panic halo status effect.

dsmith328 (approved):

PR #912, opened 2023-05-06
Naked nest fixes part2

InsightfulParasite:
## About The Pull Request
Edited targeting with what i learned from coding indigo leadership. Naked nest and its serpents will not bother themselves with nonhumans like cleanbots. This also lets me remove the Naked Nest faction since it only existed to prevent several worms from infecting the same host.
Added a goto proc on the spawned serpents so that they dont die instantly upon leaving the nest.
added a varient of the naked nest minion that will not metamorphizes into a nest until a hour has past.
Converts naked nest status effect into a removable organ

## Why It's Good For The Game
Fixes jankiness with naked nest and provides a combat subtype for special events.

## Changelog
:cl:
tweak: Naked Nest
/:cl:


Kitsunemitsu (approved):
Looks Good to me. Fix anything that breaks.

PR #930, opened 2023-05-12
Thunderbird E.G.O rework

vampirebat74:
What is says on the tin.

## About The Pull Request
This E.G.O just sucks. It's a shame, because the idea wasn't that bad. Being pushed by abnormalities or other players simply made it unusable. Furthermore, it  wasn't even that spectacular, for what was supposed to be a weapon specialized in burst damage.

This PR should change a few of those things. This will be the server's first ranged E.G.O with a charge mechanic, which I've converted the special attack into. The description should also be a bit more helpful, as I've noticed very few players knew the special attack even existed. The bug that made it misfire has also been fixed. I have also given it a minor visual improvement by creating a palette swap of its original beam projectile.

## Why It's Good For The Game

- It's good to fix things that don't work as intended.
- Should be fun to use, rather than frustrating.
- Charge is a great mechanic, and this is a weapon that fits it thematically.
- It looks a bit better

## Changelog
:cl:
tweak: feather of valor weapon now uses charge for its special attack instead of another do-after timer
tweak: do-after timer for feather of valor can no longer be interrupted with movement
fix: fixes the bug that caused the special attack to misfire
fix: removes a stray pixel from the icons
imageadd: three new beam icons
code: created charge mechanics for projectile weapons
/:cl:


Kitsunemitsu (approved):
Don't really like refinery set abilities creeping into the regular game, Still approving.

Kitsunemitsu (approved):
Been tested, well liked.

PR #942, opened 2023-05-14
Fix runtime in abno cell template loading

out-of-phaze:
## About The Pull Request
Fixes a runtime in abno cell template loading, where a bajillion copies of the template would be queued up because `new` takes time to complete.

## Why It's Good For The Game
Fewer runtimes = better. I was tired of dealing with it when testing locally.


Kitsunemitsu (approved):
One line approval. Really hope this doesnt break anything

PR #943, opened 2023-05-14
Add config option for qliphoth meltdown percentage

out-of-phaze:
## About The Pull Request
The number of abnormalities that melt down per qliphoth meltdown event is now a config option.

## Why It's Good For The Game
As discussed in `#ask-mentors`, this would be useful for local servers where you want it to be harder/easier, and for adminbus.


out-of-phaze:
This is untested btw.

EgorDinamit (approved):
The code looks fine and should work, tested or not.

PR #944, opened 2023-05-15
Fixes Duplicate Proc Definition from Double Tap PR

dsmith328:
## About The Pull Request
Discord had duplicate definition

## Why It's Good For The Game
Fixes issue

PR #951, opened 2023-05-15
Revert "Fixes Duplicate Proc Definition from Double Tap PR"

Kitsunemitsu:
Reverts vlggms/lobotomy-corp13#944

need to revert this too


dsmith328 (approved):
F

PR #957, opened 2023-05-17
Halves Violet Midnight's damage

Kitsunemitsu:
## About The Pull Request
Halves Violet Midnight's Damage.

## Why It's Good For The Game
Violet midnight could kill anyone if all 3 blades target a single person. It is extremely overtuned.
300 Damage is absolute overkill even for a midnight, when the hitbox is this big and hard to avoid.

## Changelog
:cl:
balance: Violet Midnight now deals half damage
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/violet.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/violet.dm
@@ -540,7 +540,7 @@
 
 	var/obj/effect/pale_eye/eye = null
 	var/pulsating = FALSE
-	var/pulse_damage = 15 // Dealt each second
+	var/pulse_damage = 30 // Dealt each second
 
 /mob/living/simple_animal/hostile/ordeal/violet_midnight/pale/Initialize()
 	. = ..()
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/violet.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/violet.dm
@@ -285,7 +285,7 @@
 /mob/living/simple_animal/hostile/ordeal/violet_midnight/red
 	damage_coeff = list(BRUTE = 1, RED_DAMAGE = -1, WHITE_DAMAGE = 0.7, BLACK_DAMAGE = 1.2, PALE_DAMAGE = 1)
 
-	var/attack_damage = 300 // Dealt once if hit
+	var/attack_damage = 150 // Dealt once if hit
 	var/list/been_hit = list()
 
 /mob/living/simple_animal/hostile/ordeal/violet_midnight/red/StartAbility()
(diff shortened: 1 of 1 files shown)


vampirebat74:
Worth mentioning that it lowers pale pulses from 50 to 15.

EgorDinamit:
PR for weak people with skill issue.......

EgorDinamit (changes requested):
Cutting damage in **half** is too much.
Keep in mind that it's relatively easy to dodge under most circumstances and agents are expected to wear at least 50% armor.

Kitsunemitsu:
Dying on spawn in one hit is not okay. Even at 60% armor you can take 3 hits from red and white, and 5 from black.
This ordeal does all damage types and you cannot effectively prepare WAW armor for it, which is what most people have for midnight. 
300 damage is just too much for a gigantic AOE midnight that does all 4 damage types, that has this much HP and takes this long to beat.

Sometimes multiple swords target an area in such a way where you cannot dodge one of them.
It's not fun nor fair to our players for this ordeal to constantly wipe everyone because it's damage is overtuned.

EgorDinamit:
As much as I dislike the "spawn-camping" - I feel like we need a proper "game over" condition for situations like that, instead of expecting players to brute force it by constantly respawning.
The "gigantic AOE" is on very high cooldown and easily avoidable most of the time. If damage were to be reduced, so must be their cooldown or delay before performing the attack.
When multiple pillars target same person or area - it's same issue as spawn-camping, as it usually means that everyone else is already dead.

Kitsunemitsu:
https://media.discordapp.net/attachments/982658750747926609/1122218216236793966/image.png?width=588&height=473

I witnessed this earlier, This is not okay for an instant kill attack on an ordeal that does all 4 damage types.

code/modules/mob/living/simple_animal/hostile/ordeal/violet.dm
```diff
@@ -540,7 +540,7 @@
 
 	var/obj/effect/pale_eye/eye = null
 	var/pulsating = FALSE
-	var/pulse_damage = 50 // Dealt each second
+	var/pulse_damage = 15 // Dealt each second
```
dsmith328, line 543:
```suggestion
	var/pulse_damage = 30 // Dealt each second
```
15 for pale is too big a nerf imo, primarily because this hurts one of the strongest parts way too much. Most HE+ armor has SOME pale resist and with this number you can survive 3 hits naked from this, without taking too much out of the potency of this attack.

dsmith328 (changes requested):

dsmith328:
I primarily agree with this PR based on: People getting spawned camped isn't fun and the amount of time it takes for people to 'give up' on our server is very long. People WILL respawn wave this just to finish it so they can shuttle. The main thing is this mostly gives people a CHANCE to get out of spawn. Another reason I agree with this is, whether we like it or not, part of the main handling method of this Midnight was using boxes to dodge the attacks. Did we intend for this? Don't believe so, but it was a key point. That go speed-merged out, so any actual issues this had that was being countered by box abuse is not more clear. If this needs adjusting later, we can do it again. Flipping numbers back and fourth is the song and dance we live for.

dsmith328 (approved):

PR #966, opened 2023-05-18
Zayin as first pick

Skryty-cmd:
makes the first pick always zayin. simply the best. incredible.


Kitsunemitsu (approved):
It works, looks good to me.

PR #969, opened 2023-05-19
Registers Normal resistances

Auris456852:
## About The Pull Request

It fixes the calculation so that a flat 1 is registered as Normal, as it should.

## Why It's Good For The Game

Finally, people can actually properly tell what's the best option against an abno whose resistances all are 1 or lower.

## Changelog
:cl:
fix: fixed record resistance calculations
/:cl:


Kitsunemitsu (approved):
Minor bugfix.

PR #970, opened 2023-05-19
Updates all damage resistances to be more faithful to how the original game portrays them

Auris456852:
## About The Pull Request

It fixes the other stat calculations to work like the original game. Sorry to not put it in one PR, but it was only later brought to my attention.

## Why It's Good For The Game

Here is WhiteNight's damage resistances before:
image

Here is WhiteNight's damage resistances after:
image

We can actually tell what abnormalities like WhiteNight are weakest against.
Sure, we also run into the inverse issue of Thunderbird having 3 Endured and 1 Normal, but due to how much more noticable resistances get the lower or higher you go, it's really important to highlight the really high/low ones. 

## Changelog
:cl:
fix: fixed record resistance calculations
/:cl:

### DIFF
--- a/code/__HELPERS/abnormalities.dm
+++ b/code/__HELPERS/abnormalities.dm
@@ -7,16 +7,16 @@
 			return "Normal"
 		if(-INFINITY to 0)
 			return "Absorbed"
-		if(0 to 0.5)
-			return "Resistant"
 		if(0.5 to 1)
 			return "Endured"
+		if(0 to 0.5)
+			return "Resistant"
 		if(1 to 1.5)
 			return "Weak"
-		if(2 to INFINITY)
-			return "Fatal"
 		if(1.5 to 2)
 			return "Vulnerable"
+		if(2 to INFINITY)
+			return "Fatal"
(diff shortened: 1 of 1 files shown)


code/__HELPERS/abnormalities.dm
```diff
@@ -7,16 +7,16 @@
...
 			return "Endured"
+		if(0 to 0.5)
+			return "Resistant"
 		if(1 to 1.5)
 			return "Weak"
-		if(2 to INFINITY)
-			return "Fatal"
 		if(1.5 to 2)
 			return "Vulnerable"
+		if(2 to INFINITY)
```
Kitsunemitsu, line 18: 2x damage is now only vulnerable. only higher than 2x is fatal.

Kitsunemitsu (commented):
Sadly, I don't really get the point of this PR, FATAL at 2x is a good thing for teths and HEs.

Kitsunemitsu:
Talked with the dev, you can now see exact defense values with the deepscan kits.

PR #976, opened 2023-05-21
Sephirah Update

Kitsunemitsu:
## About The Pull Request
Couple Updates to Sephirah:
1) Lets them choose between humanoid and box form, and the color of their box.
2) Their name is now automatically generated based off which department they want
3) Gives them manager access, in a minor attempt to let them fill in for a dead manager/on lowpop
4) All huds are now inherent to the job

## Why It's Good For The Game
More Sephirah changes

## Changelog
:cl:
add: Added Sephirah Box mode
tweak: minor tweaks to sephirah access
/:cl:

PR #977, opened 2023-05-21
Fix IDs disappearing upon becoming a box sephirah

out-of-phaze:
## About The Pull Request
Even if you had `nojumpsuit = TRUE` on your species, taking off your jumpsuit would cause stuff to drop. When spawning in, it would drop into nullspace. That's bad! This fixes that.

## Why It's Good For The Game
Necessary for #976.


Kitsunemitsu (approved):
Approved. Failed checks are MDB weirdness

PR #982, opened 2023-05-23
Lance fixes

Coxswain-Navigator:
## About The Pull Request
Lances no longer allow you to pull people to break the sound barrier, checks if you are actually moving for every tile, and this SHOULD Fix the exploits related to lockers and dual-wielding lances. In addition, lance charging will now knock down humans when damaging them which should reduce one-shot kills.

## Why It's Good For The Game
Bugs are bad mmk

## Changelog
:cl:
fix: fixed lances again
/:cl:


Kitsunemitsu (approved):
Fixes

PR #997, opened 2023-05-26
Replace a bunch of incorrect uses of it's with its

out-of-phaze:
## About The Pull Request
Replaces a bunch of incorrect uses of "it's" (which should only be used if "it is" would work there) with "its".

## Why It's Good For The Game
Grammar.


Kitsunemitsu (approved):
Grammar changes. Can't promise this won't happen again.

PR #999, opened 2023-05-26
Fix eyeball scooper stealing items from bags

out-of-phaze:
## About The Pull Request
OKAY SO BASICALLY
ADDING THINGS TO SOMETHING'S CONTENTS MOVES IT
ALSO LIST VARIABLES STORE REFERENCES RATHER THAN MAKING COPIES
`var/list/search_area = user.contents` MEANS ALL USES OF `search_area` ARE EQUIVALENT TO `user.contents`
`search_area |= spare_space.contents` THEREFORE IS JUST `user.contents |= spare_space.contents`
THAT MEANS IT MOVES EVERYTHING IN THE BAG INSIDE THE USER
LOL, LMAO

## Why It's Good For The Game
Fixes eyeball scooper stealing items from bags.


Kitsunemitsu (approved):
Bugfix

Kitsunemitsu:
Can't apply labels on mobile, will fix later

PR #1000, opened 2023-05-26
Gamemode Vote Redux

monochromatism:
## About The Pull Request

Gamemode votes now occur once every three rounds, and a non-Classic gamemode will only ever last one round. Two guaranteed rounds of Classic for every one time Pure might be selected!

## Why It's Good For The Game

Allows for an upcoming change to core suppressions that makes them a round-long gamemode-based thing. Also slightly less annoying to the apparently significant portion of our playerbase who doesn't notice nor care about such subtle cues as _text in chat telling them a gamemode vote has started_ or _a new verb button appearing in the top left_, and to those who get dragged into undesirable situations in spite of their due digilence thanks to them.

## Changelog
:cl:
tweak: Gamemode votes occur less and gamemode defaults to Classic when no vote occurs
/:cl:

### DIFF
--- a/code/__HELPERS/roundend.dm
+++ b/code/__HELPERS/roundend.dm
@@ -207,7 +207,20 @@
 
 	to_chat(world, "<BR><BR><BR><span class='big bold'>The round has ended.</span>")
 	log_game("The round has ended.")
-	SSvote.initiate_vote("gamemode", "automatic gamemode selection vote")
+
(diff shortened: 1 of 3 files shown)


Kitsunemitsu:
Also please remove the afk voted being counted at all

Kitsunemitsu (approved):
Approved.

PR #1010, opened 2023-05-28
Fix patrol_reset messing with target chase AI

out-of-phaze:
## About The Pull Request
Checked `patrol_path` instead of `length(patrol_path)` and `patrol_reset()` set it to `list()` instead of `null`. Replaced it with a length check and it works now.

## Why It's Good For The Game
Fixes hostile mobs doing weird accidental strafing while chasing targets.


Kitsunemitsu (approved):
Approved

PR #1012, opened 2023-05-29
Raises max PE on blubbering toad

Coxswain-Navigator:
## About The Pull Request
I forgot to update this when making the abnormality, blubbering toad now has a maximum PE of 10 instead of 6.

## Why It's Good For The Game
It has no reason to have such a low amount of PE per work.

## Changelog
:cl:
tweak: blubbering toad now awards 10 PE instead of 6 maximum
/:cl:


Kitsunemitsu (approved):
Approved

PR #1020, opened 2023-06-02
Ordeal Pets 1.1

InsightfulParasite:
## About The Pull Request
Makes ordeal pets more fleshed out and useful.
Still its a big responsibility to own a pet so you should consider their strengths and what could obliterate them before bringing them here.
Amber and Crimson Ordeal Pet Crafting
Crimson will now do a little trick when touched with a bare hand. This will restore 5 sanity to everyone in range. This is more helpful for when your in a department recovering your sanity.
Green can now have a flashlight attached to them.

Oh i should mention that i also edited the Tamed Proc in Hostiles because something bizzare was going on with that. Friends is a list but tamed converts it into one person instead of adding the person to the list.

## Why It's Good For The Game
Fleshes out ordeal pets and makes them less janky. Not a important PR.

## Changelog
:cl:
tweak: Ordeal Pets Amber, Green, and Crimson.
/:cl:


Kitsunemitsu (approved):
Looks good to me.

LanceSmites328 (changes requested):

LanceSmites328 (approved):

LanceSmites328 (approved):

PR #1024, opened 2023-06-04
Fix duplicate 'Misc' crafting tab

out-of-phaze:
## About The Pull Request
Fixes a duplicate 'Misc' crafting tab because CATEGORY_MISC was used as a food subcategory instead of CATEGORY_MISCFOOD.

## Why It's Good For The Game
Bug.


Kitsunemitsu (approved):
Bugfix

PR #1040, opened 2023-06-10
Simple Fishing Version 1 because TG Fishing is too hard to port for me.

InsightfulParasite:
## About The Pull Request
Makes a fishing rod, new aquarium, new water sprites for both water and mermaid abnormality water since the current mermaid water was a placeholder from google and we should problobly not have that around for too long.
Admiring a aquarium will now increase/decrease your sanity based on how many living/dead fish are in the tank and what rarity they are.
Most of the aquarium and fish code came from https://github.com/tgstation/tgstation/pull/67691/files#diff-90e82aeed15d4706af23d454584c04d92557e16c15b1efcb397bcea83e614302
I did use the latest version of this code so it may be much different from that initial commit.
image

ISSUES:
- [ ] Fish that are created inside the aquarium rather than placed inside dont proc the "Entered" proc which adds them to the tracked fish. Fish that are born or prefilled aquariums will not contribute to the admire sanity bonus unless taken out and placed back in.
- [ ] Fishing is not a component and right now fully relies on the object being a fishing rod and the target being deep water turf. This means it cant be applied to structures like puddles.

## Why It's Good For The Game
makes maps more unique with certain environmental activities.

## Changelog
:cl:
add: ramshackle fishing
add: water tiles
tweak: water sprites
tweak: mermaid water
/:cl:


InsightfulParasite:
Im getting these Linter Warnings:
Warning: macro redefined: COMSIG_AQUARIUM_SURFACE_CHANGED
Warning: macro redefined: COMSIG_AQUARIUM_FLUID_CHANGED
I have no clue how to fix them.

Kitsunemitsu (approved):
Nothing here should break, It doesn't runtime, gonna merge it.

PR #1055, opened 2023-06-15
Revolutionary Rat Weapon

BlueHNT:
## About The Pull Request
Here lays your revolutionary neighbor disposal device, we call it A BRICK TO THE FACE! **Now with extra damage when you throw it!!!!!**
But it's only 1 times use when thrown, so use it well.

## Why It's Good For The Game
It's a brick and you can throw it. Also you can put it on the RP map for people to mess around with.

## Changelog
:cl:
add: Added a brick
/:cl:


Kitsunemitsu (approved):
Ship it.

PR #1059, opened 2023-06-16
Forging: Ordeal Parts -> Materials

LanceSmites328:
## About The Pull Request
Allows the conversion of Ordeal parts (The things they drop when they gib/butcher) into materials for the Tres Workshop on E-Corp to utilize.
Current potential effects:
Increasing damage by a flat amount.
Increasing damage by a %.
Increasing attack speed by a %.
Changing weapon color.
Changing weapon hitsound.
Changing weapon damage type(?)
Changing weapon... edibility?

In the distillation process, if enough parts are used, one can get higher quality materials that provide a +10/20/40% bonus to force or attack speed, depending on material.

TODO:

- [x] Fix the Ordeal Extractor consuming all stacks of an item at once instead of just 1/10/20/30 at a time. (Why does this happen...!?)
- [x] Decide whether damage types should be overridden by materials.

## Why It's Good For The Game
Gives fixers a way to upgrade their gear in an interesting matter that requires effort in going out to grab the parts and come back... or potentially scavenging the remains of what others had conquered.


Kitsunemitsu (approved):
Approving now.
Will need some preliminary testing, but overall looks good.

Kitsunemitsu (approved):
re-approving

PR #1082, opened 2023-06-23
Core Suppression - Forma de Gamemode

monochromatism:
Core Suppression Gamemode time!

## About The Pull Request

Core Suppression is now a gamemode players can vote on, rather than an event that managers can choose to do after a normal round ends. You start with 80 in all stats and have a (sort of) random WAW EGO armor in hand, but the only abnormalities able to spawn are WAW and up, and of course, the core suppression begins immediately. Rather than being tied to meltdowns, ordeals also come based solely on time. Once every fifteen minutes.

## Why It's Good For The Game

Core suppressions barely occur in the regular course of the game. This will hopefully make them a little bit more accessible, and therefore make them actually happen.

## Changelog
:cl:
add: Core suppression gamemode
del: Core suppression event at end of normal shift
/:cl:


EgorDinamit (changes requested):
I'll just leave it here:
I am fine with having it as an option for gamemode, but I want to keep it in normal rounds.
I am not a big fan of having more than a single gamemode(events not included) here.

EgorDinamit:
> Core suppressions barely occur in the regular course of the game.

Happens all the time when people don't fail miserably.

EgorDinamit:
The ordeal time lock is also infuriatingly high. Core Suppression isn't meant to take longer than 20-30 minutes.
It *can* last entire round, but without white ordeals, as a modifier, which you didn't do here.

code/controllers/subsystem/lobotomy_corp.dm
```diff
@@ -189,6 +185,16 @@ SUBSYSTEM_DEF(lobotomy_corp)
...
 
+/datum/controller/subsystem/lobotomy_corp/proc/MeltdownOrdealCheck()
+	var/ran_ordeal = FALSE
+	if(qliphoth_state + 1 >= next_ordeal_time) // If ordeal is supposed to happen on the meltdown after that one
+		if(istype(next_ordeal) && ordeal_timelock[next_ordeal.level] > world.time) // And it's on timelock
+			next_ordeal_time += 1 // So it does not appear on the ordeal monitors until timelock is off
+	if(qliphoth_state >= next_ordeal_time)
+		if(OrdealEvent())
+			ran_ordeal = TRUE
+	return ran_ordeal
```
EgorDinamit, line 196:
```suggestion
/datum/controller/subsystem/lobotomy_corp/proc/MeltdownOrdealCheck()
	if(qliphoth_state + 1 >= next_ordeal_time) // If ordeal is supposed to happen on the meltdown after that one
		if(istype(next_ordeal) && ordeal_timelock[next_ordeal.level] > world.time) // And it's on timelock
			next_ordeal_time += 1 // So it does not appear on the ordeal monitors until timelock is off
	if(qliphoth_state >= next_ordeal_time)
		return OrdealEvent()
	return FALSE
```

code/controllers/subsystem/lobotomy_corp.dm
```diff
@@ -222,6 +228,25 @@ SUBSYSTEM_DEF(lobotomy_corp)
 		return meltdown_occured
 
 /datum/controller/subsystem/lobotomy_corp/proc/RollOrdeal()
+	if(GLOB.master_mode == "suppression")
```
EgorDinamit, line 231: Very ugly, but it works, I guess?
Would really prefer if you could find a way to shove it all into gamemode's code instead.

code/controllers/subsystem/lobotomy_corp.dm
```diff
@@ -248,3 +273,6 @@ SUBSYSTEM_DEF(lobotomy_corp)
 	next_ordeal = null
 	RollOrdeal()
 	return TRUE // Very sloppy, but will do for now
+
+/datum/controller/subsystem/lobotomy_corp/proc/FailSuppression()
```
EgorDinamit, line 277: Doesn't have to be a proc in LC subsystem. Move it elsewhere, since it's just a timed round-end.

code/modules/jobs/job_types/agent.dm
```diff
@@ -81,16 +81,46 @@
...
+			/obj/item/clothing/suit/armor/ego_gear/totalitarianism,
+			/obj/item/clothing/suit/armor/ego_gear/intentions
+		)
+		switch(rand(1,3))
+			if(1)
+				chosen_suit = pick(red_armor_options)
+			if(2)
+				chosen_suit = pick(white_armor_options)
+			if(3)
+				chosen_suit = pick(black_armor_options)
```
EgorDinamit, line 109: So, you are just giving people random armor..?
I feel like it should be done differently, lest people spam-respawn to get what they want.

PR #1084, opened 2023-06-24
Tweaks to range/view procs

EgorDinamit:
## About The Pull Request

Optimizes livinginview() and fixes livinginrange().

## Why It's Good For The Game

The "living in view" proc would loop through mobs, each time generating a new list by calling view(dist, center), despite only needing to do so once. That's what I "optimized".
The "living in range" used get_dist(), which, as far as I am aware, ignored z level. I added a z level check to fix potential issues.


Kitsunemitsu (approved):
Code cleanup.

PR #1087, opened 2023-06-25
Time-related and other QOL changes

EgorDinamit:
## About The Pull Request

A small experimental PR that adjusts things related to time locks and such.

- Qliphoth counter(how many works it takes for meltdown) calculation adjusted slightly. Each player will increase it by 0.65(instead of 0.5), but starts at minimum of 3 instead of 4.
- Qliphoth meltdown announcement will now say what next ordeal will be, if it happens on next meltdown.
- Currently running ordeals are stored in a list in SSlobotomy_corp. This is used in "!tgs check" Discord command.
- Some cases of world.time usage have been replaced with ROUNDTIME define, which is adjusted by round-start time.

## Why It's Good For The Game

- Makes highpop slightly more bearable. On the other hand, lowpop will get a bit more demanding.
- QOL.
- Nice way of knowing what is really going on in the round before joining.
- World.time ticks before round has started, making some time checks incorrect. This fixes it.

### DIFF
--- a/code/controllers/subsystem/lobotomy_corp.dm
+++ b/code/controllers/subsystem/lobotomy_corp.dm
@@ -165,9 +165,9 @@ SUBSYSTEM_DEF(lobotomy_corp)
 
 /datum/controller/subsystem/lobotomy_corp/proc/QliphothEvent()
 	// Update list of abnormalities that can be affected by meltdown
-	if((ZAYIN_LEVEL in qliphoth_meltdown_affected) && world.time >= 20 MINUTES)
+	if((ZAYIN_LEVEL in qliphoth_meltdown_affected) && ROUNDTIME >= 20 MINUTES)
 		qliphoth_meltdown_affected -= ZAYIN_LEVEL
-	if((TETH_LEVEL in qliphoth_meltdown_affected) && world.time >= 50 MINUTES)
+	if((TETH_LEVEL in qliphoth_meltdown_affected) && ROUNDTIME >= 50 MINUTES)
 		qliphoth_meltdown_affected -= TETH_LEVEL
 	qliphoth_meter = 0
 	var/abno_amount = all_abnormality_datums.len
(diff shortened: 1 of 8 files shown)


code/controllers/subsystem/lobotomy_corp.dm
```diff
@@ -46,9 +46,11 @@ SUBSYSTEM_DEF(lobotomy_corp)
 	// What ordeal level is being rolled for
 	var/next_ordeal_level = 1
 	// Minimum time for each ordeal level to occur. If requirement is not met - normal meltdown will occur
-	var/list/ordeal_timelock = list(10 MINUTES, 25 MINUTES, 45 MINUTES, 60 MINUTES, 0, 0, 0, 0, 0)
+	var/list/ordeal_timelock = list(10 MINUTES, 20 MINUTES, 35 MINUTES, 50 MINUTES, 0, 0, 0, 0, 0)
 	// Datum of the chosen ordeal. It's stored so manager can know what's about to happen
 	var/datum/ordeal/next_ordeal = null
+	/// List of currently running ordeals
+	var/list/current_ordeals = list()
```
dsmith328, line 53: "I wonder how the server is going? !tgs check"
```
Abnormalities in the facility: 23
Current Ordeals: Steel Noon, Green Dusk, Violet Midnight.
Join the round!
```
"Oh."

Kitsunemitsu:
I like a lot of this, except the abno spawn time.
Players have a hard time catching a breather right now.

Kitsunemitsu (dismissed):
Abno spawn time has caused serious issues regarding balance and respawn stats.
Going to request a reversion of that and the faster overall stat gain because those two are a pair.

EgorDinamit:
I'll just adjust respawn stats to follow the new timings. So far it's the only issue I've seen with faster rounds, as things usually go pretty well as long as there's no need for respawn wave.

EgorDinamit:
Stuff related to timing has been reverted.

LanceSmites328 (approved):

PR #1094, opened 2023-06-27
Pyg fix2

EnderMage99:
## About The Pull Request

This fixes the bug where the sculptor did not regain their prudence after going insane.

## Why It's Good For The Game

Bug fixes are good for the game's health.

## Changelog
:cl:
fix: fixed Pygmalion sculptor not regaining prudence
/:cl:


code/modules/mob/living/simple_animal/abnormality/waw/pygmalion.dm
```diff
@@ -128,6 +128,10 @@
 	remove_status_effect(STATUS_EFFECT_SCULPTOR)
 	threat_level = WAW_LEVEL
 	sculptor = null
+	if (sculptor)
+		sculptor.remove_status_effect(STATUS_EFFECT_SCULPTOR)
+	if (missing_prudence)
+		restorePrudence()
```
LanceSmites328, line 134: So uh, none of this will work because sculptor is set to "null" before this occurs

PR #1099, opened 2023-06-29
Doomsday Calendar Adjustments 2: Electric Boogaloo

vampirebat74:
A threat you cannot ignore

## About The Pull Request

Currently, players have 2.5 minutes to suppress the abnormality following its breach, and the abnormality itself is immobile. With these in mind, there is currently no penalty for simply ignoring the breach and moving to a different department. This PR will double the self-damage caused by Doomsday Calendar's black AOE from 500 to 1,000, reducing its maximum activations from 5 to 3. Since it has 2012 maximum health, only two will be possible after dealing 12 damage to it. However, this PR will make it more threatening by causing its radial BLACK damage AOE to reduce qliphoth counters. 

Additionally, a highly visible warning will be provided in the event of a breach. While it may be a bit flashy, I believe that it is necessary for an abnormality of a rather low risk level and tendency to breach when ignored.

## Why It's Good For The Game

In a playtest, I was able to prevent this wearing nothing but fourth match flame, sorority, and 60 fortitude & justice. The breach announcement will also make its suppression much easier to co-ordinate with other players.

## Changelog
:cl:
balance: changes doomsday calendar's penalty for failing to suppress it, makes its breach more obvious.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/doomsday_calendar.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/doomsday_calendar.dm
@@ -180,6 +180,7 @@
 	SpawnAdds()
 
 /mob/living/simple_animal/hostile/abnormality/doomsday_calendar/proc/AnnounceBreach()
+	addtimer(CALLBACK(GLOBAL_PROC, .proc/show_global_blurb, 5, "The day of the apocalypse has arrived.", 25))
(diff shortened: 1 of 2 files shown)


LanceSmites328 (approved):
No complaints, TM'd for a while. SEND IT

PR #1104, opened 2023-06-30
Parry Fix

dsmith328:
## About The Pull Request
Oversight when I wrote the code causes parry to not come off cooldown if you failed to block anything. This fixes that.

## Why It's Good For The Game
Bug fix.


Kitsunemitsu (approved):
Approved.

PR #1105, opened 2023-06-30
Fixes Yin

dsmith328:
## About The Pull Request
Something removed a single line of code that was supposed to put Yin/Yang IN the dragon. This re-adds that. It further makes it so the dragon targets the area AROUND where the spawned (with some deviation).

## Why It's Good For The Game
Bug fix and accuracy fix.


Kitsunemitsu (approved):
Important fix

PR #1106, opened 2023-07-01
Shields had more issues

dsmith328:
## About The Pull Request
So apparently the hand checking did not work and caused runtimes that stopped block from clearing properly, so the variable would remain TRUE while everything else would fix properly. Honestly no idea why this occurred but better options have been located.

## Why It's Good For The Game
Fiiiiiix. I stg Im testing these locally and not getting these issues, they only happen on the server.


Kitsunemitsu (approved):
Fix

PR #1108, opened 2023-07-01
Arbiter update

EgorDinamit:
## About The Pull Request

Adds missing features of the extraction core suppression. Work in progress.

- Added spikes. They spawn under and around all human players on z-level every 1.5 minutes. After 3 seconds they deal 100 black damage to those on their tile.

## Why It's Good For The Game

Extraction core suppression will finally be complete.


code/modules/suppressions/extraction.dm
```diff
@@ -465,3 +470,60 @@
...
+	name = "dark spikes"
+	desc = "A weird construction. You should probably stay away from it..."
+	icon_state = "binah_spike"
+	alpha = 0
+	anchored = TRUE
+	density = FALSE
+	/// How much time must pass after Initialize() to activate
+	var/activation_delay = 2 SECONDS
+	/// Amount of BLACK damage done on activation
+	var/damage = 150
```
LanceSmites328, line 507: Leaving a note that this may be a bit too much damage? Base Lobotomy says 33-40 damage, which should be ~82-100 damage for us. (We tend to do like, 2.5x from what I recall).

LanceSmites328 (approved):

LanceSmites328 (approved):
If damage proves too high it can be changed later.

PR #1110, opened 2023-07-02
HitInTurf proc

LanceSmites328:
## About The Pull Request
Adds new mob proc "HitInTurf" which allows the mob to hit all targets in a specified turf for a specified amount of damage. This comes with a good number of options to allow stuff like structure destruction/degradation, as well as the ability to pierce through stuff like lockers and crates. 

This PR changes all current applicable instances of `for(var/mob/living/L in T)` to `HurtInTurf(...)`.

## Why It's Good For The Game
We've all seen `for(var/mob/living/L in T)` copied EVERYWHERE to apply damage, this condenses that into a singular proc with support for hit-list retention. Overall reduces copy-pasta required to just make everything in a turf get slapped.


Kitsunemitsu:
Fix merge issues and then I'll merge

Kitsunemitsu (approved):
Was tested, code looks fine.
Merging

PR #1116, opened 2023-07-05
Fixes oversight in IsContained proc

PotatoTomahto:
## About The Pull Request
IsContained now also checks if an abnormality has no qliphoth but is unable to breach in the first place.

## Why It's Good For The Game
Should fix nonbreachers moving.


LanceSmites328:
If this is for the abnos moving rarely while in their room, I don't believe this solves the issue, as they have Qliphoth counters with Der and Old Lady

EgorDinamit (approved):

PR #1119, opened 2023-07-07
Reverberation Scythe

LanceSmites328:
## About The Pull Request
Adds Argalia's Scythe. Also lets items turn off their item_attack_animation via a variable, because we could do this before we just never had the variable for it.

## Why It's Good For The Game
Agony Scythe. Pain blade. Vibrating for your demise.

- [x] Alternative VFX for heavy attacks.


InsightfulParasite:
The variable seems helpful since it would allow weapons to have some visual dynamic between strong and weak attacks.

Kitsunemitsu (approved):
Looks good to me.

PR #1127, opened 2023-07-09
Naked Nest Organ Check Fix

InsightfulParasite:
## About The Pull Request
So i was alerted to the fact that naked nest would not check if you already had a naked nest organ inside you when infecting you. This was due to me not defining where to check if the organ is. Naked Nest was basically checking itself for a organ it was infecting someone else with. Its a bit embaressing since every other check is correct except for the work checks.

## Why It's Good For The Game
Fixes Naked Nest.

## Changelog
:cl:
fix: naked nest
/:cl:


Kitsunemitsu (approved):
Small fix

PR #1138, opened 2023-07-12
The Abominable Infestation (Maroon ordeals)

EgorDinamit:
## About The Pull Request

Adds the Abominable Infestation from [Tegu Station](https://github.com/vlggms/tegustation-bay12).

It is an admin-only ordeal type for now, as I try and figure out the balance and if it's worth being in rotation at all.

Work in progress.

## Why It's Good For The Game

Wanted to do it for a long time and more ordeals is always good.


Coxswain-Navigator:
If we don't use them as an ordeal we can always use more trash mobs for city maps.

PR #1133, opened 2023-07-11
Updates clown smiling at me

Coxswain-Navigator:
## About The Pull Request
Fixes an obscure bug related Clown smiling at me that allows you to trap it between two corpses by allowing it to target dead people. This is technically a buff to the abnormality, but oh well. 

Finally gives it some proper sprites. 

## Why It's Good For The Game
I think fixing bugs and visuals is good overall

## Changelog
:cl:
fix: Clown smiling at me can now butcher corpses, instead of standing idle
imageadd: Clown smiling at me now actually has its own sprites! Rejoice!
/:cl:

PR #1139, opened 2023-07-13
Fixes clown smiling at me getting stuck on mob corpses

Coxswain-Navigator:
## About The Pull Request
Fixes a bug introduced in https://github.com/vlggms/lobotomy-corp13/pull/1133 whereupon the abnormality could get stuck when attacking a non-human corpse. Technically a buff, but shouldn't cause any real issues as it fixes a rather critical bug.

## Why It's Good For The Game
Fixes a bug caused by a bugfix, classic.

## Changelog
:cl:
fix: the clown can now gib corpses
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/waw/clown_smiling.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/waw/clown_smiling.dm
@@ -86,10 +86,17 @@
 		icon_state = icon_aggro
 
 //Execution code from green dawn with inflated damage numbers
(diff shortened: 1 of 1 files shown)


LanceSmites328 (approved):

LanceSmites328 (changes requested):
Rescinding approval. If you want it to not attack non-human corpses, alter the CanAttack proc. This will make it not even target them.

LanceSmites328 (approved):

PR #1143, opened 2023-07-14
adjust templates for random rooms

Retrino:
## About The Pull Request

changed the names of the random room templates, as well as the layout a bit in the templates themselves. hopefully this will make creating a new random room more clear. 

Screenshot (183)

templates originally had door markings on both sides, which made it hard to tell at a glance which side was meant for the entrance to be placed on

(also minor changes to the robot castle room. it should no longer be airless/depressured now)

## Why It's Good For The Game

allows mappers less familiar with the layouts to more clearly see how their random room should be laid out


LanceSmites328 (approved):

LanceSmites328:
Requesting review by @Kitsunemitsu due to this being her primary domain as of late.

Kitsunemitsu (approved):
Template changes

PR #1144, opened 2023-07-15
Landmine tweaks

Trilbyspaceclone:
## About The Pull Request

Landmines now dont trigger by items being thrown over them (You can still shoot them)
Landmines now only cross trigger by mobs
If a landmine is dense (none are by default) and something throws and hits it, it now triggers
## Why It's Good For The Game

Light sources, now dont trigger landmines anymore, woo.

## Changelog
:cl:
tweak: Landmines now dont trigger by items being thrown over them and now only mobs when crossing them will trigger.
/:cl:

### DIFF
--- a/code/game/objects/effects/mines.dm
+++ b/code/game/objects/effects/mines.dm
@@ -43,9 +43,16 @@
 	if(AM.movement_type & FLYING)
 		return
 
+	if(!ismob(AM))
+		return
+
 	triggermine(AM)
 
 /obj/effect/mine/take_damage(damage_amount, damage_type, damage_flag, sound_effect, attack_dir)
+	. = ..()
+	triggermine()
+
+/obj/effect/mine/hitby()
 	. = ..()
 	triggermine()
 


code/game/objects/effects/mines.dm
```diff
@@ -43,12 +43,19 @@
 	if(AM.movement_type & FLYING)
 		return
 
+	if(!ismob(AM))
```
Kitsunemitsu, line 46: I would still like the ability to throw things on them, landmines currently are only used in one spot and there should be a way to disarm them safely (no one has guns on that map)

Kitsunemitsu (changes requested):

code/game/objects/effects/mines.dm
Trilbyspaceclone: Would it be ok if the landmine gets disarmed via clerk ID?

InsightfulParasite: You could check the weight of the item since that is a value, or have it so if the thing is currently being thrown it doesnt proc.

Trilbyspaceclone: Id want the reverse of the throwing idea, so that if an item lands on the tile it goes kaboom, otherwise dosnt.

InsightfulParasite: Thats a bit more tricky. Chasms find a way to make things inside it fall because it keeps track of the objects. So for this it would have to be crossed then start processing to see if it stays for more than just 1 decisecond if it has the throwing variable to true.

Trilbyspaceclone: Thank you for the insight! I'll have to look into the chasms code and see if i can use that for the goal.

InsightfulParasite: Check out code/datum/component/chasm component code. You may see something i missed.

> /datum/component/chasm/proc/droppable(atom/movable/AM)
> 	// avoid an infinite loop, but allow falling a large distance
> 	if(falling_atoms[AM] && falling_atoms[AM] > 30)
> 		return FALSE
> 	if(!isliving(AM) && !isobj(AM))
> 		return FALSE
> 	if(is_type_in_typecache(AM, forbidden_types) || AM.throwing || (AM.movement_type & (FLOATING|FLYING)))
> 		return FALSE

Trilbyspaceclone: This is exactly what I needed, thank you much so!

PR #1147, opened 2023-07-15
PE Has a Purpose

LanceSmites328:
## About The Pull Request
This introduces the following changes, I'll go by category:
Lobotomy Subsystem:

- Added `goal_boxes` variable, a variable designed to track unspendable PE.
- Added `total_generated` and `total_spent` variables, designed to track exactly what they say.
- Renamed `AdjustBoxes` to `AdjustAvailableBoxes` and added an `AdjustGoalBoxes` proc that redirects toward the prior if goal has been reached.
- Moved the PE goal checking function to a separate proc so it could be called by both above procs without copy-pasting code.
- `box_goal` is no longer defaulted to INFINITY, instead it's defaulted to 0 and anything checking for PE Quota will return an in-progress message if it is 0.
- `available_boxes` is no longer capped by `box_goal`.

Cargo Console:

- Shows both Available PE (The stuff you can spend) and progress toward quota, or a congratulatory message if met.

Round End:

- Added a PE Generation report that shows how much PE was generated, how much was spent, how much was made through alternative methods (such as cargo), and if PE Quota was reached.

Refinery (Sales):

- Sales machines now generate 100 PE, aimed towards `goal_boxes`. If Quota is reached, this PE is instead spendable.

## Why It's Good For The Game
The statistics are nice.
Overall helps make refining PE actually progress the round, instead of bricking Quota. Also generates excess if you've already completed the requirements. 
Otherwise, it's presentation.

### DIFF
--- a/ModularTegustation/tegu_items/records/ordealcheck.dm
+++ b/ModularTegustation/tegu_items/records/ordealcheck.dm
@@ -26,8 +26,13 @@
 
 
 	//PE stuff. He doesn't really need to know this but information and all that
-	if(SSlobotomy_corp.box_goal != INFINITY)	//Make sure it's not infinity
-		to_chat(user, "<span class='notice'>PE Goal: [SSlobotomy_corp.current_box] / [SSlobotomy_corp.box_goal].</span>")
+	if(SSlobotomy_corp.box_goal != 0)	//Make sure it's not infinity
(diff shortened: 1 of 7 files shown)


Kitsunemitsu (approved):
Code read fine, approving.

PR #1151, opened 2023-07-16
Qliphoth Unitily Abnormality Containment Kit

Trilbyspaceclone:
## About The Pull Request

New kit for abnormality pre-suppression, basically this item allows a clerk to extend the meltdown time of a meltdown by 50%
or if an abnormality is down qliphoth to restore it by 1.
No rng way to get this kit nor any on any map, tho for 200 you can get it in the cargo console
After use the item destroys itself 

## Why It's Good For The Game

This allows for bigger maps to have more active clerks running around buying time or trying to help restore containment units at a cost.

I dont mind any suggestions to help balance as im still getting the hang of things

## Changelog
:cl:
add: New kit for abnormality pre-suppression, basically this item allows a clerk to extend the meltdown time of a meltdown by 50% or if an abnormality is down qliphoth to restore it by 1.
/:cl:

Currently untested thus leaving as a draft


Gboster-0:
i don't think this is a good idea, a lot of abnormalities depend on their qliphoth meter for gimmicks and special effects. And raising it via this item could break a lot of stuff for something that is worth so little PE

code/game/objects/items/clerks/containment_kit/containment_kit.dm
```diff
@@ -0,0 +1,61 @@
...
+	if(!clerk_check(user))
+		to_chat(user,"<span class='warning'>You don't know how to use this.</span>")
+		return
+	switch(mode)
+		if(1)
+			mode = 2
+			to_chat(user, "<span class='notice'>The kit will now be able to increase a meltdown timer by [meltdowntimer_increase].</span>")
+		if(2)
+			mode = 1
+			to_chat(user, "<span class='notice'>The kit will now restore if possable a Qliphoth Count on a abnormality.</span>")
```
Gboster-0, line 44: "The kit will now restore if **possable** a Qliphoth Count on a abnormality."

Trilbyspaceclone:
> i don't think this is a good idea
The whole idea of it or just the part were it can raise Qcounter?
I dont mind cutting the Qcounter raising or changing the price, but I thought the meltdown timer increase was a non-harmful idea

Kitsunemitsu (changes requested):
Change this into an RO watch with a 10 minute timer

Coxswain-Navigator:
This probably shouldn't be usable on ALEPH abnormalities.

Trilbyspaceclone:
Apond request by [Kitsunemitsu](https://github.com/Kitsunemitsu), the Qcounter changes are being shifted to a watch for the Research Overseer on a 10 min cooldown.

Kitsunemitsu:
> This probably shouldn't be usable on ALEPH abnormalities, or at the least you could make it separated by risk level.

We're gonna give it a small test.

ModularTegustation/tegu_items/records/timestop.dm
```diff
@@ -1,23 +1,14 @@
-//Timefreeze records watch
+//Golden Watch
+//Casts Timestop for 10 seconds
```
Kitsunemitsu, line 2: It's 11 seconds actually. Small easter egg is that it's the same amount of time that dio stops time in Joko part 3

code/game/machinery/computer/abnormality_work.dm
```diff
@@ -304,3 +314,61 @@
 
 /obj/machinery/computer/abnormality/tutorial/start_work(mob/living/carbon/human/user, work_type, training = TRUE)
 	..(user, work_type, training = training)
+
+//Fancy Stopwatch Intractions
```
Kitsunemitsu, line 318: Once more, I'd prefer a use in hand near it.

code/game/objects/structures/crates_lockers/closets/secure/security.dm
```diff
@@ -301,14 +301,16 @@
...
 	name = "\proper record officer's locker"
 	req_access = list(ACCESS_COMMAND)
 	icon_state = "records"
 
 /obj/structure/closet/secure_closet/record/PopulateContents()
 	..()
-	new /obj/item/records/timestop (src)
-	new /obj/item/records/information (src)
-	new /obj/item/records/abnodelay (src)
+	new /obj/item/records/timestop(src)
```
Kitsunemitsu, line 311: I've got to fix this later; for now it works.

Kitsunemitsu (changes requested):
Small changes

Kitsunemitsu:
Okay, so thinking on it, some abnormalities can only lower their counters X amount of times, It's for the best if we scrap the qliphoth counter increase watch

Trilbyspaceclone:
Want me to just disable the watch or fully remove the code?

PR #1166, opened 2023-07-22
Adds crimson midnight

Coxswain-Navigator:
## About The Pull Request
Adds a new ordeal, crimson midnight
"Chorus of saliva" - A new unit that doesn't follow the same theme as its predecessors. It constantly spawns variants of the weaker clowns that explode after 45-60 seconds. If not killed, they will spawn tiny clowns that will breach abnormalities.
"Pinnacle of thew" - A new, larger clown that splits into two crimson dusks when killed. Instead of having a fancy special attack, it just punches you hard enough to throw you a couple tiles.

## Why It's Good For The Game
Maybe it will mix things up, or at least be difficult enough for people to complain really loudly.

## Changelog
:cl:
add: Added crimson midnight
/:cl:


code/modules/mob/living/simple_animal/hostile/ordeal/crimson.dm
```diff
@@ -283,3 +283,232 @@
...
+		return
+	visible_message("<span class='danger'>[src] suddenly explodes!</span>")
+	for(var/turf/L in view(4, src))
+		if(prob(25) && !(L.density))
+			new /obj/item/food/meat/slab/crimson (get_turf(L))
+		var/obj/effect/decal/cleanable/blood/B = new /obj/effect/decal/cleanable/blood(get_turf(L))
+		B.bloodiness = 100
+	for(var/mob/living/L in view(8, src))
+		if(!faction_check_mob(L))
+			L.apply_damage(700, RED_DAMAGE, null, L.run_armor_check(null, RED_DAMAGE))
```
LanceSmites328, line 362: holyshit

code/modules/mob/living/simple_animal/hostile/ordeal/crimson.dm
```diff
@@ -283,3 +283,232 @@
...
+	icon_dead = "crimson_midnight"
+	faction = list("crimson_ordeal")
+	maxHealth = 4000
+	health = 4000
+	pixel_x = -16
+	base_pixel_x = -16
+	melee_damage_lower = 48
+	melee_damage_upper = 56
+	move_to_delay = 4
+	ranged = TRUE
```
LanceSmites328, line 380: No Open Fire?

code/modules/mob/living/simple_animal/hostile/ordeal/crimson.dm
LanceSmites328, line 1: Might be better for the crimsons to have a variable that says what mob type they spawn and override that instead of re-writing every proc they have just to change mob type and *maybe* quantity.

LanceSmites328 (commented):
image

code/modules/mob/living/simple_animal/hostile/ordeal/crimson.dm
Coxswain-Navigator: none

code/modules/mob/living/simple_animal/hostile/ordeal/crimson.dm
Coxswain-Navigator: I did attempt this approach, but I ended up unable to find an implementation that was worth the trouble. The problem is that I needed the proc to call a proc on the mob being spawned, and not all of the mobs have said proc...

LanceSmites328 (approved):
clown

PR #1172, opened 2023-07-24
standardize watch spawn amounts

Retrino:
## About The Pull Request

this *should* in theory be the last of the extra watches. removed extra watches from k corp and f corp.

## Why It's Good For The Game

maps should always have same watch amount for the ro. i checked the other maps, so hopefully thats now true across the board


Kitsunemitsu (approved):
Approved.

PR #1181, opened 2023-07-28
Wave Nests, Wave Based Monster Spawning [DONE]

InsightfulParasite:
## About The Pull Request
A structure that summons monsters. But after summoning a certain amount it will send them to a destination.
image
The structure has high resistant to white and pale damage attacks. There may also be the issue of monsters destroying their own nests due to getting upset near it. Also if intercepted early they will not continue their assault and just stand around where they had combat.

The structure will stop processing if it does not have a assault_target so in order to restart it you need to proc BeginProcessing.
On sewers R corp map the plastic Flaps confuse the hell out of A star pathing.

Also introduces invisible effect called wave_commander. Its only visible to ghosts and its purpose is to drag the assault party to their destination.
image

https://studio.youtube.com/video/13GqLZ0Ymq0/edit

System has been changed into a component.

## Why It's Good For The Game
It **MIGHT** be fun for a wave based survival event. But overall this is mostly me messing with AI and seeing what i can do with it.

## Changelog
:cl:
add: Wave Nests
/:cl:

### DIFF
--- a/code/datums/components/monwave_spawners.dm
+++ b/code/datums/components/monwave_spawners.dm
@@ -1,0 +1,196 @@
+#define SEND_ON_SIGNAL 1
+#define SEND_ON_WAVE 2
+#define SEND_TILL_MAX 3
+
+/datum/component/monwave_spawner
+	var/assault_pace
+	//Cooldowns for procs.
+	var/resume_cooldown = 0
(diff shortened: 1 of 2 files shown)


Kitsunemitsu (approved):
This looks great for things such as COL reworks.
Approved.

PR #1190, opened 2023-08-02
Prevent pygmalion from attacking insane targets

EnderMage99:
Prevent pygmalion from attacking insane targets


Kitsunemitsu (approved):
Code looks like it would work
Dev said it worked

If there's any problems you're gonna have to refix it

PR #1191, opened 2023-08-02
Orange tree no longer spreads from corpses

Coxswain-Navigator:
## About The Pull Request
Makes orange tree a bit easier to deal with.

## Why It's Good For The Game
He's dead, Jim.

## Changelog
:cl:
balance: orange tree disease no longer spreads from dead bodies
/:cl:

### DIFF
--- a/code/datums/diseases/orange_tree.dm
+++ b/code/datums/diseases/orange_tree.dm
@@ -25,6 +25,12 @@
 		SanityCheck(TRUE)
 	return ..()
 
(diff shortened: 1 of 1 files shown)


EgorDinamit:
Diseases have flags for exactly that, use them instead.

Coxswain-Navigator:
And which flag would that be? I don't know of any that are appropriate here.

Coxswain-Navigator:
Turns out no such flag exists, so this code is the correct implementation as far as I'm concerned.

LanceSmites328 (approved):

PR #1192, opened 2023-08-02
Adds shelter from the 27th of march and static tools

Coxswain-Navigator:
## About The Pull Request
Adds wishing well to every map as a static tool, as well as Shelter from the 27'th of march

- Shelter from the 27th of March is simply a small room on the centcomm z-level loaded with some emergency food and water.  Entering takes 30 seconds and leaving stuns you for 15 seconds.

Future considerations:
I'm not a mapper and technically I have no idea what I'm doing.

## Why It's Good For The Game
No more stupid hiding in the manager's office

## Changelog
:cl:
add: Added Shelter from the 27th of March
tweak: Wishing well is now always present on every facility map
/:cl:


Kitsunemitsu (changes requested):
Closed off, windowless rooms like the ones you added are frowned upon as abnos have a hard time getting inside, and it acts as a safe room of sorts.

It's why generally only the RO backroom has one (which was designed to purposely cut line of sight, and is an evil we are willing to use as a compromise.)

Tokanova:
Isn't that the point of shelter? Abnos can't get in it.

code/modules/mob/living/simple_animal/abnormality/_tools/he/shelter.dm
```diff
@@ -0,0 +1,99 @@
...
+	..()
+
+/obj/structure/toolabnormality/shelter/exit
+	name = "exit"
+	desc = "Probably doesn't lead to the safest place on earth."
+	icon_state = "shelter_out"
+
+/obj/structure/toolabnormality/shelter/exit/attack_hand(mob/living/carbon/human/user)
+	if(!do_after(user, 3 SECONDS, user))
+		to_chat(user, "<span class='userdanger'>You decide it's safer in the shelter from the 27th of March.</span>")
```
LanceSmites328, line 44:
```suggestion
		to_chat(user, "<span class='userdanger'>You decide it's safer in the shelter.</span>")
```
Full name feels like a mouth full and this feels more organic.

code/modules/paperwork/records/info/tools.dm
```diff
@@ -151,3 +151,13 @@
...
+
+//Shelter From The 27th Of March
+/obj/item/paper/fluff/info/tool/shelter
+	name = "T-09-82"
+	info = {"<h1><center>T-09-82</center></h1>	<br>
+	Name : Shelter From The 27th Of March<br>
+	Risk Class : HE	<br>
+	- A bomb shelter that completely protects its interior from all hazardous influences of the outside world, including light.	<br>
+	- Those who take refuge inside were perfectly shielded from physical and psychological attacks, even as far as being shielded from radio communications.	<br>
+	- However, this shelter, while perfectly safe on the inside, alters the reality of the outside to be even more hopeless. It literally makes itself into the safest place on Earth."}
```
LanceSmites328, line 163: While copied, it does, in fact, not do that.

LanceSmites328 (changes requested):
Conceptually approved, I just feel like it doesn't need a whole room and could probably be attached to a wall somewhere? Uncertain about that last part.

Kitsunemitsu (commented):
Giving mapping approval.

code/modules/paperwork/records/info/tools.dm
Coxswain-Navigator: It's just flavor!

LanceSmites328 (approved):

PR #1215, opened 2023-08-06
Parasite Tree is now from Lobotomy corp

Kitsunemitsu:
## Changelog
:cl:
fix: Parasite Tree is now from Lobotomy corp
/:cl:


LanceSmites328 (approved):
Welcome home bb

PR #1257, opened 2023-08-19
adds looc

Kitsunemitsu:
## About The Pull Request
Adds LOOC

## Why It's Good For The Game
People have been talking in say for LOOC and we don't actually have LOOC

## Changelog
:cl:
add: Added LOOC
/:cl:


LanceSmites328 (changes requested):
Seems to work but I'd like a settable keybind for it, perhaps P or L by default, as those are close to O which is used for "OOC". Also uncertain if it works with the current OOC toggle, as there is no LOOC toggle preference available to set.

LanceSmites328:
> Seems to work but I'd like a settable keybind for it, perhaps P or L by default, as those are close to O which is used for "OOC". Also uncertain if it works with the current OOC toggle, as there is no LOOC toggle preference available to set.

Have personally added both a keybinding _and_ a chat tab toggle for this.

LanceSmites328 (approved):

PR #1259, opened 2023-08-20
HitInTurf Fix and the Discovery of Output Variables

LanceSmites328:
## About The Pull Request
Very slightly changes HitInTurf to not use Output variables accidentally.
If you set the return value of a proc to an argument taken by the proc it will edit the original list as well as the return value. TLDR: it sets a reference to the variable instead of duplicating it's contents/value. This would let you edit a variable's origin as well as whatever is being output. Is this useful? No clue.

Oh right, also slightly adjust KoG hit volumes and prevents the proc from sending damage notifications if damage is set to 0.

## Why It's Good For The Game
Bug(?) fix


Kitsunemitsu (approved):
Looks good to me, just merge when you're ready.

PR #1261, opened 2023-08-21
Ports radio noises

Coxswain-Navigator:
## About The Pull Request
Ports https://github.com/Quasar-13/Quasar-13/pull/443 on request of our beloved headcoder.

## Why It's Good For The Game
More annoying sounds

## Changelog
:cl:
soundadd: added radio sound effects
/:cl:


Kitsunemitsu (dismissed):
Port I 100% know works.

LanceSmites328 (approved):

EgorDinamit (changes requested):
- Make it a preference option to hear those sounds.
- Reduce the volume and range of the sound itself, so that people don't get earblasted from 5 tiles away.

code/game/objects/items/devices/radio/radio.dm
```diff
@@ -221,6 +222,9 @@
 		return
 	if(!M.IsVocal())
 		return
+	if(radio_sounds.len) //Sephora - Radios make small static sounds now.
+		var/sound/radio_sound = pick(radio_sounds)
```
EgorDinamit, line 226: What you are picking isn't a type of sound datum, just a path.
```suggestion
		var/radio_sound = pick(radio_sounds)
```

code/game/objects/items/devices/radio/radio.dm
```diff
@@ -221,6 +222,9 @@
 		return
 	if(!M.IsVocal())
 		return
+	if(radio_sounds.len) //Sephora - Radios make small static sounds now.
```
EgorDinamit, line 225: This is honestly a nitpick, but:
```suggestion
	if(LAZYLEN(radio_sounds)) //Sephora - Radios make small static sounds now.
```
Never use A.len, instead opt in for length(A) proc.

Coxswain-Navigator:
> * 
> * Reduce the volume and range of the sound itself, so that people don't get earblasted from 5 tiles away.

I did the latter, but I believe the former would require me to write a new copy of the radio proc or the playsound() proc, which seems like unnecessary bloat. How would this respect radios that are on the floor with no player otherwise?

EgorDinamit:
What? You just tie preference to sound being sent to the player with it disabled.

LanceSmites328 (dismissed):

EgorDinamit (approved):
Looks more or less fine; If nothing breaks during test-merge - it is good to go.

PR #1265, opened 2023-08-22
Fixes Apex Predator records

JackFrost2424:
## About The Pull Request

Apparently, it had "name" instead of "abno_code". Assuming that's why it doesn't have a code in record cabinet. This changes that.

## Why It's Good For The Game

Fixes a really annoying but ultimately very minor bug.

## Changelog
:cl:
fix: Fixes a record bug that's been around for a bit
/:cl:


Kitsunemitsu (approved):
Approved.

PR #1269, opened 2023-08-23
Adds backwards clock

Coxswain-Navigator:
## About The Pull Request
Adds a new static tool, backwards clock
Works differently than in Lobotomy Corp, an agent must be at least level 3 to crank it and can only crank it once per ckey.
Activating the abnormality also requires at least level 3, and deals damage to every abnormality or ordeal based on a formula of  level * 500. 
Being higher than level 5 causes the abnormality to deal 10,000 damage. Regardless of the outcome, the user is turned to dust.

An active midnight ordeal prevents it from being activated entirely, and event bosses are put in a blacklist of mobs affected.

## Why It's Good For The Game
No.

## Changelog
:cl:
add: Added backwards clock
/:cl:


Kitsunemitsu (commented):
Why is this a static tool, perchance?

Coxswain-Navigator:
> Why is this a static tool, perchance?

The gameplay impact is pretty high, and I can definitely see people trying to restart the round just for a chance to spawn it. I think that like button, it should always be around.

code/modules/paperwork/records/info/tools.dm
```diff
@@ -161,3 +161,16 @@
...
+
+//Backwards Clock
+/obj/item/paper/fluff/info/tool/clock
+	name = "D-09-104"
+	info = {"<h1><center>D-09-104</center></h1>	<br>
+	Name : Backwards Clock<br>
+	Risk Class : WAW	<br>
+	- Every time a different employee at or above level 3 winds up the machine, a light turns on.	<br>
+	- When all four lights are turned on, the abnormality's ability is ready.	<br>
+	- When an employee at or above uses the machine when the abnormality's ability is ready, the abnormality's ability is triggered.	<br>
```
LanceSmites328, line 173: at or above what

LanceSmites328 (dismissed):

Kitsunemitsu (approved):
Alright, approving

PR #1284, opened 2023-08-26
Fish Value based on Fish Rarity

InsightfulParasite:
## About The Pull Request
Fish value is now based on fish rarity and bone hooks can be collected into fish bags for quick fish points.

## Why It's Good For The Game
The player syndicates are realizing I've coded fishing points to be a true scam.

## Changelog
:cl:
tweak: fish_market, fish_bag
/:cl:


TemperanceTempy:
Everything else looks good, though I'd prefer base fish to have a higher value (maybe like 2 or 5?). The 1000 ahn being put to 1000 pts is redundant though, as fish already are sold at the selling machines for 10 per fish. I'd suggest either changing it back or if there's a worry about a rare fish getting caught and breaking the economy removing that option entirely.

InsightfulParasite:
> Everything else looks good, though I'd prefer base fish to have a higher value (maybe like 2 or 5?). The 1000 ahn being put to 1000 pts is redundant though, as fish already are sold at the selling machines for 10 per fish. I'd suggest either changing it back or if there's a worry about a rare fish getting caught and breaking the economy removing that option entirely.

I dunno. I feel like the minimum should be the minimum though.

TemperanceTempy:
> > Everything else looks good, though I'd prefer base fish to have a higher value (maybe like 2 or 5?). The 1000 ahn being put to 1000 pts is redundant though, as fish already are sold at the selling machines for 10 per fish. I'd suggest either changing it back or if there's a worry about a rare fish getting caught and breaking the economy removing that option entirely.
> 
> I was unaware the selling machine accepts fish. I may just replace the 1000 ahn with a lantern and water mister. as for the base fish having a higher value i just feel weird about it. Because then what's the point of a minimum. Like what would become 1 pebble... actually i guess 1 pebble would equal 1 pebble your right ill change it to 2. Also when fishing you tend to fish up about 10 basic fish per 1 rare fish due to rarity so i dont want rares to equal 2x the value of a basic fish since basic fish are going to be the bulk of your haul.

Thats fair, 2 per common fish should be fine then, and yea i think a lantern or something else would work

LanceSmites328 (approved):
Approving and leaving unmerged incase remaining a WIP, Ping me if otherwise

PR #1293, opened 2023-08-29
The great attribute update-ening

Coxswain-Navigator:
image

## About The Pull Request
Expands the show attributes panel to display a more accurate reflection of the player's attributes, and introduces the concept of attribute bonuses.
Attribute bonuses are different from normal attributes in the following ways :

- Only affect the player's attributes specifically.
- Does NOT affect attribute requirements when working on abnormalities.
- Does NOT affect whether or not the player meets requirements to use E.G.O.
- Can go in the negatives. (max sanity and health cannot fall below their minimums of 100, whether or not I change this TBD.)
- Typically obtained through tools and acquiring buffs and de-buffs from items or abnormalities

The following abnormalities will be affected by this pull request : 

- Dr Jekyll's Tincture
- Firebird
- You must be happy (When implemented)
- Pygmalion
- Singing Machine
- The Jester of Nihil
- So that no one will cry
- Any abnormality I arbitrarily determine needs this in the near future, typically those with temporary de-buffs.

## Why It's Good For The Game
Gives developers more freedom in how they create things that modify attributes, no more having to worry that players won't be able to wear E.G.O. when weakening them, and no more having players use a short-term buff to wear high-level E.G.O. and then never take it off.

## Changelog
:cl:
tweak: Certain buffs and debuffs no longer affect E.G.O. and work requirements
balance: Getting certain debuffs will no longer cause you to be unable to use your E.G.O.
code: added attribute bonus variables and procs to attributes
/:cl:


code/datums/attributes/_attribute.dm
```diff
@@ -73,6 +89,15 @@ GLOBAL_LIST_INIT(attribute_types, list(
 		return 1
 	return max(1, atr.get_level())
 
+//Getting modified level, which is level + buff + bonus
+/proc/get_modified_attribute_level(mob/living/carbon/human/user, attribute)
+	if(!istype(user) || !attribute)
+		return 1
+	var/datum/attribute/atr = user.attributes[attribute]
+	if(!istype(atr))
+		return 1
```
LanceSmites328, line 98: Why not FALSE / 0? This would allow for checking if an attribute/person even exists

code/datums/attributes/_attribute.dm
```diff
@@ -91,8 +116,15 @@ GLOBAL_LIST_INIT(attribute_types, list(
...
 	return max(1, atr.get_level_buff())
 
-// Attribute buffs
+/proc/get_level_bonus(mob/living/carbon/human/user, attribute)
+	if(!istype(user) || !attribute)
+		return 1
+	var/datum/attribute/atr = user.attributes[attribute]
+	if(!istype(atr))
+		return 1
+	return max(1, atr.get_level_bonus())
```
LanceSmites328, line 125: Same as above

LanceSmites328 (changes requested):
I won't lie, I was very worried with how this would be implemented but everything it affects I can really only see a positive change. Well done IMO

code/datums/attributes/_attribute.dm
Coxswain-Navigator: The original attribute-related procs follow this same methodology, for a good reason I presume. I'm not the person who originally wrote them, I could look at doing it for all of them if it doesn't break things.

LanceSmites328 (approved):

LanceSmites328:
Approving and waiting for testing

code/datums/attributes/_attribute.dm
EgorDinamit: It is to avoid divisions by 0 in certain places.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -80,7 +80,25 @@
...
+				bonusname = "Max Health"
+				bonus_amount = 100
+			if("Prudence")
+				bonusname = "Max Sanity"
+				bonus_amount = 100
+			if("Temperance")
+				bonusname = "Work Speed"
+			if("Justice")
+				bonusname = "Movement Speed"
+		dat += "[bonusname] : [round(atr.get_level()) + bonus_amount] + [round(atr.level_bonus)]"
```
EgorDinamit, line 99: This feels like a very hacky way to do it...

Coxswain-Navigator: It IS, but it is fine for now. Someday down the line I might separate work speed and movement speed into work speed/work chance and movement speed/force but this works for the current system.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -80,7 +80,13 @@
 		var/datum/attribute/atr = attributes[atrname]
 		dat += "[atr.name] [get_attribute_text_level(atr.get_level())]: [round(atr.level)]/[round(atr.level_limit)] + [round(atr.level_buff)]"
 
-	var/datum/browser/popup = new(viewer, "skills", "<div align='center'>Attributes</div>", 300, 300)
+	dat += ""
+	for(var/atrname in attributes) //raw stats (health, sanity etc)
+		var/datum/attribute/atr = attributes[atrname]
+		for(var/i=atr.affected_stats.len, i>=1, i--)
```
EgorDinamit, line 86: Why this and not a more simple loop?

LanceSmites328 (approved):

PR #1294, opened 2023-08-29
Excludes Disposal Pipes from HurtInTurf

LanceSmites328:
## About The Pull Request
Gives a list for type-based exclusion from HurtInTurf

## Why It's Good For The Game
They were breaking disposals through the floor, which while funny shouldn't happen.


Kitsunemitsu (approved):
Very fucking funny

PR #1296, opened 2023-08-30
Misc City Touch-Ups

LanceSmites328:
## About The Pull Request
Touches Up the Ordeal Extractor and Item Selling machine to be overall more user-friendly.

Ordeal Extractor now lets you select what to refine instead of just going through everything.
Ordeal Extractor and Ingots now deposit their resources (if more than one) in a nice box built to hold metals and ordeal parts, and can even pick them up on click. Very useful.
Ordeal Extractor now accepts Tressium so you can now refine the bulk you get from the Item Selling machine into more potent ingots, however costly that may be.

Item Selling Machine now accepts input from storage containers, like bags and boxes.
Item Selling Machine now allows for the display of lists of items and their values, with automatic compensation for more items being added to these lists.

Dice Template had an issue with it's force fixed and should be working properly.

## Why It's Good For The Game
General fixes and UX improvements.

### DIFF
--- a/ModularTegustation/tegu_items/associations/machines.dm
+++ b/ModularTegustation/tegu_items/associations/machines.dm
@@ -1,5 +1,5 @@
 /obj/structure/itemselling
-	name = "Item Selling machine"
+	name = "item selling machine"
 	desc = "A machine used to sell items to the greater city"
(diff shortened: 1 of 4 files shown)


Kitsunemitsu (approved):
Looks good to me.

PR #1298, opened 2023-08-30
Shrimps Drop From Pods

LanceSmites328:
## About The Pull Request
Makes Shrimp from Exec drop from Drop Pods instead of manifesting. Prevents Shrimp from surprise shotgunning people minding their own business.

## Why It's Good For The Game
Improvement


Kitsunemitsu (approved):
TRUE!

PR #1311, opened 2023-08-31
Slowdown Bullet Rework

InsightfulParasite:
## About The Pull Request
Reworks slowdown bullets and mob speeds.
Replaces abnormality auto speed calculation with a proc.
Discovered move_to_delay was the lag in deciseconds while speed was a multiplication of the users speed.
Slowdown Bullets now work on all mobs but managers can only fire slowdown on mobs that dont match their faction.
Slowdown bullets now work on carbons as well as abnormalities without relying on stamina damage.

> Big bird has 3.5 Cache_move_delay but it should have 1.5 move_delay. 
> Nosferatu and Titania is 3 but it should be 1
> Golden Apple, Blue Shep, and All Around Helper is 2.5 but should be 0.5
> Picene Mermaid is 1.5 but should be -0.5
> 
> is the basis of speed 2 and the speed is just added onto that?
> 2 + 1.5 = 3.5 
> 2 + 1 = 3
> 2 - 1 = 2.5 

## Why It's Good For The Game
People have been reporting a bug where abnormalities will speed up or slow down after being slowed.

## Changelog
:cl:
tweak: Slowdown Bullets
tweak: Auto Abnormality Speed Calculation
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -981,21 +981,23 @@
 
 
 //~~~LC13 General Debuffs~~~
-
+#define CARBON_HALFSPEED /datum/movespeed_modifier/qliphothoverload
 /datum/status_effect/qliphothoverload
 	id = "qliphoth intervention field"
 	duration = 10 SECONDS
 	alert_type = null
 	status_type = STATUS_EFFECT_REFRESH
 	var/statuseffectvisual
-	var/originalstamina = 0
 
 /datum/status_effect/qliphothoverload/on_apply()
 	. = ..()
-	var/mob/living/simple_animal/hostile/L = owner
-	L.adjustStaminaLoss(160, TRUE, TRUE)
-	L.stamina_recovery *= 0.01 //anything with below 10 stamina recovery will continue to lose stamina
-	L.update_stamina()
+	if(ishostile(owner))
+		var/mob/living/simple_animal/hostile/L = owner
+		L.TemporarySpeedChange(4, duration)
+	if(iscarbon(owner))
+		var/mob/living/carbon/M = owner
+		M.add_movespeed_modifier(CARBON_HALFSPEED)
+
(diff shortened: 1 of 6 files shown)


code/datums/status_effects/debuffs.dm
```diff
@@ -981,38 +981,44 @@
...
-
+#define CARBON_HALFSPEED /datum/movespeed_modifier/qliphothoverload
 /datum/status_effect/qliphothoverload
 	id = "qliphoth intervention field"
 	duration = 10 SECONDS
 	alert_type = null
 	status_type = STATUS_EFFECT_REFRESH
 	var/statuseffectvisual
-	var/originalstamina = 0
+	var/speed_memory = 0
```
LanceSmites328, line 991: You could obliterate this entire variable by using TemporarySpeedChange(4, 10 SECONDS) from hostile, as it more or less does the exact same thing. It also lowers the quantity of code in `on_remove` since it's done automatically.

code/datums/status_effects/debuffs.dm
```diff
@@ -981,38 +981,44 @@
...
 	. = ..()
-	var/mob/living/simple_animal/hostile/L = owner
-	L.adjustStaminaLoss(160, TRUE, TRUE)
-	L.stamina_recovery *= 0.01 //anything with below 10 stamina recovery will continue to lose stamina
-	L.update_stamina()
+	if(ishostile(owner))
+		var/mob/living/simple_animal/hostile/L = owner
+		speed_memory = 4
+		L.SpeedChange(speed_memory)
+		L.UpdateSpeed()
```
LanceSmites328, line 999: Redundant, SpeedChange() calls UpdateSpeed()
```suggestion
```

LanceSmites328 (changes requested):

code/datums/status_effects/debuffs.dm
InsightfulParasite: Ill swap it with the temp speed change.

LanceSmites328 (approved):

PR #1316, opened 2023-09-01
Unaimed shots miss cloaked Apex Predator

Anonmare:
## About The Pull Request
Shots not aimed at Apex Predator will now pass over it whilst it is cloaked (except magic bullet which hits non-dense entities). Decloaked, it will be hit by stray and unaimed shots as before.

This will also let cloaked Apex occupy the same tile as other abnormalities but I don't anticipate this causing balance concerns.

## Why It's Good For The Game
It probably isn't 

## Changelog
:cl:
balance: Unaimed shots now miss Apex Predator
/:cl:


Kitsunemitsu (approved):
Looks good to me.

PR #1318, opened 2023-09-01
Worse Dragon

LanceSmites328:
## About The Pull Request
Does 2 things:
1. Makes the Yin-Yang Dragon bigger, as it was apparently not big enough. It's now 50% bigger and hits a 5x5 around itself instead of a 3x3.
2. Makes the Yin-Yang Dragon immune to deep water.

## Why It's Good For The Game
1. Makes the threat more threatening.
2. Prevents water-based fuckery.


Kitsunemitsu (approved):
Simple change

PR #1325, opened 2023-09-03
Attempting to make AI Leadership Component [Done but unapproved]

InsightfulParasite:
## About The Pull Request
The system works on a component that is attached to a mob and perhaps a item. The component processes and scans its surroundings for living mobs. When a mob that can be recruited is found they register it with a collective management datum that catalogs all of the followers and leaders.

- [x]  The component works and can be attached to a mob
- [x]  The component doesnt lag on a personal server even when attached to 3+ leader mobs
- [x]  Followers are not stuck in the list when their leader perishes.
- [x]  The component can be attached to carbons.
- [x]  It works for at least 1 minute.
- [x]  A mob can be recruited by one leader, disbanded, and then recruited by another very easily.
- [x]  The team composition feature works.

Honestly if there is a more acceptable way to do this im all ears. Converting these into a component and datum has been tiresome. Maybe i should look into abilities and if i can change these into those and the people who want to have leader mobs can just do that instead of adding a component in one line. Putting in a new variable for hostiles like the spawner datum var/nest that serves no purpose other than remembering the components parent nest.

## Why It's Good For The Game
Makes leadership AI less hardcoded into mobs.

## Changelog
:cl:
add: ai_leadership
tweak: indigo dusk
tweak: steel dusk
/:cl:

### DIFF
--- a/code/datums/components/ai_leadership.dm
+++ b/code/datums/components/ai_leadership.dm
@@ -1,0 +1,301 @@
+/* Component for AI leadership. Ai will leader
+	other mobs of the listed type. */
+
+/datum/component/ai_leadership
+	dupe_mode = COMPONENT_DUPE_UNIQUE
+	//Cooldowns for recruitment
+	var/recruit_cooldown = 0
+	var/recruit_delay = 1 SECONDS
+	//Cooldowns for headcount
+	var/headcount_cooldown = 0
+	var/headcount_delay = 15 SECONDS
+	//Amount of allowed followers
+	var/unit_amount
+	//If the team should only have a certain amount of type.
+	var/unique_team
+	//If followers will return to a base when disbanded
+	var/return_to_fob
+	//Datum that tracks all of the components
+	var/static/datum/collective_management/follower_tracker
+	//Allowed types of followers
+	var/list/possible_followers = list()
+	//This components current followers.
+	var/list/followers = list()
+
+/* This could use some improvement.
+	Or a full on replacement. */
+/datum/component/ai_leadership/Initialize(allowed_types, amount = 6, unique = FALSE, forwardbase = FALSE)
+	if(!isatom(parent))
+		return COMPONENT_INCOMPATIBLE
+
+	if(!follower_tracker)
+		follower_tracker = new /datum/collective_management()
+	var/atom/L = parent
+
+	return_to_fob = forwardbase
+	unique_team = unique
+	unit_amount = amount
+	possible_followers += allowed_types
+
+	follower_tracker.AddLeader(L)
+	ScanLocation()
+	START_PROCESSING(SSdcs, src)
+
+/datum/component/ai_leadership/RegisterWithParent()
+	if(isliving(parent))
+		RegisterSignal(parent, COMSIG_LIVING_DEATH, PROC_REF(RemoveLeader))
+		if(ishostile(parent))
+			RegisterSignal(parent, COMSIG_PATROL_START, PROC_REF(HeadCount))
+
+/datum/component/ai_leadership/UnregisterFromParent()
+	if(isliving(parent))
+		UnregisterSignal(parent, COMSIG_LIVING_DEATH)
+		if(ishostile(parent))
+			UnregisterSignal(parent, COMSIG_PATROL_START)
+
+//On death remove this component and disband their followers.
+/datum/component/ai_leadership/proc/RemoveLeader()
+	SIGNAL_HANDLER
+
+	follower_tracker.RemoveLeader(parent, followers)
+	STOP_PROCESSING(SSdcs, src)
+	qdel(src)
+
+/datum/component/ai_leadership/process()
+	//If not a hostile then it would not get a signal.
+	if(!ishostile(parent) && world.time > headcount_cooldown && followers.len)
+		HeadCount(parent)
+	if(world.time > recruit_cooldown && followers.len < unit_amount)
+		ScanLocation()
+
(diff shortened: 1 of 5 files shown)


code/datums/components/ai_leadership.dm
```diff
@@ -0,0 +1,291 @@
...
+//Is the recruit already recruited in the system?
+/datum/component/ai_leadership/proc/FollowerLedger(mob/living/recruit)
+	if(follower_tracker.FindRecruit(recruit))
+		return TRUE
+	return FALSE
+
+//Check the type system
+/datum/component/ai_leadership/proc/CheckUnitType(list/type_list, mob/living/unit_type)
+	for(var/i in type_list)
+		if(unit_type.type == i)
```
LanceSmites328, line 198:
```suggestion
		if(istype(unit_type, i))
```
This will allow for subtypes, to be fair. If you don't want that ignore this.

code/datums/components/ai_leadership.dm
```diff
@@ -0,0 +1,291 @@
...
+		if(M.type == i)
+			our_capacity = possible_followers[i]
+			break
+	if(our_capacity <= many_we_have)
+		return FALSE
+	return TRUE
+
+/*Used to be count by type until i realized count by
+	type includes subtypes when i need a strict type.*/
+/datum/component/ai_leadership/proc/CountByTroopType(mob/living/M)
```
LanceSmites328, line 216: This would count by a mob's type, as opposed to a simple typepath, meaning it'd need a reference mob to compare from.

code/datums/components/ai_leadership.dm
```diff
@@ -0,0 +1,291 @@
...
+		var/mob/living/simple_animal/hostile/H = parent
+		//Are they the same faction as us?
+		if(!H.faction_check_mob(L))
+			return FALSE
+		//If they currently have a target let them finish up.
+		if(H.target)
+			return FALSE
+	var/recruit_type = CheckUnitType(possible_followers, L)
+	//Are they a type in our possible follower types?
+	if(!recruit_type)
```
LanceSmites328, line 167:
```suggestion
	//Are they a type in our possible follower types?
	if(!CheckUnitType(possible_followers, L))
```

code/datums/components/ai_leadership.dm
```diff
@@ -0,0 +1,291 @@
...
+	if(follower_tracker.FindRecruit(recruit))
+		return TRUE
+	return FALSE
+
+//Check the type system
+/datum/component/ai_leadership/proc/CheckUnitType(list/type_list, mob/living/unit_type)
+	for(var/i in type_list)
+		if(unit_type.type == i)
+			return i
+	return null
```
LanceSmites328, line 200:
```suggestion
	return FALSE
```
A lot better than Null

LanceSmites328 (changes requested):

code/datums/components/ai_leadership.dm
```diff
@@ -0,0 +1,290 @@
...
+	STOP_PROCESSING(SSdcs, src)
+	qdel(src)
+
+/datum/component/ai_leadership/process()
+	//If not a hostile then it would not get a signal.
+	if(!ishostile(parent) && world.time > headcount_cooldown && followers.len)
+		HeadCount(parent)
+	if(world.time > recruit_cooldown && followers.len < unit_amount)
+		ScanLocation()
+		recruit_cooldown = world.time + recruit_delay
```
EgorDinamit, line 70: Any reason to put it here instead of the proc's body?

code/datums/components/ai_leadership.dm
```diff
@@ -0,0 +1,290 @@
...
+		if(!H.faction_check_mob(L))
+			return FALSE
+		//If they currently have a target let them finish up.
+		if(H.target)
+			return FALSE
+	//Are they a type in our possible follower types?
+	if(!CheckUnitType(possible_followers, L))
+		return FALSE
+	//If we only have the command capacity to command more than one of them.
+	if(unqiue_team)
```
EgorDinamit, line 169: *Grammar check*
```suggestion
	if(unique_team)
```

code/datums/components/ai_leadership.dm
```diff
@@ -0,0 +1,290 @@
...
+	follower_tracker.RecruitFollower(L)
+	FollowLeader(L)
+
+//Releases the recruit from our command
+/datum/component/ai_leadership/proc/Disband(mob/living/L)
+	followers -= L
+	follower_tracker.DismissFollower(L)
+
+//Is the recruit already recruited in the system?
+/datum/component/ai_leadership/proc/FollowerLedger(mob/living/recruit)
```
EgorDinamit, line 189: "Ledger"..?

code/datums/components/ai_leadership.dm
InsightfulParasite: What?

LanceSmites328: You need an actual instance of a mob as opposed to just the path in this proc.

InsightfulParasite: Okay so i did this weird. The count by troop type apparently is counting how many of the unit being scanned are in the team and if we should just ignore that unit.

InsightfulParasite: I need to fix this.

Kitsunemitsu (approved):
Some Code de-shittification. Appreciate it; and approved.
Egor and Redacted both had their issues resolved.

PR #1326, opened 2023-09-03
Damage Coeff Change Standardization

LanceSmites328:
## About The Pull Request
Creates a system of datums that can be used to properly manage changes to a `simple_animal`'s damage_coeff without causing issues, such as a debuff wearing off after the animal re-defines it's resistances, causing it to buff those resistances further.

So, this now also rewrites existing `damage_coeff` code to use a new datum that is extremely similar to armor for the purposes of performance, as the old way of optimization using lists was simply not compatible.

These modifiers are meant to typically be provided by status effects that later remove them. Some cases, such as Doomsday Calendar are exceptions, due to their theoretically infinitely scaling self-debuff.

This should probably be stress tested a bit.

## Why It's Good For The Game
Solves issues of incredible resistances by stuff like Arbiter when faced with certain realized EGO.
Allows armor debuffs to be used more realistically without worrying about certain issues.


Kitsunemitsu (dismissed):
Code cleanup

EgorDinamit (approved):
Let's get it merged and see what happens.

PR #1331, opened 2023-09-04
Add COMSIG_GLOB_PATROL_START

InsightfulParasite:
## About The Pull Request
Adds patrol signal since im making a component that has a check that only happens when the hostile is beginning a patrol. May also have other uses.

Im worried about the lag a bunch of mobs all patrolling at once would cause if they send signals but im concerned about the lag caused by a bunch of mobs all patrolling anyways.

## Why It's Good For The Game
I'm making a component that could use it but if its deemed unnecessary ill use some other method.

## Changelog
:cl:
add: Signals COMSIG_GLOB_PATROL_START and COMSIG_PATROL_START
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals.dm
+++ b/code/__DEFINES/dcs/signals.dm
@@ -560,6 +560,9 @@
 // /mob/living/simple_animal/hostile signals
 #define COMSIG_HOSTILE_ATTACKINGTARGET "hostile_attackingtarget"
 	#define COMPONENT_HOSTILE_NO_ATTACK (1<<0)
+/// a hostile has started their patrol (datum/source, mob/living/simple_animal/hostile/mover, turf/target_location)
+#define COMSIG_GLOB_PATROL_START "!patrol_start"
+#define COMSIG_PATROL_START "patrol_start"
 
 // /obj signals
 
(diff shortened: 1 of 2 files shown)


code/__DEFINES/dcs/signals.dm
```diff
@@ -560,6 +560,9 @@
 // /mob/living/simple_animal/hostile signals
 #define COMSIG_HOSTILE_ATTACKINGTARGET "hostile_attackingtarget"
 	#define COMPONENT_HOSTILE_NO_ATTACK (1<<0)
+/// a hostile has started their patrol
+#define COMSIG_GLOB_PATROL_START "!patrol_start"
+#define COMSIG_PATROL_START "patrol_start"
```
LanceSmites328, line 565:
```suggestion
/// a hostile has started their patrol
/// (datum/source, mob/living/simple_animal/hostile/mover, turf/target_location)
#define COMSIG_GLOB_PATROL_START "!patrol_start"
#define COMSIG_PATROL_START "patrol_start"
```

LanceSmites328 (changes requested):

LanceSmites328 (approved):

PR #1332, opened 2023-09-04
Re-Wound

Kitsunemitsu:
They literally fuckign asked for it.


code/datums/wounds/_wounds.dm
```diff
@@ -110,8 +110,12 @@
...
  */
 /datum/wound/proc/apply_wound(obj/item/bodypart/L, silent = FALSE, datum/wound/old_wound = null, smited = FALSE)
-	return
-	/*
+	if(SSmaptype.maptype != "city")
+		return
+
+	if(prob(20))	//.Make this more rare
+		return
+
```
LanceSmites328, line 118: These should probably qdel(src).
```suggestion
	if(SSmaptype.maptype != "city")
		qdel(src)
		return

	if(prob(20))	//.Make this more rare
		qdel(src)
		return

```

LanceSmites328 (changes requested):

Kitsunemitsu:
This was mostly a test PR.

PR #1333, opened 2023-09-04
melee_reach Variable and controlled mob attack rate fix

LanceSmites328:
## About The Pull Request
Adds the "melee_reach" variable that lets mobs perform melee attacks at a range. Used in Porccubus, to be NoriNori, and potentially one of the LOVE bois.
Switches Porccubus to used `ranged_cooldown` and `ranged_cooldown_time` instead of the dash variants because it's simply better.
Makes client-occupied mobs have a attack-cooldown equal to the normal version and gives them an equal attack rate when they have rapid melee.

## Why It's Good For The Game
1. We're adding things that do this and it expands on it for further things that want to.
2. Code Improvement
3. Balancing admin bussing and playables.


Kitsunemitsu (approved):
Code looks good

PR #1339, opened 2023-09-04
Revert "Liver Parasites get blended in Cybernetic Liver"

Kitsunemitsu:
Reverts vlggms/lobotomy-corp13#1314


Kitsunemitsu:
"Emergency"

Issue #1352, opened 2023-09-05
Insane dead bodies can't be aggressive grabbed


Fireiy:
You can't aggressive grab people who died while insane. This is basically only a problem for corpse disposal, but still a bug.

PR #1353, opened 2023-09-05
Massive Sanity Edits

LanceSmites328:
## About The Pull Request
1. IS_DEAD_OR_INCAP no longer things you're knocked the fuck out just because you can't use your HANDS.
2. The Resist AI behavior no longer has ANY minimum range because there's cases where the person COULD be on top of them and this may give a 0 as a result.
3. Insane people have their resist chance reduced to 60 from 90% (it increases over time anyway)
4. UnpossessPawn now cleans up insanity signals.
5. ResistCheck proc is added and is overridden for Fortitude panics.
6. Murder panic speed altered. Is now SLIGHTLY slower than normal but still takes corners better, so...
7. Murder panic now has an interest mechanic which makes them actually lose aggro if they can't see their target for long enough, as opposed to infinitely pathing toward them through walls.
8. Murder panic now defaults to the only available target if there's only one. It also is more likely to pick someone the longer it has to pick.
9. Murder panic tries to find a weapon BEFORE assault someone.
10. Murder panics now attack with appropriate CD, because they attack WHENEVER they can with this CD, meaning they're still likely to get more hits than players.
11. Fixed a BUNCH of null runtimes.
12. Removes insane grab immunity (they actually resist out so it's no longer an issue)
13. Allows insane people (that aren't the target) to walk through hostile mobs
14. Allows insane people to walk through insane people
15. Allows patrolling hostiles to walk through other hostiles.

## Why It's Good For The Game
1. Fixes AI not resisting grabs.
2. Covers edgecase.
3. In testing they were breaking out near-instantly on passive grab.
4. Prevents potential errors
5. Allows for altered break out conditions
6. They were almost running down someone with 100 stats when they had **0**.
7. No more staring at walls for extended periods of time.
8. No more ignoring people. No more targeting no one in an entire room.
9. Armed people are better than unarmed.
10. There's no reason for them to be faster attacking when they already swing at every opportunity they get, compared to players who can miss.
11. Code improvement
12. No longer needed
13. Fixes door stuck
14. Fixes door stuck
15. Fixes door stuck

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -52,3 +52,17 @@
 	if(dna.species.space_move(src))
 		return TRUE
 	return ..()
+
+/mob/living/carbon/human/CanAllowThrough(atom/movable/mover, turf/target)
+	. = ..()
+	if(.) // If we're already letting them through, then might as well not check anything else.
+		return
+	if(!sanity_lost) // If we're not crazy, why the hell would we do these things?
+		return
+	if(ishostile(mover)) // Is it a threat that we now don't care about?
+		return TRUE
+	if(ishuman(mover)) // Is it another crazy person?
+		var/mob/living/carbon/human/H = mover
+		if(H.sanity_lost)
+			return TRUE
+	return
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -497,6 +497,23 @@
 	else
 		return ..()
 
+/mob/living/simple_animal/hostile/CanAllowThrough(atom/movable/mover, turf/target)
+	. = ..()
+	if(.)
+		return
+	if(target == mover) // "I'm KILLING YOU, I'm KILLING YOU" - Jerma985
+		return FALSE
+	if(ishostile(mover))
+		var/mob/living/simple_animal/hostile/H = mover
+		if(LAZYLEN(H.patrol_path)) // Don't block patrolling guys
+			return TRUE
+		return
+	if(ishuman(mover))
+		var/mob/living/carbon/human/H = mover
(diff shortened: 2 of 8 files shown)


TravisAngeI:
pretty good PR, 9. makes good sense, it's pretty common to see someone go murder insane and start trying to punch the closest person when they have an ALEPH weapon on their waist

Coxswain-Navigator:
Fixes https://github.com/vlggms/lobotomy-corp13/issues/1352

Kitsunemitsu:
Going to test this.

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -52,3 +52,17 @@
 	if(dna.species.space_move(src))
 		return TRUE
 	return ..()
+
+/mob/living/carbon/human/CanAllowThrough(atom/movable/mover, turf/target)
+	. = ..()
+	if(.) // If we're already letting them through, then might as well not check anything else.
+		return
```
EgorDinamit, line 59: I am pretty sure an empty `return` will return a null; Test this please.

LanceSmites328: Won't return null due to `. = ..()` which sets the return value `.` to the parent value `..()`

LanceSmites328: Empty return returns `.` by default

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -497,6 +497,25 @@
 	else
 		return ..()
 
+/mob/living/simple_animal/hostile/CanAllowThrough(atom/movable/mover, turf/target)
+	. = ..()
+	if(.)
+		return
```
EgorDinamit, line 503: Same here.

EgorDinamit (approved):
Good to go, very good PR.

PR #1354, opened 2023-09-06
Liver Parasites get blended in Cybernetic Liver 2.0

InsightfulParasite:
## About The Pull Request
Liver Parasites will now check if you have a cybernetic liver and instead of persisting deal a one time 20 liver damage and be cured. The logic behind the liver being damaged is that the blender function is not perfect. Coder logic is that people should not be drinking vile fluid even if they got cogs instead of guts.

update: turns out removing the liver does not remove parasites AT ALL. 
I have to fix this now.

## Why It's Good For The Game
Its only fair that a cybernetic liver has a blending function.

## Changelog
:cl:
fix: Parasitic Infection
tweak: Parasitic Infection
/:cl:


LanceSmites328 (approved):
"Yes it works fine"

PR #1355, opened 2023-09-06
Fixed Quirk Blacklist

PotatoTomahto:
## About The Pull Request
Quirk blacklist actually works now.

## Why It's Good For The Game
So turns out porting all of TG quirk improvements requires rewriting the entire backend of preferences, so this is good enough for now.

## Changelog
:cl:
fix: quirk blacklist
/:cl:


LanceSmites328:
But what If I like taking over 2x damage?

Kitsunemitsu (approved):
Makes things as intended, make a new PR regarding cleaning out the blacklists.

PR #1356, opened 2023-09-06
Disables Broken Light Sparks on City

LanceSmites328:
## About The Pull Request
Yeah

## Why It's Good For The Game
Sparks are lagging out and causing further lag by not deleting. This patches that by disabling them from sparking on city, though if called they'll still make the SFX.


Kitsunemitsu (approved):
Looks great to me.

PR #1363, opened 2023-09-07
Noodle bar spacing fix

LancashireladOnGitHub:
Fixes a issue with some space tiles being included in the "Noodle bar" causing atmospheric fuckery.


Kitsunemitsu (approved):
Map fix

PR #1373, opened 2023-09-09
Fixes a runtime and gives apple reach

LanceSmites328:
## About The Pull Request
Slight runtime could occur if a target had rapid-melee and lost their target during the CheckAndAttack callbacks. This resolves this by adding a target check and early return *prior* to other checks.
Also we give Apple reach because it performed the same override as Porccubus.

## Why It's Good For The Game
Fixes an error and reduces abno code.


Kitsunemitsu (approved):
Approved.

PR #1382, opened 2023-09-10
Sweeper get diminishing returns from eating other sweepers

InsightfulParasite:
## About The Pull Request
Sweepers using the devour ability on other sweepers will result in a 20 hp heal instead of a maximum heal due diminishing returns. Also it sorta makes sense since it would be pre-digested fuel.

## Why It's Good For The Game
Not entirely nessesary for gameplay but there has been a complaint about sentient sweepers doing this.

## Changelog
:cl:
tweak: indigo ordeal devour
/:cl:


LanceSmites328 (commented):
Do sweepers even target other sweepers? The dead bodies still have the same faction so they shouldn't...

TaculoTaculo:
> Do sweepers even target other sweepers? The dead bodies still have the same faction so they shouldn't...

Playable sweepers do.

Kitsunemitsu (approved):
Approved.

PR #1383, opened 2023-09-10
Buffs Swan

TaculoTaculo:
## About The Pull Request
Changes Black Swan to need 2 deaths, 2 insanities and 2 breaches from 5 deaths, 5 insanities and 3 breaches.

## Why It's Good For The Game
Swan was 1:1 to LC, which resulted to it only breaching from clowns/qliph reduction/meltdowns, since the requirements for the breach, which were already high in the base game, were even harder to meet here, this should make it able to actually terrorize some folk sometimes, but the conditions for the breach are still convoluted to achieve,

## Changelog
:cl:
tweak: tweaks records to display the new information
balance: makes swan breach more easily
/:cl:


Kitsunemitsu (approved):
Looks good to me.

PR #1384, opened 2023-09-11
More fixes and tweaks to Indigo Midnight Matriarch AI

InsightfulParasite:
## About The Pull Request
Indigo Midnight has some akwardness in its AI priority. Hopefully this fixes its AI a bit more.
Her hate values should prioritize adjacent enemies more than distant ones and being wounded doesnt override the priority of close enemies.

> 	Priority from greatest to least:
> 	dead close: 90
> 	close: 80
> 	dead far: 40
> 	far: 30
> Being wounded adds around 0.1 to 9.9 based on remaining health to the priority so far away enemies that are carbons will have their health assessed and she will chase after the weakest one.

image
Screenshot of testing ☝️ 
I am aware there is a big gap between dead far and close. Im not sure how and if i should fix that because it leaves room for other priorities.

image

## Why It's Good For The Game
Makes Matriarch not mindlessly seek wounded,

## Changelog
:cl:
fix: matriarch AI
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/indigo.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/indigo.dm
@@ -341,26 +341,61 @@
 	//Default to normal targeting if already phase 3 or only have 1 target
 	if(phase >= 3 || Targets.len <= 1)
 		return ..()
+
+	/* If we have a target and their value is
+		above 80 or "adjacent living creature" then
+		just keep killing them. */
+	if(target)
+		// Focus on finishing them off.
+		if(ValueTarget(target) > 86)
+			return target
+
+	/* Form a list of our targets, value how much we hate
+		them, and then pick the target who has the MOST hate. */
 	for(var/i in Targets)
 		Targets[i] = ValueTarget(i)
 	return ReturnHighestValue(Targets)
 
 //Remind me to return to this and make complex targeting a option for all creatures. I may make it a TRUE FALSE var.
 /mob/living/simple_animal/hostile/ordeal/indigo_midnight/proc/ValueTarget(atom/target_thing, hate_value = 0)
+	/* This is in order to treat Mechas as living by
+		instead considering their pilot for the hate value. */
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/ordeal/indigo.dm
```diff
@@ -349,14 +349,28 @@
...
-			return 100
-		hate_value += 1
-		if(iscarbon(L))
 			hate_value += 10
-		if(L.health <= (L.maxHealth * 0.6))
-			hate_value += (10 * ( L.health / L.maxHealth ))
+		//Highest possible addition is + 9.9
+		if(iscarbon(L))
+			if(L.stat != DEAD && L.health <= (L.maxHealth * 0.6))
+				hate_value += (10 * ( L.health / L.maxHealth ))
```
LanceSmites328, line 361: wait, isn't this math wrong?
This means the lower health you get, the lower your "hate value".
Having 100 health out of 100 health means you get 10x1, so +10.
Having 50 out of 100 health means you get 10x0.5, so +5.
```suggestion
				var/upper = L.maxHealth - HEALTH_THRESHOLD_DEAD
				var/lower = L.health - HEALTH_THRESHOLD_DEAD
				hate_value += min( 2 * ( 1 / ( max( lower, 1 ) / upper ) ) ), 20)
```
This serves to include the death threshold as well as making it now a range of ~3.3 (when it starts) to 20 (when it caps at 10% hp)

LanceSmites328 (changes requested):

code/modules/mob/living/simple_animal/hostile/ordeal/indigo.dm
InsightfulParasite: No your right higher health would be prioritized the way i used to do it.

LanceSmites328 (approved):

EgorDinamit (approved):
You should do something similar for other sweeper mobs.

PR #1385, opened 2023-09-11
Drifting Fox - (Draft)

Moribox:
## About The Pull Request

If it functions as I need it to, SHOULD put the bare-bones framework im throwing together for fox into its own PR instead of trying to put it in with the wallshove one.

## Why It's Good For The Game

Not sure really, more abnos is always neat though.

## Changelog
:cl:
- Really just creates the PR so I have somewhere to throw my work.
- And receive help with it because I am nowhere near confident in the work I do.
/:cl:


JackFrost2424:
I know this is a draft but here's some things I noticed. The attachment work rates are FUCKED. Work rates should never go above 60 lest it's a ZAYIN with six PE boxes, the work always lowers the counter, or it kills you. And it doesn't seem you're doing a reverse stat check or a death thing, so I would lower that. With how it is now, you can get 100% success rate way too easily. 

Work damage is kinda wimpy for a HE, to be honest. Should probably be 10 instead. 

A three counter is too high for what seems to be an abno that only lowers if you fail a stat check or get a bad. I would at least add in a "lowers sometimes on neutrals" and/or lower the counter. 

1000 HP for a fast moving HE seems fine but without seeing the breach mechanics, I can't comment. 2.5 move delay sounds weird, though. Just settle on 2 or 3.

I'll leave the other things for the others, those are just what stuck out to me.

Moribox:
Ah I appreciate the feedback, though In the current code I have yet to sync I should mention that the work rates are way lower as the idea is to have a list that tracks who has pet the fox, and those who do get work rates around 50-60 and those who don't get closer to 20-30. I will take all of your advice in mind though!

code/modules/mob/living/simple_animal/abnormality/he/drifting_fox.dm
```diff
@@ -0,0 +1,81 @@
...
+	ego_list = list(
+		/datum/ego_datum/weapon/sunshower
+		/datum/ego_datum/armor/sunshower
+	)
+	gift_type = /datum/ego_gifts/sunshower // NEED TO ACTAULLY MAKE THE GIFT / EGOS
+
+/mob/living/simple_animal/hostile/abnormality/drifting_fox/FailureEffect(mob/living/carbon/human/user, work_type, pe)
+	datum_reference.qliphoth_change(-1)
+	// "borrowed" from forsaken murderers code to give a punishment for failing to do well.
+/mob/living/simple_animal/hostile/abnormality/scarecrow/PostWorkEffect(mob/living/carbon/human/user, work_type, pe, work_time)
```
Coxswain-Navigator, line 49: Scarecrow?

code/modules/paperwork/records/info/he.dm
```diff
@@ -422,3 +422,13 @@
 		"Agents who witnessed the movements of Wayward Passenger reported unusually high levels of mental corruption.",
 		"When Wayward Passenger breached containment, it created spacial anomalies, dubbed T-01-170-1.",
 		"Agent Alex who entered T-01-170-1 found that it aided the suppression process.")
+
+//Drifting Fox
+/obj/item/paper/fluff/info/he/drifting_fox
+	abno_type = /mob/living/simple_animal/hostile/abnormality/drifting_fox
+	abno_code = "Placeholder"//No idea how to get this, I know there is a schtick to it but gonna need assistance.
```
Coxswain-Navigator, line 429: Use O-02-173

PR #1396, opened 2023-09-12
Handicap Accessible Work

LanceSmites328:
## About The Pull Request
Changes work console's output to audible message and adds an output for deaf people. This means it prioritizes being audible, and if you can see it but not hear it it'll still show you the output because you can *see the screen*.

## Why It's Good For The Game
Makes work console output visible to deaf and blind people equally.


Kitsunemitsu (approved):
Sure, looks good to me.

PR #1401, opened 2023-09-14
Allows Potential Machines to function with empty hand.

civilCornball:
## About The Pull Request

This pull request causes Potential Machines (citymap) to only function when the player clicks on them with an empty hand.

Before now, players had to have some kind of object in their hand.

(Do check my code, though. I'm not sure of the difference between `attack_hand(mob/living/user)` and `attack_hand(mob/user)`.)

## Why It's Good For The Game

Confuses players less. Why would you need to have, say, your ID in your hand in order to scan?

## Changelog
:cl:
tweak: Potential Machines now want empty hand.
/:cl:


civilCornball:
...god damn it why is it including my other planned PR

How do I fix this?

LanceSmites328 (changes requested):
Can you re-add the attackby proc and just have both attack_hand and attackby point to a new proc that has the scanning check in it? That way we preserve old functionality and also add new functionality

Kitsunemitsu (changes requested):
I'd add it to both tbh.

civilCornball:
I'm working on the 'combine into single proc' part, but something detonated my .dme

If you're alright with it being untested, lemme know.

civilCornball:
update: it's been tested, but now my VSC **and** fork are refusing to push changes
great!

LanceSmites328 (approved):

Kitsunemitsu (approved):
Yeah looks good to me

PR #1405, opened 2023-09-14
PE Report Update

LanceSmites328:
## About The Pull Request
image
Adds a check for uninitialized quota, and if so adds different text.

## Why It's Good For The Game
Fixes text that doesn't make sense.


Kitsunemitsu (approved):
Looks good to me

PR #1407, opened 2023-09-15
Adds Persistence to Agents and PE Quota

LanceSmites328:
## About The Pull Request
Adds tracking to and of the following:
PE Quota
* Was PE Quota reached?
* Was it reached but all of it was spent?
* How much PE was left over if Quota was completed?

Agents (Technically all humans)
* Points per ordeal survived
* Points for quota
* Points for surviving with quota reached

Creates a CONFIG entry for Agent max Rep/Points.

Drains all leftover Available Boxes into Box Goal as well.

This persistence doesn't do anything *yet* and will be added in a later PR.

## Why It's Good For The Game
Allows us to grant bonuses and penalties for a variety of situations based on previous round performance and should make PE Quota more impactful instead of just "Free Pizza" and nothing else.

Potential bonuses:
Keep a portion, perhaps 10%, of previous round PE as leftover for the new one, allowing an early start on refinery or some early goodies.
Partially prepaid representatives
Bonus starting PE to the first abno arrived
Bonus quirks to Agents


code/controllers/configuration/entries/game_options.dm
```diff
@@ -139,6 +139,11 @@
 	min_val = 0
 	max_val = 1
 
+/datum/config_entry/number/agent_rep_maximum
```
Kitsunemitsu, line 142: Uncap this, we can use it for a scoreboard

Kitsunemitsu (dismissed):
One tiny change

Kitsunemitsu (approved):
Issue fixed.

PR #1409, opened 2023-09-15
Fix training core debuffs & buffs

EgorDinamit:
## About The Pull Request

Training core, after Cox's attribute update, was changing the "bonuses" and not buffs. This PR fixes it.

## Why It's Good For The Game

Training core is meant to affect attributes, not max hp and other "bonuses"


Coxswain-Navigator:
I suggest simply raising the lowered bonus rather than changing it back to buffs, as people being able to use E.G.O. makes for more interesting gameplay.

LanceSmites328 (approved):

PR #1410, opened 2023-09-16
Reworks Death Stare's ability

CrabbytheCrab:
## About The Pull Request

Made the cooldown of the ability 30 seconds but reverted the nerfs. buffed slow down to a 50% slow with a 4 second duration. You also throw the cocoon instead of it spawning on where you click.

## Why It's Good For The Game

This is meant to address the abilities spamablity which let the high hp be even higher since you could place one down every 15 seconds. This should hopefully fix that and make its slow more impactful

## Changelog
:cl:
add: Added new things
del: Removed old things
tweak: tweaked a few things
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

### DIFF
--- a/code/modules/spells/ability_types/realized_aimed.dm
+++ b/code/modules/spells/ability_types/realized_aimed.dm
@@ -320,16 +320,16 @@
 
 /obj/effect/proc_holder/ability/aimed/cocoon_spawn
 	name = "Cocoon summon"
-	desc = "An ability that allows its user to summon a cocoon to take hits and slow and damage enemies near it."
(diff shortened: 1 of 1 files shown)


LanceSmites328:
The reason the slow isn't very impactful is it's a +0.5 to move_to_delay, not a "50% increase"

PR #1369, opened 2023-09-08
Removes wallshove knockdowns

Moribox:
## About The Pull Request
This simply makes it when a player shoves another player / humanoid into a wall, they no longer get knocked down.

## Why It's Good For The Game
The combat isn't really made with this sort of stun in mind, and wallshoves have on several occasions resulted in scenarios where a fight that should otherwise be in one sides favor is instantly lost to another side due to " oh well i'm stunned guess I die now."

## Changelog
:cl:
del: Removed the ability to knock down players / humanoids with wallshove.
/:cl:

PR #1417, opened 2023-09-17
Nerfs sea-pushing

Lexen74:
## About The Pull Request

This might be a scuffed way to do this, i'll admit, but such is the way of answering a WYCI. 

Deep water deals hilariously high damage capable of putting most players into softcrit instantly, on top of inflicting paralysis just long enough for anyone who survives the initial damage to be finished off regardless. This isn't much of an issue in LC13 where PvP is a rare occurence, but with the addition of LOR13 this has become an issue and has been to instantly dispatch people regardless of attributes and gear.

This is a problem on its own, but water has also been used as a way to keep dead people permanently out of reach of revival by throwing the corpse away from shore, as its near-fatal properties make wading in it to retrieve a body impossible unless the body is next to solid ground.

Perma-kills are fine, we wouldn't have a gibber otherwise, but this specific method takes no effort outside of activating harm intent and running into someone until they are pushed in a water tile, while the gibber requires you to have already killed or critted the unlucky victim beforehand. Moreover, we've been trying to move away from so-called "cheap" SS13 tactics that allow someone to gain a severe combat advantage with low effort such as wallshoves (#1369), and sea-pushing fits that category.

## Why It's Good For The Game

Stops rats from being able to win 1v1s with every character under the sun and making them unreviveable by inviting them to dip their toes in the ocean.

## Changelog
:cl:
balance: Deep water damage reduced from 100 brute and 50 oxy per step to 10 brute per step, and knocks players instead of paralyzing them.
/:cl:

### DIFF
--- a/ModularTegustation/fishing/code/turfs.dm
+++ b/ModularTegustation/fishing/code/turfs.dm
@@ -67,14 +67,13 @@
 		var/mob/living/L = thing
 		if(L.movement_type & FLYING)
 			return
-		//100 brute damage to living mobs. If they are human add 50 oxygen damage to them.
-		L.adjustBruteLoss(100)
+		//20 brute damage to living mobs. Adds up pretty quick if you're not warped away.
+		L.adjustBruteLoss(30)
 		if(ishuman(L))
 			var/mob/living/carbon/human/H = L
-			H.Paralyze(30)
-			H.adjustOxyLoss(50)
-			visible_message("<span class='boldwarning'>[H] sinks into the deep!</span>")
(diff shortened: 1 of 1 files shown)


TravisAngeI:
Pretty cool, still leaves water to be an environmental hazard but it's not a fight winner.

InsightfulParasite:
This is a much better system than what i had initially. I approve as the horrid creator of the deep tile.

Lexen74:
I coded this while sleep deprived and put _30_ brute while saying it deals 10 in the CL. 30 brute being death in 4 steps. 
10 really has to be the maximum damage to avoid people making bodies irrecoverable, even then getting some of the deeper tiles will still require you to heal halfway through to make it.

Kitsunemitsu:
It stuns you and nearly kills you so that you cannot jump into water to escape humans and abnormalities.

There is a doctor on LOR13; and the only people that really are near the beach is fishermen, someone where like
Why would you kill them?

PR #1420, opened 2023-09-18
Work Type Sublist

LanceSmites328:
## About The Pull Request
Adds a new variable to abnormalities that goes along-side their Work Chances, it's a list that associates work type with attribute type. This list defaults to the one that's always been used.
Also changes a little bit of logic so that works that don't have an attribute still do the following:
* Produce PE
* Get Logged
* Increase Overload

## Why It's Good For The Game
This allows for custom works to train attributes if they're supposed to, as well as allowing for logging of all work types. The only real issues that would exist is if someone makes a work that actually has a 100% success rate with 0 effect outside of simply being a work, which does not currently exist. The closest we have is Shepard and he breaches when Release work is performed.

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
@@ -47,6 +47,8 @@
 							ABNORMALITY_WORK_ATTACHMENT = list(50, 55, 60, 65, 70),
 							ABNORMALITY_WORK_REPRESSION = list(50, 55, 60, 65, 70)
 							)
+	/// Work Types and corresponding their attributes
+	var/list/work_attribute_types = WORK_TO_ATTRIBUTE
 	/// How much damage is dealt to user on each work failure
 	var/work_damage_amount = 2
 	/// What damage type is used for work failures
--- a/code/datums/abnormality/datum/abnormality.dm
+++ b/code/datums/abnormality/datum/abnormality.dm
@@ -144,7 +144,7 @@
 	if(!console?.recorded && !console?.tutorial) //only training rabbit should not train stats
 		return
 	if(pe > 0) // Work did not fail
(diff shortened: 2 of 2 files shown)


code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
```diff
@@ -47,6 +47,8 @@
 							ABNORMALITY_WORK_ATTACHMENT = list(50, 55, 60, 65, 70),
 							ABNORMALITY_WORK_REPRESSION = list(50, 55, 60, 65, 70)
 							)
+	/// Work Types and corresponding their attributes
+	var/list/work_attribute_types = WORK_TO_ATTRIBUTE
```
Kitsunemitsu, line 51: Shouldn't this be a list?

Kitsunemitsu (changes requested):

code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
LanceSmites328: It IS a list, WORK_TO_ATTRIBUTE is a list defined in attributes.dm
```
#define WORK_TO_ATTRIBUTE list(\
							ABNORMALITY_WORK_INSTINCT = FORTITUDE_ATTRIBUTE,\
							ABNORMALITY_WORK_INSIGHT = PRUDENCE_ATTRIBUTE,\
							ABNORMALITY_WORK_ATTACHMENT = TEMPERANCE_ATTRIBUTE,\
							ABNORMALITY_WORK_REPRESSION = JUSTICE_ATTRIBUTE,\
							)
```

Kitsunemitsu: Oh what the fuck

Kitsunemitsu (approved):
Code looks good.

PR #1423, opened 2023-09-18
Implements the snow queen arena

Coxswain-Navigator:
image

## About The Pull Request
Adds the snow queen map in its final state. Not a complete feature by any means, but should be in a state that is usable for the developer without needing to do any mapping. 

## Why It's Good For The Game
We can finally get snow queen

## Changelog
:cl:
tweak: the snow queen map now works
/:cl:


Kitsunemitsu (approved):
Yeah sure

PR #1434, opened 2023-09-20
More Manager Choices

LanceSmites328:
## About The Pull Request
Gives Managers more control over the abnormalities in their facility.
Does this by:
* Giving them the ability to choose from all abnormalities of all available ranks.
* Making this choice cost PE
* Giving PE if it pushes the round forward by picking an abnormality of equal or higher rating to the current highest automatic threat level able to be chosen. (If the game can give you Teths, choosing a Teth or HE grants you PE, while a Zayin would cost PE)
* Lets the manager choose abnormalities of a threat rank higher earlier, and lets a manage manually choose to take weaker abnormalities later at the cost of PE.

It also introduces inherited cooldown for manually picking abnormalities. For example, if you pick SoW on one round, it can't be manually picked for the next 2 rounds. It can, however, still be randomly rolled via Yellow roll or the initial RNG.

Also adds reputation penalties for not meeting Quota and further penalties for spending all of your quota instead.

## Why It's Good For The Game
Allows the Manager to have more choice, including pushing the game forward faster, while also adding a tangible use to PE for the Manager and a method of encouraging managers to push the game forward. 

The inherited cooldown on picking an abnormality between rounds is designed to push diversity in rounds instead of just going for a meta group every round. With this feature, it would require at least 3 separate "metas" which I think is a decent amount so far.

### DIFF
--- a/code/__DEFINES/abnormalities.dm
+++ b/code/__DEFINES/abnormalities.dm
@@ -52,3 +52,4 @@
 #define PE_GOAL_REACHED	"goal_reached"
 #define PE_GOAL_SPENT	"goal_spent"
 #define PE_LEFTOVER		"leftover_pe"
+#define USED_ABNOS		"used_abnos"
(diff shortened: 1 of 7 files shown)


LanceSmites328:
Abno Weights Solved This.

PR #1442, opened 2023-09-22
Removes Release Work On Yin

LanceSmites328:
## About The Pull Request
Removes the copy-pasta work that doesn't do anything but likely make you fail due to bad result.

## Changelog
:cl:
code: removed release from the work chances
/:cl:


EgorDinamit (approved):
Seems more like a fix than a tweak.

PR #1444, opened 2023-09-22
Makes Violet Midnight not Racist

LanceSmites328:
## About The Pull Request
At the moment it only targets humans and specifically humans with clients. Why? This lets 75% of the ordeal be solo'd be any of the following:
Blue Smocked Shepard
Forsaken Murderer
Pygmalion
Shrimp Grenades
Shrimp Summons
Queen of Hatred
Servant of Wrath
Yang
A dog

Literally any abnormality that randomly breached because this lacks the hostile faction.

This PR does ~two~ a few things:
Sets every reference of a human being to a living mob reference
Adds a CanAttack check to all damage and targeting procs, which already checks for faction and godmode.
Adds an AFK check to CanAttack for any mob with a Ckey, meaning some people may be spared from lag.
Gives Violet Midnight the "hostile" faction

## Why It's Good For The Game
Makes Violet Midnight not a shrimp related joke.

## Changelog
:cl:
code: added CanAttack checks to violet midnight
code: changed all human checks into living ones.
code: Set the stat_attack var for Violet Midnight to HARD_CRIT
code: Adds an AFK check to CanAttack
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -262,6 +262,9 @@
 		var/mob/M = the_target
 		if(M.status_flags & GODMODE)
 			return FALSE
+		if(M.ckey)
+			if(M.client?.is_afk()) // AFK protection
+				return FALSE
 
 	if(see_invisible < the_target.invisibility)//Target's invisible to us, forget it
 		return FALSE
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/violet.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/violet.dm
@@ -229,7 +229,7 @@
 	if(world.time > ability_cooldown)
 		INVOKE_ASYNC(src, .proc/StartAbility)
 
-/mob/living/simple_animal/hostile/ordeal/violet_midnight/CanAttack(atom/the_target)
+/mob/living/simple_animal/hostile/ordeal/violet_midnight/AttackingTarget(atom/attacked_target)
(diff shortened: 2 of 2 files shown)


EgorDinamit:
It cared for client so as to prevent AFK players(i.e. your internet died for a moment) from being unfairly killed.
Ideally abnormalities themselves should not be targeting violet midnight, as it has miniscule damage all things considered, plus it makes it too easy on agents.

LanceSmites328:
> It cared for client so as to prevent AFK players(i.e. your internet died for a moment) from being unfairly killed. Ideally abnormalities themselves should not be targeting violet midnight, as it has miniscule damage all things considered, plus it makes it too easy on agents.

I previously intended to also give it the hostile faction, would this suffice for that?

EgorDinamit:
Hostile faction would help with some(most) cases, but I wonder if we could make a trait/proc to make certain mobs untargetable entirely.

LanceSmites328:
> Hostile faction would help with some(most) cases, but I wonder if we could make a trait/proc to make certain mobs untargetable entirely.

Create TRAIT_MOB_UNTARGETABLE
Add check into CanAttack (if(target.has_trait(TRAIT_MOB_UNTARGETABLE) return FALSE)

EgorDinamit:
Oh, I actually did add a reverse version of CanAttack, it is used on WN's apostles when they are dead. You can probably copy it for violet midnight.

Kitsunemitsu (dismissed):
Looks good to me

Kitsunemitsu (approved):
Minor change, re-approving.

PR #1455, opened 2023-09-23
LoR13 water tweaks

Lexen74:
## About The Pull Request

Adds a check to deep water tiles making it so that they do not drown you if the map-type is LoR13, instead allowing you to wade through them.
Unlike my previous PR #1417 this will change nothing for other round types, as this appears to have been the main issue with it.

This also replaces the freshwater near the beach by saltwater. The freshwater tiles behind the clinic remain untouched, making all fish types accessible in LoR13. 

## Why It's Good For The Game

A less shitcode way to prevent people from using the sea as an easy kill and permanent body disposal ground in a way that won't affect LC13, R-Corp or W-Corp rounds.
Fishing offices can now really get on their gotta-catch-em-all grindset.

## Changelog
:cl:
tweak: Cityfolk have taken swimming lessons and can safely wade accross deep water.
fix: The sea has been re-salted.
/:cl:


Lexen74:
Hol' up. The sea already shows up as saltwater on the live server, is something tm'd that i failed to notice?

Kitsunemitsu (changes requested):
I can't approve of this here, as it is just the previous pr but on the city.
Fighting people near the beach should be avoided, plain and simple.

Lexen74:
> I can't approve of this here, as it is just the previous pr but on the city. Fighting people near the beach should be avoided, plain and simple.

I understand this much, but people end up fighting there anyway due to the location of the beach being near HHPP as well as farming fish for money and being mugged by rats.

I also want to note that the main purpose of this PR isn't to remove a murder option but to _stop people from using the sea as effortless body disposal._ Once a body is one tile away from the shore it's incapable of being moved and thus impossible to revive, in a mode where having to respawn comes with a massive stat penalty.

This is not an intended feature of water tiles, this is essentially an exploit that has already caused issues in multiple rounds. I do not believe that "people should fight away from the sea" justifies not fixing it.

Kitsunemitsu:
Just add in a dropoff point. There was a PR to do this in the past.

Lexen74:
I'll try and see if i can do that.

PR #1462, opened 2023-09-24
Prevents Multiple Actions from Justice Panic Controller

LanceSmites328:
## About The Pull Request
Adds a check to see if the insane mob is already doing the breach action to prevent it stacking MoveTo calls and having the mob move multitudes faster than it should.

## Why It's Good For The Game
Fix

## Changelog
:cl:
fix: fixed justice panic moving multiple spaces at once and instant breaching things
/:cl:


Kitsunemitsu (approved):
Looks good to me.

PR #1465, opened 2023-09-24
Null Loot fix

LanceSmites328:
## About The Pull Request
Loot was sometimes null, this caused runtimes when drop_loot was called due to `null.len`. This fixes that.

## Why It's Good For The Game
Fixes runtime


Kitsunemitsu (approved):
Singly line fix.

PR #1469, opened 2023-09-24
Timer soft removal

Kitsunemitsu:
## About The Pull Request
Adds a temp check to the timers.

## Why It's Good For The Game
Work timers are really inconsistent and can't be trusted, nor can we expect players to calculate workrates.
This is just a catch for them that's more consistent

Coxswain approved.

## Changelog
:cl:
balance: Adds hard temp checks to timer checks.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/aleph/blue_star.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/aleph/blue_star.dm
@@ -111,7 +111,9 @@
 		datum_reference.qliphoth_change(-1)
 	if(user.sanity_lost)
 		datum_reference.qliphoth_change(-1)
-	if(work_time > 40 SECONDS)
+
+	//Safety, just in case for some reason you can't hit the timer as it's inconsistent and hard to judge
+	if(work_time > 40 SECONDS && get_attribute_level(user, TEMPERANCE_ATTRIBUTE) > 100)
 		datum_reference.qliphoth_change(-1)
 		playsound(src, 'sound/abnormalities/bluestar/pulse.ogg', 25, FALSE, 28)
 		user.death()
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/simple_animal/abnormality/aleph/blue_star.dm
```diff
@@ -111,7 +111,9 @@
 		datum_reference.qliphoth_change(-1)
 	if(user.sanity_lost)
 		datum_reference.qliphoth_change(-1)
-	if(work_time > 40 SECONDS)
+
+	//Safety, just in case for some reason you can't hit the timer as it's inconsistent and hard to judge
+	if(work_time > 40 SECONDS && get_attribute_level(user, TEMPERANCE_ATTRIBUTE) > 100)
```
kesumod, line 116: Is this supposed to only go off while the agent took too long AND had a temperance ABOVE one hundred?

code/modules/mob/living/simple_animal/abnormality/teth/meat_lantern.dm
```diff
@@ -121,7 +121,7 @@
 	..()
 
 /mob/living/simple_animal/hostile/abnormality/meat_lantern/PostWorkEffect(mob/living/carbon/human/user, work_type, pe, work_time, canceled)
-	if(work_time < 18 SECONDS)
+	if(work_time < 18 SECONDS && get_attribute_level(user, TEMPERANCE_ATTRIBUTE) <= 60)
```
kesumod, line 124: similar thing to my last comment, is it supposed to be if yo work too fast WHILE having bad temperance?

code/modules/mob/living/simple_animal/abnormality/aleph/blue_star.dm
EgorDinamit: This is a very weird part of code that makes no sense.

Coxswain-Navigator: Sounds like she meant to put an "OR"

Coxswain-Navigator:
A better solution would be to check for temperance + temperance bonus. Not only is that easier for players to understand than work speed, but it's also closer to the original intention.

Kitsunemitsu:
Not interested in this PR any further.

PR #1474, opened 2023-09-26
Brass Nail Fix

Kitsunemitsu:
## About The Pull Request
Makes it so you can only charge a hammer once with brass nails.

## Changelog
:cl:
fix: fixed infinitely stacking Ncorp weapon bug
/:cl:


Kitsunemitsu:
I was meaning to fix this, never did. 
City is out of rotation anyways; this shouldn't cause any issues.

PR #1475, opened 2023-09-27
Manager Camera now has Following!

InsightfulParasite:
## About The Pull Request
I butchered AI code to make it a feature for the manager camera. Testing shows that it essentially works but we should test it first due to the tracking datum being shared by sephirah and managers. 
DRV and meat lantern have a false on their track so they wont be able to be followed by the managers ability.
Two creatures with the same name will show up the same on the managers follow choices. It is randomized who you end up following. Dead agents will need to be either disfigured or cleaned up to prevent the system from getting confused.

Doubleclicking to follow code is hardcoded into AI mobs so unless we are allowed to make a signal for double click, which may be weird since we already have a signal for clicking, we may have to use more roundabout solutions.

## Why It's Good For The Game
Ive been requested to code this by a few manager players. Quality of life even if it is very rough in its development.

## Changelog
:cl:
add: follow ability for manager console
/:cl:

### DIFF
--- a/code/game/machinery/computer/manager_camera.dm
+++ b/code/game/machinery/computer/manager_camera.dm
@@ -22,8 +22,6 @@
 	var/commandtype = 1
 	var/command_delay = 0.5 SECONDS
 	var/command_cooldown
-	var/mob/living/tracking_subject
-	var/tracking = FALSE
 	///Variable stolen from AI. Essential for tracking feature.
 	var/static/datum/trackable/track = new
 	var/static/list/commandtypes = typecacheof(list(
@@ -64,6 +62,7 @@
 /obj/machinery/computer/camera_advanced/manager/GrantActions(mob/living/carbon/user) //sephirah console breaks off from this branch so any edits you want on both must be done manually.
 	..()
 
+	//List abilities here:
 	if(cycle)
 		cycle.target = src
 		cycle.Grant(user)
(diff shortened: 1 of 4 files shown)


Gboster-0:
there is a good chunk of single-letter variables and bare-bones comments on this, but otherwise looks good

code/game/machinery/computer/manager_camera.dm
```diff
@@ -233,6 +244,88 @@
...
+
+	return targets
+
+/obj/machinery/computer/camera_advanced/manager/proc/ActualTrack(mob/living/target)
+	if(!istype(target))
+		to_chat(current_user, "<span class='warning'>Invalid Tracking Error.</span>")
+		return
+
+	tracking_subject = target
+	tracking = 1
```
Gboster-0, line 282: cant this be a `TRUE` instead of `1`?

Kitsunemitsu:
I'm willing to merge this until a better solution is achieved!

InsightfulParasite:
Simplified the code with a orbit proc that ghosts use. Downside is that the camera no longer loses targets automatically.

InsightfulParasite:
I have two commits. The first one is the old code that used AI role code and the second is the new simplified one using orbit code. Im hesitant on merging both together since someone may say that the old code is better suited.

Kitsunemitsu (approved):
Saw video footage, works excellently

PR #1478, opened 2023-09-28
Adds library maps (Except Kether)

LancashireladOnGitHub:
As this implies, all maps are added  bar Kether
labelled as
floor_history
floor_techsci
floor_literature
floor_art
floor_naturalsci
floor_socialsci
floor_language
floor_philosophy
floor_religion

## About The Pull Request

Adds a collection of maps for admins to use instance spawn at their leisure (Note I was told instance spawn was a thing, if its not, close this PR and we will forget it exists until some way comes that these maps can be used)

The maps themselves will be spawned in by the admins via instance spawn, Z-level spawn, whatever mechanism LC13 uses to load map files into an existing session.

From there the admins will have creative freedom to decide the equipment, combat scenario, rewards etc, without worrying about the arena itself.

## Why It's Good For The Game

At the moment we have one main type of event that the admins like to run on LoR for example, that being "distortions" and while thats fine and all, it can get extremely repetitive and downright annoying for some people who dont want anything to do with it, these maps aim to do two things

1. Provide an area for admins to provide a challenge to players willing to enter, without the collateral damage of anyone on the outside getting upset at being dragged unwillingly into a situation they didnt ask for

2. Provide more varied scenario's and events that will generally breathe a bit of life in LoR and maybe LC13 itself, since these will be admin spawn maps, its mostly up to admin discretion on what happens.


LanceSmites328:
Fix changelog, also these are so good

LanceSmites328 (approved):
Doesn't change anything old and they're nice maps

PR #1485, opened 2023-10-01
fix backstreet room spacetiles

Retrino:
## About The Pull Request

fix a space tile under the money backroom prefab and replace the airless tiles in the church with normal air'd up tiles

## Why It's Good For The Game

air is probably good to have


Kitsunemitsu (approved):
Looks great

PR #1496, opened 2023-10-04
Adds Purple Tear Weapons

TheBronJameOffical:
## About The Pull Request
Adds in Purple Tear weapon
Contains 4 stances that can be switched between relatively quickly. (Slash, Pierce, Blunt, Guard)
Every time stance is changed you gain a 1.5 force boost on the first hit
Each weapon can also be Sheathed in order to have a special effect on your next attack. (Effects are based off of LoR cards)

- Slash:
Fast hitting red weapon
Sheathe your blade for that tasty Power effects this die x2

- Pierce:
Normal speed white weapon
Sheathing makes your next strike inflict double damage taken to the target for 5 seconds

- Blunt:
Big boy black damage at slow speed
Sheathe and hit to apply a buff to yourself that decreases all damage taken by half for 4 seconds (Better make it count)

- Guard:
Weak ass red damage, you're not really supposed to fight in this stance anyways.
Comes with a block similar to Old Boys from Black Silence, a successful block will make the next hit do True Damage

Lastly, it comes with Mirage Storm which will charge as you hit enemies, requiring 40 charge to use the ability.
Using it will have you attack with the Slash, Blunt, and Pierce stance weapons in quick succession.

## Why It's Good For The Game
John Furioso Forma de Purple

## Changelog
:cl:
add: Added PT Weapons
soundadd: PT Sounds
/:cl:


Coxswain-Navigator:
What happened to this?

TheBronJameOffical:
> What happened to this?

think i fucked up trying to rebase it and can't go back, i've moved it to another branch and will open it there when its ready.

PR #1499, opened 2023-10-05
Adds Weapon Sweeping and Thrusting

LanceSmites328:
## About The Pull Request
Adds the ability for most weapons to make Sweeping attacks or Thrusting attacks. Both of these require the user to be in Harm intent. If a target is clicked on, these are _purely_ cosmetic.

Sweeping attacks occur when using non-shield, non-spear, and EGO without reach, and swing from whatever hand they're in, towards the opposite side. The first target in it's path takes the attack.

Thrusting attacks occur when using an item with reach and thrust out from the user, hitting targets closest to farthest.

Both prioritize hostiles, standing people, then resting people.

This DOES NOT:
* Allow weapons to hit more than they normally can
* Increase damage
* Ensure you don't murder your friend.

VFX is colored for handedness in these videos.

https://github.com/vlggms/lobotomy-corp13/assets/61567407/6c6e1797-bfea-4153-8122-2ff083321fc9

https://github.com/vlggms/lobotomy-corp13/assets/61567407/742db72b-fa95-4c02-93f2-a53b51778ff1

### Greatswords and Other Large Weapons
Sweeps variable turns from TRUE/FALSE to a value, denoting how many turfs it hits around the player.
Sweeping weapons hit 1 mob per turf covered.
Default melee weapons have 1 turf sweep, reducing the forgiveness this PR gives by default to that 1 tile.

## Why It's Good For The Game
Gives dynamic leniency to melee weapons that ranged weapons get built in.

## TODO

- [x] Sprites for the swings would make this look infinitely better, current visualization utilizes `do_attack_animation` which works but isn't perfect.

## Changelog
:cl:
code: added `sweeps` variable to items, FALSE by default. If TRUE, an item can make a Sweeping attack
code: Added Sweep proc and supporting procs to cleanly calculate turfs and perform animations.
/:cl:


LanceSmites328:
Player-Testing has gone well overall. Decent positive response from what I've seen and I've addressed every issue I know of so far. (Was TM'd for 3-4 rounds)

EgorDinamit:
Could you try recording a video/gif of it in action?

Kitsunemitsu:
I'd like to gain my full thoughts on this before moving forwards, as it's a massive change.

EgorDinamit (dismissed):
This is overall a very good change.
Could you perhaps add variables for items to have unique/personalized sweeping/thrusting effects? Also, the attack message should *probably* specify that you "thrust at the floor" with a weapon.

Kitsunemitsu:
I would like to take a look at this one personally before merge.

Kitsunemitsu (changes requested):
I will insist that you make it a var on how many tiles to swing, and set the default to 1.

This increases the margin of error for melee weapons too much, increases FF unnecessarily and is overall a pretty massive buff to something to melee weapons, which I'm not really interested in giving an overall buff at this time.
I want to reduce pixel hunting without making melee weapons being the equivalent of "hitting the broadside of the barn". Implementing it this way makes it a positive for a weapon to have a multitile swing as well; allowing for more variety in melee gameplay.

Kitsunemitsu:
We spoke about bringing this PR back

Kitsunemitsu:
We had issues to where if you didn't have the stats you could still swing (just visually)

Kitsunemitsu (approved):
Love this PR, Going to get to work on updating weapons.

PR #1503, opened 2023-10-06
remove wall blocking 27th march

Retrino:
## About The Pull Request

unblocks 27th march containment unit which previously had a wall in the doorway (on eta)

## Why It's Good For The Game

the agents must cower


Kitsunemitsu (approved):
Approved.

PR #1504, opened 2023-10-07
The great out of space rework of 2023

CrabbytheCrab:
## About The Pull Request

Heavily reworks out of space(loos' weapon)
Now it either does black or white(always does white to insane agents)
Dash looks cooler and no longer knocks you down when colliding with a player or wall(and new visual effects too)

## Why It's Good For The Game

Out of space was a neat idea just sucked in every way possible. doing both black and white isn't that useful due to most abnormalities that are weak to one usual resists or are immune/absorb the other. the dash is honestly horrid. Its easy to troll others by knocking them down with it and trying to use it normally is hard due to how easy it is to accidentally hit a wall or another player. these 2 changes are meant to address both of these problems making out of space a good aleph weapon by making it either black or white and making the dash suck less. Also nerfs the damage to 64 to compensate it being basically 2 weapons in one

## Changelog
:cl:
tweak: tweaked a out of space
balance: reworked out of space and nerfs the total damage to 64
code: changed some out of space code
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/aleph.dm
+++ b/code/game/objects/items/ego_weapons/aleph.dm
@@ -552,6 +552,7 @@
 							JUSTICE_ATTRIBUTE = 80
 							)
 	var/canaoe
+	var/damagemode = 0
 
 /obj/item/ego_weapon/space/attack_self(mob/living/carbon/user)
 	if(!CanUseEgo(user))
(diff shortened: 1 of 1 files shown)


Kitsunemitsu:
This literally does not change what you think it does

CrabbytheCrab:
what you mean?

Kitsunemitsu:
Half and half is just more consistent. If you think the damage is lackign bump it up to 90 total.

CrabbytheCrab:
on paper yea but in practice most alephs heavily resist once while being weak to the other buffing its damage to 90 would mean it would do only half or around that to those alephs which would be worse than a weapon that just does one of the 2. I wanted to make the damage change before hitting something but I wasn't sure how, if I did I think it could nerf the damage to 65 or 60 due to being 2 weapons in 1 might become too much.

CrabbytheCrab:
Made a list to show why doing both white and black damage is not good at all

Kitsunemitsu:
I'm going with a simpler solution to this.

PR #1509, opened 2023-10-08
Modular abnormalities folder, modularizes all abnormalities with 96x64 sprite dimensions, icon bot fixes

Gboster-0:
I expect this PR to be denied. But hey, worth a shot. Worst comes to worst this idea will be confirmed to not be viable

This PR needs light test-merging and if its observed to change ANY gameplay aspect, thats very much un-intended

This used to be a PR that modularized ZAYIN's, but i realized that would straight up cause a LOT of PR's to get icon conflicts, shifted it to abnormalities in the 96x64 sprite sheet instead to minimize icon conflicts.

## About The Pull Request

This PR Puts all abnormalities that use 96x64 icons into their own new modular folder, that includes: their code, their sprites, their sounds
By doing that all future icon bot checks will never fail on 96x64 abnormalities, and be 100% accurate

This PR only one of many PR's to come should it get merged, the intention is to eventually get all abnormalities in their modular folders so we can actually use the icon bot with how its intended and work easier on them.

## Why It's Good For The Game

This PR makes it MUCH easier to work on individual abnormalities since all their mob code, sounds and sprites are in 1 place instead of being scattered all over the code (cough cough, the sound folder with like 30-40 folders inside of it for abnormalities)

It also allows actual use of the icon bot without false positives, and thats neat. Since we will actually know if someone accidentally made an un-intentional change to an abnormality in their PR, along with reviewing sprites

Since it also takes icons away from the big folders, we basically delete all future icon conflicts

## Changelog
:cl:
code: abnormalities that use 96x64 sprites are now in their own modular folder. Nothing concerning players
/:cl:


ModularTegustation/modular_abnormalities/zayin/a_quiet_day/code/quiet_day.dm
```diff
@@ -5,7 +5,7 @@
 /mob/living/simple_animal/hostile/abnormality/quiet_day
 	name = "A Quiet Day"
 	desc = "An old weather damaged bench, it feels oddly nostalgic to you. Like a spring day at the side of a lake."
-	icon = 'ModularTegustation/Teguicons/48x48.dmi'
+	icon = 'ModularTegustation/modular_abnormalities/zayin/a_quiet_day/icons/48x48.dmi'
```
InsightfulParasite, line 8: Okay i thought at first you were doing like icons generalized to all zayins instead of a dmi for each abnormality. I worry that having a non generalized dmi for abnormalities would, similar to the sound effects, result in duplicate icons in two separate files. I honestly thought the file would be Zayin -> Icons, Sound, Code -> individual abnormality code. But since i dont truely understand why the elder coders made the code so decentralized like this im not sure if my suggestion is even good.

InsightfulParasite: I do feel like it would be better to have a generalized zayin_32x32.dmi instead of 8 separate 32x32.dmi's.

Gboster-0: because thats how modular files work, each of the zayin's counts as a seperate modular file.
This makes it so icon conflicts are basically impossible, and you dont have to look at anything except the modular file itself for all that the abnormality needs to function

Gboster-0: the only way to get an icon conflict in this way, is for 2 seperate spriters to decide that they are changing the same exact abnormality's sprite at the same exact moment.
Or for 1 spriter to change the main sprite, and a coder to add a new mechanic to the abnormality that uses a second sprite (for example adding a breaching sprite to an abnormality that does not have one)

LanceSmites328:
I think this is a pretty good idea but the biggest thing is that this would result in an entire restructuring of our current abnormality file system which I'd like Kirie and/or Egor's approval of

LanceSmites328:
notably this SHOULD result in less game load, if slightly, due to the issue of "if an icon state from a file is loaded, all icon states from that file are"

Gboster-0:
> I think this is a pretty good idea but the biggest thing is that this would result in an entire restructuring of our current abnormality file system which I'd like Kirie and/or Egor's approval of

eh, generally if it helps im fine with developers sticking to doing their own non-modular things with abnormalities. I'll just stick to seperatelly modularizing em as time goes on

Gboster-0:
> Not a hard "No", but I would probably prefer if only icons were "modularized". Every other file is in a place it belongs, and does not cause conflicts.

While yes, other files dont really cause as much conflicts as icons do. Would it not be better to go a few miles ahead and just fully modularize them?
Having all the files in one place and not searching the 25 named sound folders is fairly nice

EgorDinamit:
> While yes, other files dont really cause as much conflicts as icons do. Would it not be better to go a few miles ahead and just fully modularize them? Having all the files in one place and not searching the 25 named sound folders is fairly nice

No. Sounds belong in sound folder, icons belong in icons folder. The way this is structured is good as is, except for icons.
I said my word, and the only thing I approve of is moving icon states of each abno in separate dmi files to prevent conflicts.

Gboster-0:
> > While yes, other files dont really cause as much conflicts as icons do. Would it not be better to go a few miles ahead and just fully modularize them? Having all the files in one place and not searching the 25 named sound folders is fairly nice
> 
> No. Sounds belong in sound folder, icons belong in icons folder. The way this is structured is good as is, except for icons. I said my word, and the only thing I approve of is moving icon states of each abno in separate dmi files to prevent conflicts.

Fine, if thats your stance on the matter we can do it that way.
But i will say that this is an abomination from modular code perspective

Gboster-0:
le close, making a follow-up PR for icons only

PR #1511, opened 2023-10-09
Fixes Charge Weapons

scionalu:
## About The Pull Request

Charge weapons were being used by low leveled users because returning false or breaking ..() does nothing to prevent the rest of the code from running
So I added several extra checks if the parent returns FALSE

## Why It's Good For The Game

no more silly infinite charges

## Changelog
:cl:
fix: fixed charge weapons
/:cl:


LanceSmites328 (changes requested):
So, while this _would_ fix the issue, it's not the best way to do it by a long shot and will in all likelihood result in double messages. What _should_ be getting used is
```
. = ..()
if(!.)
	return
```

LanceSmites328 (approved):

PR #1513, opened 2023-10-10
Purple Tear Weapon

TheBronJameOffical:
Ok lets do this again, hopefully i dont rebase it into the void.

## About The Pull Request
Adds in Purple Tear weapon
Contains 4 stances that can be switched between relatively quickly. (Slash, Pierce, Blunt, Guard)
Every time stance is changed you gain a 1.5 force boost on the first hit
Each weapon can also be Sheathed in order to have a special effect on your next attack. (Effects are based off of LoR cards)

- Slash: (45 Red at 0.5 Speed)
Fast hitting red weapon
Sheathe your blade for that tasty Power effects this die x2

- Pierce: (90 White at 1 Speed)
Normal speed white weapon
Sheathing makes your next strike inflict double damage taken to the target for 5 seconds

- Blunt: (135 Black at 1.5 speed)
Big boy black damage at slow speed
Sheathe and hit to apply a buff to yourself that decreases all damage taken by half for 4 seconds (Better make it count)

- Guard: (90 Red at 1.5 Speed)
Weak ass red damage, you're not really supposed to fight in this stance anyways.
Comes with a block similar to Old Boys from Black Silence, a successful block will make the next hit do True Damage

Lastly, it comes with Mirage Storm which will charge as you hit enemies, requiring 40 charge to use the ability.
Using it will have you attack with the Slash, Blunt, and Pierce stance weapons in quick succession.
At the end it will put you into the slash stance and gives you the defense buff from the Blunt Stance.

## Why It's Good For The Game
Big day for gacha gamblers, new color weapon just dropped

https://github.com/vlggms/lobotomy-corp13/assets/91028605/0fc68c89-5f46-4d7d-9ae5-fb85104291dc

## Changelog
:cl:
add: Added Purple Tear Weapon
soundadd: added PT sounds
imageadd: added relevant Icons and Beam
/:cl:


Kitsunemitsu (approved):
Looks good to me. Approving.

PR #1514, opened 2023-10-10
A very questionable change to out of space

CrabbytheCrab:
## About The Pull Request

Probably not going to get in but just something I did for out of space out of curiosity

retains the reworked dash of the old pr

lowers damage to 40 black and white(80 total) but every third hit makes an aoe doing 20 black and white(40 total). The big aoe of the dash does 50 black/white damage still

## Why It's Good For The Game

Probably not but it would make the dash unable to stun you and another person if you collided with them and giving it a weak aoe every few attacks would help the theme of a jack of all trades with an aoe and dash to stand out more maybe probably not

## Changelog
:cl:
balance: rebalanced out of space again to have a small aoe every few attacks and the reworked dash
code: changed some out of space code
/:cl:


CrabbytheCrab:
wondering if I could of just kept the dash with this instead of making it slightly better at aoe

PR #1527, opened 2023-10-14
replaces ". = ..()" parent calls with "return ..()" where applicable without changing surrounding code. Forces some abnormality procs to be called by children

Gboster-0:
## About The Pull Request

changes ". = ..()" calls for "return ..()" where it can be changed without affecting surrounding code
forces initialize(), destroy(), life(), WorkComplete() and GiftUser() to be called by an abnormality if they implement custom code
i would have also forced people to call FearEffect() but meat lantern requires to not call parent to be stealthy

## Why It's Good For The ~Game~ Codebase

juuuuuuuust in case contributors forget to call . = ..() on some procs and have to spend time diagonising the issue, this makes it a lil bit faster to show up

## Changelog
:cl:
code: . = ..() calls were replaced with return ..()
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/you_strong.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/you_strong.dm
@@ -241,17 +241,15 @@
 /mob/living/simple_animal/hostile/grown_strong/Move(atom/newloc, dir, step_x, step_y)
 	if(status_flags & GODMODE)
 		return FALSE
-	. = ..()
-	return
+	return ..()
 
 /mob/living/simple_animal/hostile/grown_strong/AttackingTarget(atom/attacked_target)
 	if(status_flags & GODMODE)
 		return FALSE
-	. = ..()
-	return
+	return ..()
 
 /mob/living/simple_animal/hostile/grown_strong/Initialize()
-	. = ..()
+	return ..() // replace with . = ..() once the icon state code gets un-commented
 	//icon_state = pick("1", "2", "3")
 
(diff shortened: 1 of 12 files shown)


LanceSmites328 (approved):
I'll say a lot of the changes from `. = ..()` to `return ..()` aren't necessary for functionality, as `.` is the default return value of the proc (among other things that are out of scope), so doing `. = ..()` runs the original parent proc and sets the current return value to that. However this is an overall improvement in code legibility.

code/modules/mob/living/simple_animal/abnormality/he/you_strong.dm
```diff
@@ -241,17 +241,15 @@
...
 	if(status_flags & GODMODE)
 		return FALSE
-	. = ..()
-	return
+	return ..()
 
 /mob/living/simple_animal/hostile/grown_strong/Initialize()
-	. = ..()
+	return ..() // replace with . = ..() once the icon state code gets un-commented
 	//icon_state = pick("1", "2", "3")
```
EgorDinamit, line 253: Just remove it entirely; We don't need unused/commented out code blocks.

Coxswain-Navigator:
I suspect that this could cause some bugs, be careful with merging it.

code/modules/mob/living/simple_animal/abnormality/he/you_strong.dm
Gboster-0: if you say so, just removed the whole initialize override code since that comment was preety much the only reason it existed

EgorDinamit (approved):

PR #1554, opened 2023-10-19
AI Targeting Overhaul

InsightfulParasite:
## About The Pull Request

- [x] Made AI targeting less "pick from a list of everything in the room" to "pick from a weighted list of everything in the room".
- [x] Altered Patience and LoseTarget() a little so that they dont reaggro on the same mob.
- [x] Added explanations to how the targeting works and where people can override for their mobs.
- [x] Mobs hit by a bullet outside of their sight range will now investigate the area the bullet was fired correctly instead of walking 5 steps and then having their AI turned to idle and losing target.
- [x] Reduces idlenpcpool fire delay from 60 to 25

## Why It's Good For The Game
Value targeting can make it so that even if a high value target is in sight a monster can rule a person who is right next to them as a more important target.

## Changelog
:cl:
add: //label to simple mob flags in defines.
add: ValueTarget()
code: completely changes picktarget()
code: Hostile.dm
/:cl:

### DIFF
--- a/code/controllers/subsystem/idlenpcpool.dm
+++ b/code/controllers/subsystem/idlenpcpool.dm
@@ -2,7 +2,7 @@ SUBSYSTEM_DEF(idlenpcpool)
 	name = "Idling NPC Pool"
 	flags = SS_POST_FIRE_TIMING|SS_BACKGROUND|SS_NO_INIT
 	priority = FIRE_PRIORITY_IDLE_NPC
-	wait = 60
+	wait = 25
 	runlevels = RUNLEVEL_GAME | RUNLEVEL_POSTGAME
 
 	var/list/currentrun = list()
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -1,3 +1,5 @@
+#define MAX_DAMAGE_SUFFERED 250
+
 /mob/living/simple_animal/hostile
 	faction = list("hostile")
 	stop_automated_movement_when_pulled = 0
@@ -51,6 +53,8 @@
 	var/attack_all_objects = FALSE //if true, equivalent to having a wanted_objects list containing ALL objects.
 	var/lose_patience_timer_id //id for a timer to call LoseTarget(), used to stop mobs fixating on a target they can't reach
 	var/lose_patience_timeout = 300 //30 seconds by default, so there's no major changes to AI behaviour, beyond actually bailing if stuck forever
+	// Experimental Target Memory. Short term HATE. Things added to this list will have their accossiated values considered.
+	var/list/target_memory = list()
 
 	///When a target is found, will the mob attempt to charge at it's target?
 	var/charger = FALSE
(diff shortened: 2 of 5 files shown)


code/controllers/subsystem/idlenpcpool.dm
```diff
@@ -39,8 +39,6 @@ SUBSYSTEM_DEF(idlenpcpool)
 			continue
 
 		if(!SA.ckey)
-			if(SA.stat != DEAD)
-				SA.handle_automated_movement()
```
InsightfulParasite, line 43: I hate wander and apparently when a AI is idle it will still wander... just in the middle of nowhere. Im most likely going to have to put this back. At the top the idleNPC pool has a 6 second delay so i wonder if that is why some AI are blindsided when a player runs past them.

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -235,31 +243,26 @@
...
+	niche conditions. In some cases using Found() will suffice.*/
+/mob/living/simple_animal/hostile/CanAttack(atom/the_target)
+	if(isturf(the_target) || !the_target || the_target.type == /atom/movable/lighting_object)
+		// bail out on invalids
 		return FALSE
 
 	if(ismob(the_target)) //Target is in godmode, ignore it.
 		var/mob/M = the_target
+		//This is in order to make mobs not instantly reaggro on mobs they lost patience on.
+		if(mob_ignore_list[M] > world.time)
```
InsightfulParasite, line 264: New list that empties itself if its more than 2. In testing this usually results in ranged monsters pausing for a moment before resuming fire.

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -691,6 +818,46 @@
 		friends += fren
 		faction = fren.faction.Copy()
 
+/mob/living/simple_animal/hostile/proc/summon_backup(distance, exact_faction_match)
+	do_alert_animation()
+	playsound(loc, 'sound/machines/chime.ogg', 50, TRUE, -1)
+	for(var/mob/living/simple_animal/hostile/M in oview(distance, targets_from))
+		if(faction_check_mob(M, TRUE))
+			if(M.AIStatus == AI_OFF)
+				return
```
LanceSmites328, line 827:
```suggestion
				continue
```
Old code issue, might as well fix it while you're here (no clue why I didn't... maybe I did in Pink midnight+.)

LanceSmites328 (changes requested):

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -339,14 +421,113 @@
...
+			return TRUE
+
+// Adds entity to THE LIST OF GRUDGES which is reset upon gaining a new target.
+/mob/living/simple_animal/hostile/proc/RegisterAggroValue(atom/remembered_target, value, damage_type)
+	if(!remembered_target)
+		remembered_target = "nobody"
+	if(!target_memory[remembered_target])
+		target_memory += remembered_target
+	if(damage_type)
+		value *= damage_coeff[damage_type]
```
LanceSmites328, line 467: Should be updated to use GetCoeff() instead of []

LanceSmites328 (changes requested):
Overall please rebase this to make sure all conflicts/unupdated code is resolved. That's it for me

LanceSmites328 (approved):

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -339,14 +421,113 @@
...
+|Other Far = -50
+\-------------------*/
+/mob/living/simple_animal/hostile/proc/ValueTarget(atom/target_thing)
+	if(!target_thing)
+		return
+	//This is a safety net just in the case that no value is returned.
+	. = 0
+
+	//If your farther than 5 tiles from us you suffer the max 50 of hate penalty.
+	. -= clamp(get_dist(src,target_thing) * 10,0,50)
```
LanceSmites328, line 485:
```suggestion
	. -= clamp(get_dist(targets_from, target_thing) * 10,0,50)
```

LanceSmites328 (changes requested):
Realized this is the last thing I think needs changing, as it keeps it in line with old code. (Targets_From is whatever this thing actually takes targets from, so a bee hive and it's bees, the bees care about what's near the hive not themselves.)

InsightfulParasite:
> Realized this is the last thing I think needs changing, as it keeps it in line with old code. (Targets_From is whatever this thing actually takes targets from, so a bee hive and it's bees, the bees care about what's near the hive not themselves.)

That actually makes alot more sense.

LanceSmites328 (approved):

PR #1555, opened 2023-10-19
a bunch of weapons double-check if the user can use EGO, so lets fix it

Gboster-0:
## About The Pull Request

replaces a bunch of useless checks, this PR only affects the most blatant ones that are 100% safe to remove.

## Why It's Good For The Game

it does not matter much, makes the code more readable

## Changelog
:cl:
code: no more double-checks
/:cl:


EgorDinamit (approved):
All things considered it should work just fine.

PR #1564, opened 2023-10-21
ReSummons Hammer

LanceSmites328:
## About The Pull Request
Makes it so that the hammer checks after 5 minutes and 5 seconds if it exists anymore, and if it doesn't it re-materializes.

## Why It's Good For The Game
Prevents Well from hard-locking hammer.

## Changelog
:cl:
fix: fiixed hammer being lost forever
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/zayin/hammer_light.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/zayin/hammer_light.dm
@@ -150,12 +150,14 @@
 	user.hairstyle = "Bald"
 	user.update_hair()
 	update_icon()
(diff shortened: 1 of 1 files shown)


Coxswain-Navigator:
This does look good, although I am worried at the possibility of the duration going past 5 minutes and 5 seconds (i.e admins messing around or the stat cap being raised to an amount over 200). It may be worth considering moving the check of existence to the Check() proc which fires every time a mob dies or something breaches. Of course, this would probably hurt performance ever so slightly.

Coxswain-Navigator (approved):
Looks good.

PR #1567, opened 2023-10-21
Agent work statistics

EgorDinamit:
## About The Pull Request

- Adds agent work statistics for both, each individual abnormality and globally.
- Adds round-end report on agent records in terms of works(highest amount of works, highest attributes gained from works, highest PE generated).

## Why It's Good For The Game

All in all - just nice little things to look at. Now you can finally tell who is that sick fuck that worked on one sin 189 times.


Kitsunemitsu (approved):
Approved. Looks good to me, was tested on the server.

PR #1571, opened 2023-10-23
Fixes Armor Update hitby Proc Runtimes

LanceSmites328:
## About The Pull Request
Thrown Objects were throwing runtimes when they hit things because the default damage type for them is... Brute. Brute was also not being handled properly. This should resolve both issues.

## Why It's Good For The Game
Fixes runtimes

## Changelog
:cl:
fix: thrown movables creating runtimes
/:cl:


Kitsunemitsu (approved):
Approved.

PR #1586, opened 2023-10-30
Damage Coeff order of operations

LanceSmites328:
## About The Pull Request
Makes the unmodified before the modified one so it's not replaced before it can be copied.

## Why It's Good For The Game
Should fix an oversight.

## Changelog
:cl:
fix: unmodified not being right.
/:cl:


Kitsunemitsu (approved):
Approved. Minor fix.

PR #1588, opened 2023-10-31
Deep Scanner Upgrade

LanceSmites328:
## About The Pull Request
Upgrades the Deep Scanner to be usable while moving and from a range, but requires Line of Sight (uses can_see proc and, as such, is interrupted by some corners)
Causes non-god-moded aka "active" simple_animals to take 10% more damage from all sources upon being scanned. This bonus does not stack but lasts until the mob is killed.

Cleans up a bit of do_after code, partially porting https://github.com/tgstation/tgstation/pull/73117 

## Why It's Good For The Game
Gives extra use to the deep scanner which will give it a purpose that the manager doesn't fill.

## Changelog
:cl:
balance: buffs the deep scanner
code: tweaks a bit of do_after code
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/gadgets/unpowered.dm
+++ b/ModularTegustation/tegu_items/gadgets/unpowered.dm
@@ -128,7 +128,8 @@
 //Deepscanner
 /obj/item/deepscanner //intended for ordeals
 	name = "deep scan kit"
-	desc = "A collection of tools used for scanning the physical form of an entity."
+	desc = "A collection of tools used for scanning the physical form of an entity.\n\
+			Scanning an active hostile entity will make it 10% weaker to all damage."
 	icon = 'icons/obj/storage.dmi'
 	icon_state = "maint_kit"
 	slot_flags = ITEM_SLOT_BELT | ITEM_SLOT_POCKETS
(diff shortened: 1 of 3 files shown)


ModularTegustation/tegu_items/gadgets/unpowered.dm
```diff
@@ -143,21 +144,32 @@
...
+	if(!istype(target))
+		return
+	Scan(target, user)
+
+/obj/item/deepscanner/proc/Scan(mob/living/target, mob/user)
+	if(!isanimal(target) && !ishuman(target))
 		return
-	if(ishuman(M))
-		var/mob/living/carbon/human/H = M
+	user.visible_message("<span class='notice'>[user] takes a tool out of [src] and begins scanning [target].</span>", "<span class='notice'>You set down the deep scanner and begin scanning [target].</span>")
```
Gboster-0, line 161: maybe remove the "You set down the deep scanner and begin scanning [src]" instead replacing it with just "you start scanning [src]"

Kitsunemitsu (approved):
Cleans up and buffs an old gadget. Approved.

PR #1589, opened 2023-10-31
Warning level buffs and debuffs

LanceSmites328:
## About The Pull Request
Warning Levels now apply a +0/5/10/20 bonus to Justice and a -0/10/20/40 bonus to Temperance. This comes with an alert and applies on Warning Level change as well as applying on individually on join.

## Why It's Good For The Game
Encourages using Warning Level for alert and nudging employees towards combat, as opposed to keeping it on for alert lights. Different levels serve as a different kick in the ass for employees by the Manager.

## Changelog
:cl:
add: Warning Level buffs/debuffs
/:cl:


Kitsunemitsu (dismissed):
Yeah sure, I like this change.

EgorDinamit (commented):
Is there any reason to put it as a signal in lobotomy_events subsystem? As far as I can tell it doesn't interact with it in any way shape or form, might as well put this code directly in job code; Also, shouldn't you be checking for agents(those who can work)? I don't see any reason to give buffs to clerks and ERTs.

EgorDinamit (approved):

PR #1590, opened 2023-10-31
Vitals Projector and Enkephalin Injector Buff

LanceSmites328:
## About The Pull Request
Vital Projects no longer needs a WRENCH to change it's mode (??????????) Vital Projector no longer sucks most of the battery out every use (why was it more or less 20% + 15% to 25% of the battery????) It can also be changed while moving.

Enkephalin Injector now uses GiveTarget instead of just... changing the target. This is the proper way to do this.
Enkephalin Injector now gives you the gift of the abnormality you use it on. Yes, it's still locked to Clerks.
Enkephalin Injector no longer works on mobs in Godmode.

## Why It's Good For The Game
Makes these gadgets more useful.

## Changelog
:cl:
tweak: Removed Wrench requirement for Vital Projector
balance: reduced battery usage of Vital Projector
balance: Enkephalin Injector now gives the user the Abno's gift.
/:cl:


Kitsunemitsu (approved):
Yep, looks good to me.

PR #1606, opened 2023-11-04
RBA buffs

JackFrost2424:
## About The Pull Request

The current version of RBA is way too easy to keep contained and even if it does breach, it's a bit of a non-issue. I'm staying a bit on the safer side since it is still a TETH but ideally, this makes it a bit more of a threat
-Bumped up its HP and its red resistance a tiny bit. It's meant to be more fragile than other abnos due to having a ranged attack but it feels too squishy right now.
-Lowered the Counter from 3 to 2. 3 is way too generous for an abno whos counter can be raised easily.
-Bumped its ranged damage. It's a slow abno with a ranged attack that has a lot of spread. It should hurt if you just take the full brunt of it.
-Edited the description. Just to fit better with the sprite.

## Why It's Good For The Game

In its current state, RBA feels like a pushover both when it comes to keeping it contained and while in a breach. This is to make it slightly more threatening. 

## Changelog
:cl:
balance: rebalances RBA.
/:cl:


Kitsunemitsu (approved):
Sure. I'll approve this

PR #1611, opened 2023-11-07
Increases WCCA's attachment and repression work rates

Anonmare:
## About The Pull Request
Increases WCCA's attachment and repression work rates to be in line with its instinct and insight work rates.

## Why It's Good For The Game
It's struggling to compete with other ZAYINs for relevancy and it dealing constant damage on positive and negative ticks makes working it painful even if you perform perfectly, at least this way you'll have an excellent trainer. 
I have a bone to pick with Hammer but I'd prefer to discuss the matter before I consider reducing its work rates.

## Changelog
:cl:
balance: WCCA's work rates have been adjusted to be High on All works.
/:cl:


Kitsunemitsu (approved):
Sure fuck it. You do take consistent damage during it

PR #1612, opened 2023-11-08
[TM ONLY] Removes Realization Engine

Kitsunemitsu:
## About The Pull Request
Removes Realization engine temporarily

## Why It's Good For The Game
I don't like the impact this has had a good impact on the meta of endgame, and would like to experiment on gameplay without it

## Changelog
:cl:
del: Removed Realizer
/:cl:


EgorDinamit (changes requested):
1. This is not how you should temporarily remove something;
2. There's no way anyone is approving removal of a somewhat important end-game mechanic as is. The only way this gets through is if you rework how it works if it has such a "bad" effect on the game as you think it does.

Kitsunemitsu:
> 1. This is not how you should temporarily remove something;
> 
>     2. There's no way anyone is approving removal of a somewhat important end-game mechanic as is. The only way this gets through is if you rework how it works if it has such a "bad" effect on the game as you think it does.

image

Kitsunemitsu:
Testing over, Closing PR.

PR #1614, opened 2023-11-09
[TM Only] Temporarily removes antags from LOR13.

Kitsunemitsu:
## About The Pull Request
Removes Antagonists temporarily from LOR13.

## Why It's Good For The Game
Mel wanted to check what the map would be like without PVP. Let's see what goes on?

## Changelog
:cl:
del: Removed Antags from LOR13.
/:cl:


Kitsunemitsu:
LOR is getting a full rework.

PR #1624, opened 2023-11-11
Reversion to LC13

Kitsunemitsu:
## About The Pull Request
Removes all the code for:
City Map
R-Corp Assault
W-Corp Cleanup
Facility E
Fishing Role on F-Corp
Representatives and Sephirah

## Why It's Good For The Game
About time we focus on something more important around here.

## Changelog
:cl:
del: Removed everything but LC13
/:cl:


EgorDinamit:
Just because I *personally* dislike side modes doesn't mean I want them gone.
People like the idea, so improve on it. There **are** issues here and there, but it is inevitable with a project of this scale; The only thing it means is that we should be working on it, not removing it.

PR #1625, opened 2023-11-11
Reworks manager shields, fixes welf reward & changes how info core works

EgorDinamit:
## About The Pull Request

- Adds a return value to apply_damage() proc, which performs an early return (blocks damage).
- Manager shields and welfare reward now use said return value.
- Welfare Core Suppression nerfed. Instead of increasing damage modifiers by 1 each ordeal, it does so by 0.5
- Information Core Suppression changed. Instead of screwing with normal telecomms - it ruins all announcements. With each new ordeal - severity increases.

## Why It's Good For The Game

- A proper way to do it, I believe.
- Code improvement. Shouldn't affect players much.
- Less difficult, eh...
- More fun..! You can talk to people again.


Kitsunemitsu (approved):
Mostly a refactor, with a couple important buffs. Code looks great.

PR #1626, opened 2023-11-11
EO Text Adventure

InsightfulParasite:
## About The Pull Request
Adds a text adventure minigame to the EO's area.

TEXT BASED ADVENTURES!!! Adventures that are mostly predefined paths.	This was difficult to finalize since i have not made a text based adventure before.
- Currently is its own isolated datum with event datums that have been simplified in order to encourage the creation of more events by people.
- Pseudo battle system with block and attack.
- Adventure system where you can travel down paths in order to gain event progress.
- Events that have stat challenges.

I did specifically try to make it as simple as possible with the least amount of variables. This will prove to possibly be a mistake since there is no leveling or scaling system.

## Why It's Good For The Game
I was assigned to make a minigame for the EO since their job can get some lengths of time with no agent traffic.

## Changelog
:cl:
add: text_adventure_console.
/:cl:

### DIFF
--- a/ModularTegustation/lc13_adventuredatum.dm
+++ b/ModularTegustation/lc13_adventuredatum.dm
@@ -1,0 +1,1038 @@
+/*TEXT BASED ADVENTURES!!! Adventures that are mostly predefined paths.
+	This was difficult to finalize since i havent made a text based
+	adventure before. */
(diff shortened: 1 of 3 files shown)


Kitsunemitsu (approved):
Looks good to me, just something small to do

PR #1644, opened 2023-11-18
Facility upgrades

EgorDinamit:
## About The Pull Request

Adds facility upgrades to the auxiliary console, allowing manager to spend LOB points on various things.

- Manager bullets need to be unlocked/purchased before being available in the manager camera console;
- Manager bullets effect can be improved via upgrades (amount of HP/SP healed, health of shields);
- Maximum bullet ammo count is also an upgrade now;
- Agent spawning attributes can be augmented via an upgrade, for a maximum of +20.
- Agent spawning attributes now depend on current abno count, similar to how abnormality queue decides the risk level of available abnormalities. This should help with high pop rounds.
- Amount of abnormalities available for picking is initially 2, being able to be increased up to 5 with upgrades.

Has a ton of random content/code refactors along the way:
- Added persistent list that stores all core suppressions each player has cleared in their lifetime;
- Added all Keter core suppressions; All aside from "Proving Oneself" (Day 46) have requirements in form of having 50% of living agents with certain cores cleared before, utilizing the above persistent list;
- Agent spawning stats depend on amount of abnormalities in the facility instead of time;
- Random amount of code improvements.

At the round start the manager has 2 LOB points at their disposal. By working abnormalities up to full understanding, you can earn up to 10 LOB points.

## Why It's Good For The Game

Adds more flexibility to manager (or someone acting as one), allowing them to choose a different approach to the round.
Bullets at the round start were rather strong, so that was moved to an upgrade you first need to purchase.
As for Keter cores - this will most likely be the ultimate challenge, the absolute finale.


Kitsunemitsu (dismissed):
I'm going to have to ask for a removal of upgrading Abno queue. 
We do not want to give players more choices on this matter.

Instead, in a future PR, properly weight it with a database.

carshalash:
It kind of sucks to be totally kneecapped as a manager based on whether agents **feel like** getting abnormalities to 100% or not.

EgorDinamit:
> I'm going to have to ask for a removal of upgrading Abno queue. We do not want to give players more choices on this matter.

You will have to elaborate on why "we" don't want to give players more options into diversifying their round.

EgorDinamit:
> It kind of sucks to be totally kneecapped as a manager based on whether agents **feel like** getting abnormalities to 100% or not.

It does. I'd like to hear ideas on how to change/improve it; One thing should remain however, and that is earning LOB points for something out of the ordinary (like getting 100% understanding is).

EgorDinamit:
It's more or less ready for a merge, all future changes will be just new upgrades and tweaks, which can be done in new PRs.

Kitsunemitsu (changes requested):
I'm STILL issuing a reversal on Abnormality queue increases.
This could have multiple issues with the game.
1) Giving players more options means that they are more likely to pick the meta options every time. **This has been proven** by our other abnormality queue increase PR #1434
2) 3 is the sweet spot for abnormality choices. If you started with 2 and could only upgrade to 4, I would approve this PR I guess.
3) Giving the option if it's SUBOPTIMAL means that players won't pick it, and it won't be used. Giving the option if it's OPTIMAL means that less interesting abnormalities will be picked, and only the usual suspects will. 

Cox asked for my opinion on this PR again, and I am issuing a Do Not Merge until this is fixed.

EgorDinamit:
> 1. Giving players more options means that they are more likely to pick the meta options every time. **This has been proven** by our other abnormality queue increase PR [More Manager Choices #1434](https://github.com/vlggms/lobotomy-corp13/pull/1434)

I do not care what some other PR did, because this is not my PR. You should consider playing the game more often to see how exactly it affects it. This point has no value behind it and is a simple "I don't like it".

> 2. 3 is the sweet spot for abnormality choices. If you started with 2 and could only upgrade to 4, I would approve this PR I guess.

Don't care. Upgrading beyond 3 costs extra LOB points and is not much more than a meme option. I am not removing "meme options" either, players must have freedom to do what they want. Even dumb decisions.

> 3. Giving the option if it's SUBOPTIMAL means that players won't pick it, and it won't be used. Giving the option if it's OPTIMAL means that less interesting abnormalities will be picked, and only the usual suspects will.

Okay? Good. See above. The fact that people want to purchase additional queue slots, despite being a "shoot yourself in the foot" option (since you sacrifice actual upgrades for it), means that we have a bit too many of the bad abnormalities, which is something that should be fixed on its own. There's also the fact that managers simply want to have more options on how to handle the round, this includes how many abnormalities they can select from.

> Cox asked for my opinion on this PR again, and I am issuing a Do Not Merge until this is fixed.

There's nothing to fix.

LanceSmites328:
Pulling this out of the discord issue, if the issue is that our current choice amount works and works WELL, why don't we simply switch it for the ability to re-extract once? Cost still goes up per purchase but it allows the manager to bail out of 1 VERY BAD rotation (or even just one underwhelming) per purchase. You trade consistent buffs like Agent stats and bullet potency for MAYBE 1 better abno.

EgorDinamit:
PR updated.
The maximum amount of abno queue options will be 4 until other issues are resolved (i.e. amount of "off-meta" abnos, the queue system itself).

Coxswain-Navigator (approved):
Experimental Changes: 2 is waiting.

PR #1662, opened 2023-11-27
Adds Library area types and replaces library rooms with corresponding types

LancashireladOnGitHub:
Adds the new area types and replaces each corresponding floor with their corresponding area

## About The Pull Request

Created new area types, gave them gravity, power, etc, applied them to the corresponding floors

## Why It's Good For The Game

At the moment spawning the library is a step by step process, which is already an ordeal in of itself, then theres the ordeal of "finding" the library, which before this PR has "backstreets room" as its default room type, this causes, many, many issues and greatly limits an admins ability to do some more wacky and creative events

Now, we can just teleport to the room which is uniquely labelled for each individual floor, saving time and frustration.


LancashireladOnGitHub:
It appears that I have done as they say a "fuckey wuckey"

TaculoTaculo:
No admins were harmed in the making of this PR

Kitsunemitsu (approved):
Yeah sure, approved.

PR #1666, opened 2023-11-29
Records core suppression

EgorDinamit:
## About The Pull Request

At long last - adds the Records core suppression.

Challenge:
- Each meltdown, random abnormalities (up to 20%, 40%, 60%) swap places. Could probably leave a temporary effect when it happens.
- Every 30, 18, 6 seconds a random amount of all living, uncontained and not working mobs get teleported a short distance (not inside the walls or space). An after-image (similar to AAH realized ability, but without rotation) is shown to indicate where they moved.
Reward:
- Attributes limit is raised by 70 for **all jobs**.
- Manager is able to swap abnos at will via manager console.

## Why It's Good For The Game

The last "normal" core suppression with unique challenge and reward.
Having more options is always good anyway, but this one is much more fun :)


code/modules/mob/living/simple_animal/abnormality/teth/punishing_bird.dm
```diff
@@ -82,6 +82,14 @@
 	UnregisterSignal(SSdcs, COMSIG_GLOB_HUMAN_INSANE)
 	return ..()
 
+// Deletes perch if we moved while not breaching; Generally caused by cell swaps
+/mob/living/simple_animal/hostile/abnormality/punishing_bird/Move()
+	if(IsContained())
+		var/obj/structure/pbird_perch/P = locate() in get_turf(src)
+		QDEL_NULL(P)
+	. = ..()
+	PostSpawn()
```
LanceSmites328, line 91: This would spawn a new perch in every spot it moves due to calling PostSpawn(), which has been overridden to spawn perch.

code/modules/suppressions/records.dm
```diff
@@ -0,0 +1,149 @@
...
+	UnregisterSignal(SSdcs, COMSIG_GLOB_MELTDOWN_START)
+	UnregisterSignal(SSdcs, COMSIG_GLOB_ABNORMALITY_SWAP)
+	// EVERYONE GETS INCREASED STAT LIMIT, LET'S FUCKING GOOOOOO
+	for(var/datum/job/J in SSjob.occupations)
+		J.job_attribute_limit += 70
+	for(var/mob/living/carbon/human/H in GLOB.human_list)
+		if(!H.ckey)
+			continue
+		H.adjust_attribute_limit(70)
+		to_chat(H, span_notice("You feel more powerful than ever before!"))
```
LanceSmites328, line 40:
```suggestion
		to_chat(H, span_notice("You feel new potential unlock within you!"))
```
Previous one implies immediate power growth, which is not what happens with the attribute limit being raised.

LanceSmites328 (changes requested):
While I do not believe the rewards for this suppression are amazing, mainly being useless for about a third of players yet still being given to them, that is irrelevant to this thing's functionality which is honestly my main concern.

code/modules/suppressions/records.dm
EgorDinamit: Sure, that works.

code/modules/mob/living/simple_animal/abnormality/teth/punishing_bird.dm
EgorDinamit: That's the point - it deletes perch before moving and then spawns another on a new place.

EgorDinamit: Saw the issue. Got it fixed.

LanceSmites328 (approved):

PR #1667, opened 2023-11-30
CENSORED tweaks

EgorDinamit:
## About The Pull Request

A very tiny update to CENSORED:
- CENSORED now has a ranged ability, pretty much the same as its EGO weapon;
- CENSORED applies a status effect to all humans around it, which constantly drops sanity of its target until it reaches 30%;
- CENSORED and its minions have higher damage and movement speed.

## Why It's Good For The Game

- Allows CENSORED to be a *tiny bit* more dynamic in a fight. It is still not ideal, but it is something;
- "Nah, I'll just wait until my sanity is back to full and then fight CENSORED", said a foolish clerk. Makes it less of an option to heal sanity mid-fight.
- They were really weak.


LanceSmites328:
I don't think CENSORED itself needs the raw damage buffs, especially with it gaining an ability and pure-sanity damage aura. It already swung for more damage than Nothing There with a deadlier damage type. With the DoT, this makes you two-hit from insanity with Black 5, barely becoming 3 hits away with Black 8. Just kinda shuts melee out of the equation entirely.

LanceSmites328 (approved):

PR #1672, opened 2023-12-01
Eris Hotfixes

Kitsunemitsu:
## About The Pull Request
Eris now has been fixed to only attack dying agents (and instantly killing them)
Eris now scales her damage to the same as the healing.

## Why It's Good For The Game
Eris was always intended to remain neutral, and only kill dying agents.
Her counter damage would get absurd extremely quickly. I want to curb this by making it scale roughly at damage healed x2.
She'll still be annoying to take down but it's not that difficult if a team of players wants to.

## Changelog
:cl:
balance: Stopped Eris from dealing scaling damage on her counter.
fix: fixed Eris attacking all agents.
/:cl:


Coxswain-Navigator (approved):
It's an improvement, although I kind of wish all of this girlboss math wasn't so needlessly confusing.

PR #1673, opened 2023-12-01
Updates some abnormalities

Coxswain-Navigator:
## About The Pull Request
Adjusts a couple abnormalities, may add more soon.

Makes pale horse slightly more dangerous.

- Status effect now deals a small amount of pale damage over time.
- Increased the range of most attacks.

Fixes an issue with shelter from the 27th of march
- Players can no longer drag objects in and out of the shelter, only players.

Book without pictures or dialogue has been mostly reworked

- Performing repression work early now lowers qlipthoth in addition to previous effects.
- A breach effect has been added, spawning between 3 and 5 "paper" mobs. They have 1/10th of the usual health.
- Performing repression work when the book is full as intended spawns a friendly animal in paper form.

## Why It's Good For The Game
Stuff is less bad.

## Changelog
:cl:
tweak: Pale horse now has a larger area of attack
tweak: Book without pictures or dialogue has been entirely reworked
fix: crates can no longer be dragged out of the shelter
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/teth/pale_horse.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/teth/pale_horse.dm
@@ -40,7 +40,7 @@
 	var/teleport_cooldown_time = 30 SECONDS
 	//attack
 	var/mob/living/set_target
-	var/pulse_range = 7 //same range as scorched girl
+	var/pulse_range = 11 //fairly large area - enough to breach several abnormalities
 	var/fog_damage = 3
 	var/ash_damage = 20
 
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/simple_animal/abnormality/teth/pale_horse.dm
```diff
@@ -114,12 +114,14 @@
 
 
 /mob/living/simple_animal/hostile/abnormality/pale_horse/Move() //more damaging fog when moving
```
EgorDinamit, line 116: Use Moved() proc. See big bird for t.

LanceSmites328 (approved):

PR #1675, opened 2023-12-01
Weighted Abnormality Choices

LanceSmites328:
## About The Pull Request
Makes all abnormalities weighted by pickweight, primarily making more frequently picked abnormalities appearing less often. 
Abnormalities in relevant groups (Yin and Yang) gain a varying weight increase whenever one member of this group is chosen to make others of that group appear more often.
Some groups have been included cosmetically, most (if not all) of these have a 1.5x multiplier to the weight per member picked. Fairies make fairies more likely, etc. 

Also refactors Servant of Wrath's code to be wrath_servant as opposed to servant_wrath, so that it's in line with EVERY OTHER magical girl.

## Why It's Good For The Game
Encourages diversity without permanently excluding any ONE abnormality. No abnormality will ever have a 0% pick chance. Also allows for easier acquisition of sets.

## Changelog
:cl:
add: List weighting to abnormality queue
add: Persistence to abnormality picking.
add: abnormality group modifiers to many abnormalities
refactor: initial "possible abnormality" list filling has been moved to Persistence subsystem
server: changed initialization order for abnormality queue to always be behind Persistence
/:cl:

### DIFF
--- a/code/controllers/subsystem/abnormality_queue.dm
+++ b/code/controllers/subsystem/abnormality_queue.dm
@@ -4,6 +4,7 @@ SUBSYSTEM_DEF(abnormality_queue)
 	name = "Abnormality Queue"
 	flags = SS_KEEP_TIMING | SS_BACKGROUND
 	runlevels = RUNLEVEL_GAME
+	init_order = INIT_ORDER_PERSISTENCE - 1 // Always after Persistence, as that's where we get our abno list from.
 	wait = 10 SECONDS
 
 	/// List of(preferably) 3 abnormalities available for manager to choose from.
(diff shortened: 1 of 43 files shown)


code/controllers/subsystem/abnormality_queue.dm
```diff
@@ -29,13 +30,10 @@ SUBSYSTEM_DEF(abnormality_queue)
...
 /datum/controller/subsystem/abnormality_queue/Initialize(timeofday)
-	var/list/all_abnos = subtypesof(/mob/living/simple_animal/hostile/abnormality)
-	for(var/i in all_abnos)
-		var/mob/living/simple_animal/hostile/abnormality/abno = i
-		if(initial(abno.can_spawn))
-			possible_abnormalities[initial(abno.threat_level)] += abno
 	if(LAZYLEN(possible_abnormalities))
 		pick_abno()
+	else
+		stack_trace("[src] initialized before Persistence subsystem!")
```
EgorDinamit, line 36: As per code standards.
```suggestion
	if(!LAZYLEN(possible_abnormalities))
		stack_trace("[src] initialized before Persistence subsystem!")
		return
	pick_abno()
```

LanceSmites328: Doing this prevents starting room count from being set, next abno spawn time from being decremented, and parent call from being sent. This is intentionally set this way as if Persistence loads later, the initial spawn will still work it'll just not give the manager that first choice.

EgorDinamit (approved):

code/controllers/subsystem/abnormality_queue.dm
EgorDinamit: If that's the case, it should be fine.

PR #1693, opened 2023-12-04
Updates W-Corp L2A Equipment

Moribox:
## About The Pull Request
Changes Wcorp L2A's to have a different set of comms to better command and control squads alongside the L3 Commanders

## Why It's Good For The Game
Allows L2A's to better fill their role as a recon specialist and assistant to the L3 Commanders, rather than being treated as just another L2 Wcorper.

## Changelog
:cl:
Changes their headset to match that of the commanders to allow en-mass communication.
:cl:


Kitsunemitsu (approved):
Approved. Code looks good, and this was always intended

PR #1698, opened 2023-12-05
Adds R-corp map "Nest" 

LancashireladOnGitHub:
Adds the "Nest" a PvP map with a focus on long range engagements to start with, then entering CQC as the abnos get pushed further and further back, or the abnos just dont push, its up to them really.

Weight is set to 5 using the standard of maps like "Maze" and "Standard"


Kitsunemitsu (changes requested):
We don't use the button anymore, look into other R-Corp maps so that the VIP can spawn

Kitsunemitsu (approved):
Looks fine to me.

PR #1710, opened 2023-12-07
Wcorp carriage pack

LancashireladOnGitHub:
Adds the "Zoo" and "Stage" variants for the Wcorp revamp

Short sweet lil maps that I made while on alot of medication, quality may vary


Kitsunemitsu (approved):
Minor mapping PR.

PR #1720, opened 2023-12-08
Adds an option to generate typecaches as zebras, a TG mirror

Gboster-0:
## About The Pull Request

TG creator's comments:
Adds an argument to typecache generation that allows specifying the whether to include/exclude types in the input list.
Also adds another argument to specify whether to remove falsey values after the typecache is generated.

I also added in some small changes to code TG had and we didnt, like /proc/is_type_in_list having single-letter variables. Only those that conflicted were admitted

## Why It's Good For The Game

TG creator's comments:
Might make zaps slightly faster???
Honestly I just thought it would be a good way to condense some whitelist/blacklist typecache sets.

## Changelog
:cl:
code: a lot of lists now dont look like shit
/:cl:

### DIFF
--- a/code/modules/events/anomaly.dm
+++ b/code/modules/events/anomaly.dm
@@ -24,8 +24,8 @@
 		/area/holodeck,
 		/area/shuttle,
 		/area/maintenance,
-		/area/science/test_area)
-		)
+		/area/science/test_area,
+	))
 
 		//Subtypes from the above that actually should explode.
 		var/list/unsafe_area_subtypes = typecacheof(list(/area/engine/break_room))
--- a/code/__HELPERS/_lists.dm
+++ b/code/__HELPERS/_lists.dm
@@ -90,16 +90,49 @@
 
 			return "[output][and_text][input[index]]"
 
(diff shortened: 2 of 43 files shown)


code/modules/events/anomaly.dm
```diff
@@ -24,8 +24,8 @@
 		/area/holodeck,
 		/area/shuttle,
 		/area/maintenance,
-		/area/science/test_area)
-		)
+		/area/science/test_area,
```
Coxswain-Navigator, line 27: What is with these new commas at the end of some lists?

Coxswain-Navigator (approved):
Good PR overall. The trailing commas are apparently a TG coding standard that allows for less deletions when adding to lists so it's fine by me.

code/modules/events/anomaly.dm
Gboster-0: was discussed in discord but if anyone in tha future is reading the PR:
This makes it so new additions to the lists count as only 1 addition instead of 2 additions and 1 deletion

PR #1738, opened 2023-12-16
Health calculation fixes and Red damage spread

Halonexus:
## About The Pull Request
Legs and arms contribute only 75% of their damage to total health damage and since all our damage is spread over all limbs it ends up increasing human effective health by 20%.
Pale damage is calculated based on 100% of human health but since all humans have 120% effective health it ends up being wrong.
health-before
For example here it requires 240 red damage or 120 pale damage to reduce 200 unarmored health to 0.

Additionally made red damage behave the same way black and pale do: it now always spreads damage over all limbs.

## Why It's Good For The Game
Removes obscure and hidden damage resistance.
Pale damage will actually deal correct amount of damage.
Black damage is no longer almost guaranteed to deplete sanity before critting a human.
Red and pale damage will be scarier due to fixed health.

Red damage spreading over all limbs improves consistency with other damage types.

## Changelog
:cl:
code: removed limb coefficients from health calculation reducing effective human health from what it used to be.
code: red damage is spread over all limbs.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -558,9 +558,9 @@
 	var/total_stamina = 0
 	for(var/X in bodyparts)	//hardcoded to streamline things a bit
 		var/obj/item/bodypart/BP = X
-		total_brute	+= (BP.brute_dam * BP.body_damage_coeff)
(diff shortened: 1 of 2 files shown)


LanceSmites328 (approved):

EgorDinamit (commented):
I wanted this fixed since forever... Thanks.

PR #1747, opened 2023-12-19
Particle Emitter Fadeout Proc

LanceSmites328:
## About The Pull Request
Adds a proc to turn off particle emissions and then delete the emitter after all particles have, in theory, been deleted. 

## Why It's Good For The Game
This allows for emitters to not just "vanish" all of their VFX the moment they need to be disposed of, allowing for lingering effects without the production of more. While this was previously easy to do this simplifies it further into a singular proc.

## Changelog
:cl:
add: fadeout() proc for particle_emitter objects
/:cl:

### DIFF
--- a/code/game/objects/effects/particles/particles.dm
+++ b/code/game/objects/effects/particles/particles.dm
@@ -119,6 +119,13 @@
 	else
 		particles.spawning = 0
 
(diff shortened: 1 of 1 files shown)


Coxswain-Navigator (approved):
Oh yes indeed

PR #1753, opened 2023-12-21
Adds abnormality portraits

Coxswain-Navigator:
## About The Pull Request
image

Adds abnormality portraits to work consoles. 

## Why It's Good For The Game
Work is really boring so we need to do everything we can to spice things up.

## Changelog
:cl:
add: Added abnormality portraits
/:cl:


EgorDinamit (changes requested):
The portraits are too big for the current UI of work console, and PR cannot be merged in such state.
A proper rework would be preferred, but the image size could be simply reduced with buttons moved around.

EgorDinamit (approved):

PR #1785, opened 2024-01-04
Deltacorp2.0

ScarSkit:
A rework of Deltacorp

## About The Pull Request

Old Map:

image

New Map:

image

## Why It's Good For The Game

higher concentration of abnormality cells.
reduced redundant departments to admin-only access.
rotated map from vertical to horizontal to better fit the gimmick of being a facsimile of the facility in PM's L Corp game.
decreased height of hallways, increased width of hallways due to bias of a player character's perspective (see more tiles west to east than north to south)
more consistent theming.
extraction office is more accessible.
elevators are implemented to reduce on distance traveled.
distance traveled are reduced in comparison to the old map.

## Changelog
:cl:
add: Added deltacorp.dmm
del: Removed deltacorp.dmm
tweak: deltacorp.json
/:cl:


ScarSkit:
NOTE: the three departments, record, extraction and architecture, are NOT available in normal play. Their function is just because they existed in the original map in the original game of lobotomy corp.

NOTE: they are only accessible via the Gamma corp method of admin intervention. They should not reflect on the overall balance of the map, its just an extension that admins can open if the players decide to have the FULL lobotomy hell experience

PR #1791, opened 2024-01-06
Skittish goes from 4 to 2 points

Ben10083:
## About The Pull Request

Skittish Trait goes from 4 to 2 points

## Why It's Good For The Game

Skittish trait is on par with other traits like night vision, having it be 4 points was too much and putting it to 2 should better reflect it's impact.

## Changelog
:cl:
balance: Skittish trait reduced from 4 to 2 trait points

/:cl:

PR #1792, opened 2024-01-07
Officer Upgrade Injector now 200 PE

Ben10083:
## About The Pull Request

Officer Upgrade Injector cost now 200 from 400.

## Why It's Good For The Game

Officers are in a odd state where they are both vital to the facility running smoothly, yet also immensely fragile without substantial investment. This PR aims to reduce the cost to invest into officer(s) and allow them to get out of 20 stat hell easier

## Changelog
:cl:
balance: Officer Upgrade Injector now 200 PE
/:cl:

### DIFF
--- a/code/datums/quirks/good.dm
+++ b/code/datums/quirks/good.dm
@@ -195,7 +195,7 @@
 /datum/quirk/skittish
 	name = "Skittish"
 	desc = "You're easy to startle, and hide frequently. Run into a closed locker to jump into it, as long as you have access. You can walk to avoid this."
-	value = 4
+	value = 2
 	mob_trait = TRAIT_SKITTISH
 	medical_record_text = "Patient demonstrates a high aversion to danger and has described hiding in containers out of fear."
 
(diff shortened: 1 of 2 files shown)


CrabbytheCrab:
finally I wont spend 10000 pe getting to ex after a records core

EgorDinamit:
> vital to the facility running smoothly

???

code/datums/quirks/good.dm
LanceSmites328, line 1: What's this doin here

LanceSmites328 (changes requested):

code/datums/quirks/good.dm
Ben10083: https://github.com/vlggms/lobotomy-corp13/pull/1791 Is that, unsure why its shown as changed, did it get changed again?

Kitsunemitsu:
400 isn't that expensive in the grand scheme of things. Goes double because the EO is in charge of officer injectors.
Along with this, it is still purchased at 400, and both the EO and RO have been enjoying a short renaissance

PR #1645, opened 2023-11-18
Hana update

Kitsunemitsu:
## About The Pull Request
Needed for a test today. Updates various aspects of Hana and City to be more polished.

## Changelog
:cl:
add: Tweaked City.
/:cl:

PR #1802, opened 2024-01-09
Clinic Blood

KingSuperBOB:
## About The Pull Request
Adds blood bags of multiple blood types to the clinic

## Why It's Good For The Game
The clinic having to draw their own blood with syringes throughout the round to store it for patients is not good gameplay

## Changelog
:cl:
add: Added a blood crate type which does not have shock blood or lizard blood, gave new crate type to clinic
/:cl:

### DIFF
--- a/_maps/map_files/Event/city.dmm
+++ b/_maps/map_files/Event/city.dmm
@@ -3863,6 +3863,10 @@
 	},
 /turf/open/floor/facility/white,
 /area/city/fixers)
+"JI" = (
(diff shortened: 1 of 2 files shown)


Kitsunemitsu:
Map currently being touched in #1645 

I will make sure to add bloodbags into that map, if I didn't already.
For the sake of easy of use, and this being a very minor PR, I will be closing it, but be not afraid. It will be fixed.

PR #1829, opened 2024-01-14
Fix a number of runtime errors

liz-lavenza:
## About The Pull Request
Fixes a runtime in smoothing code due to anvils having improperly set smoothing values. I used table racks as a reference for the new ones.
Fixes a runtime when an owner-less brain gets brain trauma.
Fixes many runtimes due to null client and null client variables.

## Why It's Good For The Game
Some of these were in important code, like the chatmessage runtimes. Hopefully this will fix some of the issues reported in the new player round.


Kitsunemitsu (approved):
Excellent first PR. Great work.

PR #1850, opened 2024-01-17
Butcher spawns with hook

KingSuperBOB:
## About The Pull Request
Removes the meathook which spawns on the bistro, makes butchers spawn with their meathooks instead

## Why It's Good For The Game
I don't think it's healthy for CoL to have rats and syndicates metarush the strongest stun weapon currently available to win any combat while ignoring level difference (unless you get solar flared for 10 seconds but that only applies to fixer combat)

## Changelog
:cl:
add: Butchers spawn with their meathooks on them
del: Removed bistro meathook
/:cl:

### DIFF
--- a/_maps/map_files/Event/city.dmm
+++ b/_maps/map_files/Event/city.dmm
@@ -1936,15 +1936,6 @@
 /obj/structure/bed,
 /turf/open/floor/carpet/donk,
 /area/city/house)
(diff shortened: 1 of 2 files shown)


Kitsunemitsu (approved):
Approved. Great change I've been meaning to do

PR #1857, opened 2024-01-18
N-Corp Representative now gets temporary attribute increasers

KingSuperBOB:
## About The Pull Request
As the title says N-Corp rep gets three variants of attribute increasers, one lasts for 5 minutes and buffs a specific stat by 15, the other lasts for 2 minutes but buffs all stats by 15 last one is the same as the first one but 20 stats instead of 15. (Im going to give them sprites eventually though theyll likely just look like the ampules but blue)

## Why It's Good For The Game
N-Corp rep should not become the most useless rep in the game after all the ampules are used, also Ahn should be used in every representative so it is not a useless currency.

## Changelog
:cl:
add: Added temporary attribute increasers that buff one stat by 15 for 5 minutes
add: Added temporary attribute increasers that buff all stats by 15 for 2 minutes
add: Added temporary attribute increasers that buff one stat by 20 for 5 minutes
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/representative/items/ncorp.dm
+++ b/ModularTegustation/tegu_items/representative/items/ncorp.dm
@@ -64,6 +64,47 @@
 	allowed_roles = list("Records Officer", "Extraction Officer")
 
 //Temporary attributes
+#define STATUS_EFFECT_FORTITUDE /datum/status_effect/ncorp/fortitude
+#define STATUS_EFFECT_PRUDENCE /datum/status_effect/ncorp/prudence
+#define STATUS_EFFECT_TEMPERANCE /datum/status_effect/ncorp/temperance
+#define STATUS_EFFECT_JUSTICE /datum/status_effect/ncorp/justice
+
+/atom/movable/screen/alert/status_effect/ncorp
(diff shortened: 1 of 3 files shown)


ModularTegustation/tegu_items/representative/items/ncorp.dm
```diff
@@ -63,3 +63,308 @@
 	amount = 80
 	allowed_roles = list("Records Officer", "Extraction Officer")
 
+//Temporary attributes
```
Kitsunemitsu, line 66: Look at A Quiet Day actually on how to cut lines by like half

Kitsunemitsu (changes requested):
Optimizations requested.

ModularTegustation/tegu_items/representative/items/ncorp.dm
KingSuperBOB: Took the code right out of Flesh Idols realised ability which may have been a bad idea, also fixes a bug with the K-Corp research text while at it and formatted N-Corp text a bit better.

Kitsunemitsu (approved):
Not perfect, but yes. We will take this.

PR #1860, opened 2024-01-18
P Corp Rep now actually get his gloves

BLUREMOTE:
## About The Pull Request

P Corp Rep now get the intended items when getting the glove researches instead of pouches.

## Why It's Good For The Game

Might be cool if you could actually get the items you are promised.


Kitsunemitsu (approved):
Approved.

PR #1869, opened 2024-01-19
BL swords are no longer cardboard!

KingSuperBOB:
## About The Pull Request
Every BL sword has been replaced with weapon/bladelineage instead of weapon/bladelineage/city which was weaker

## Why It's Good For The Game
Blade Lineage have no armor and even their strongest combatant does not do that much damage with claim their bones, this should actually make them work within the city.

## Changelog
:cl:
tweak: Blade Lineage blades swapped with the non weak variants
/:cl:


Kitsunemitsu (approved):
Approved.

PR #1871, opened 2024-01-19
adds fetus, philip and child of the galaxy to flesh idol blacklist

KingSuperBOB:
## About The Pull Request
These three were added each for their respective reasons
Galaxy is a instant unavoidable fuck you, arguments may be made about abnos like prince however their effects are minor and can be easily undone unlike death.
Philip was added due to the same reason as DF and Nihil, being a major boss.
Fetus may sound out of place in this blacklist however he is also a unavoidable middle finger as there is no way of stopping his crying without death, if he were to instantly kill the agent there would be no difference.

## Why It's Good For The Game
We should not have a abnormality instantly kill agents with no counterplay, the excuse of "Don't work Flesh Idol" also does not work as meltdown works count towards his breach counter this will hopefully make the game healthier.

## Changelog
:cl:
add: Adds Phillip, Child of the galaxy and nameless fetus to Flesh Idol blacklist
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/waw/flesh_idol.dm
@@ -36,7 +36,7 @@
 	var/list/blacklist = list(/mob/living/simple_animal/hostile/abnormality/melting_love, /mob/living/simple_animal/hostile/abnormality/distortedform,
 				/mob/living/simple_animal/hostile/abnormality/white_night, /mob/living/simple_animal/hostile/abnormality/nihil,
 				/mob/living/simple_animal/hostile/abnormality/galaxy_child, /mob/living/simple_animal/hostile/abnormality/fetus,
-				)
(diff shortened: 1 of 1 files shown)


CrabbytheCrab:
child does black damage based off of how many people has the pebbly

CrabbytheCrab:
There can also be situations where an agent that is near prince's cell gets selected by prince when flesh idol breaches it

KingSuperBOB:
> There can also be situations where an agent that is near prince's cell gets selected by prince when flesh idol breaches it

The difference between prince and the pebble is the pebble instantly kills you with no counterplay, prince insanes you and allows for resaning the only situation theres zero counterplay against prince is if someone is literally standing outside his containment.
The issue here is the lack of what you can do as a agent to stop the blacklisted abnos, which is why theyre blacklisted, unavoidable breaches with no counter.

Kitsunemitsu (approved):
Approved.

PR #1877, opened 2024-01-21
mech damage fixes

Halonexus:
## About The Pull Request
All simple animals deal "20 * environment_smash" damage to mechs and environment_smash is 1 for all abnormalities regardless of threat level. This pr makes combat mechs take actual simple mob damage instead.
Additionally added directional armor and justice scaling to weapon attacks from humans.

Since most abnoramality damage is more than 20 this change will require mechs to get more armor and/or integrity to have similar survivability as before.

## Why It's Good For The Game
All abnormalities no longer deal same damage to mechs regardless of their actual damage stats.
Support for mech vs human combat if that ever happens.

## Changelog
:cl:
tweak: Justice modifier and direction affect weapon damage dealt to mechs
fix: fixed mechs taking 20 damage regardless of abnormality damage stats
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -156,6 +156,18 @@
 	else
 		return ..()
 
+/obj/vehicle/sealed/mecha/attacked_by(obj/item/I, mob/living/user)
+	if(I.force)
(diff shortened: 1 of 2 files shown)


Kitsunemitsu:
This makes a lot of sense. Merging.

Kitsunemitsu (dismissed):

Kitsunemitsu (approved):
Re-approving.

PR #1882, opened 2024-01-22
Possible fix to dead followers following

InsightfulParasite:
## About The Pull Request
When a steel loses aggro it auto returns to its leader without checking if the steel is already dead. Now it checks the stat of the steel losing aggro.

## Changelog
:cl:
fix: dead steels following commanders.
/:cl:


Gboster-0:
they were too determined to follow their commander...

LanceSmites328 (approved):

PR #1890, opened 2024-01-23
changes animalism

CrabbytheCrab:
## About The Pull Request

Does some changes to animalism. Gibs on kill now, lowered the damage from 20 to 12 but now the hits after the first scale with justice,  checks if the target is next to you, and has hit text
## Why It's Good For The Game

Makes the secondary hits of animalism more like melee hits and the gibbing on kill makes it cooler.

## Changelog
:cl:
balance: rebalanced animalism's secondary hits to be more like melee hits and gibs on kill now
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/waw.dm
+++ b/code/game/objects/items/ego_weapons/waw.dm
@@ -1436,7 +1436,6 @@
 	for(var/i = 1 to 3)
 		sleep(2)
 		if(target in view(reach,user))
-			gibcheck(target)
 			target.send_item_attack_message(src, user,target)
 			target.apply_damage(force, damtype, null, target.run_armor_check(null, damtype), spread_damage = TRUE)
 			user.do_attack_animation(target)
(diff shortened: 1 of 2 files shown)


CrabbytheCrab:
is there a problem with gibbing?

LanceSmites328:
> is there a problem with gibbing?

Can be? I think that's something we need to set a precedent about with how we want it to work and if it should be treated as an 'execution' effect with some abnos or not

TravisAngeI:
> is there a problem with gibbing?

Isn't this weapon dropped from one of the distortions in the CoL gamemode?
If I'm right, then gibbing is probably a bad idea as that is effectively a guaranteed permakill

CrabbytheCrab:
Well no its apex predator's weapon

vampirebat74:
> Well no its apex predator's weapon

It is re-used by the bunnyman

Coxswain-Navigator (approved):
Should be an improvement.

PR #1907, opened 2024-01-26
Jeong Weapon Fix

LanceSmites328:
## About The Pull Request
Jeong weapons are supposed to sacrifice 20% hp for 3x damage. They currently sacrifice double this.

## Why It's Good For The Game
Makes weapon more accurate in the better of the two ways.

## Changelog
:cl:
fix: Jeong office weaponry now takes 20% hp instead of 40%
/:cl:


Coxswain-Navigator (approved):
Ez fix

PR #1921, opened 2024-01-29
Mermaid Qliphoth Counter lose due to her crown cannot happen while working her as the lover

BLUREMOTE:
## About The Pull Request
The Qliphoth Counter reduction from Mermaid's Crown doesn't happen if the lover is working her.
She can still lose a counter midwork if anyone else but the lover work her.

## Why It's Good For The Game
Kinda lame to lose a counter midwork.


Kitsunemitsu (approved):
Sure, This is a nice fix.

PR #1941, opened 2024-02-07
Update my_sweet_home.dm

StrangeUser:
## About The Pull Request
home spawned at 500 hp for some reasons, while it max health at 600 actually

## Why It's Good For The Game
minor bugfixing


Coxswain-Navigator (approved):
How did no one else notice this before?

PR #1943, opened 2024-02-08
Wishing well rework

Coxswain-Navigator:
## About The Pull Request
Overall buffs to the wishing well tool abnormality so that its usefulness is comparable to other systems that have been added since its introduction.

- Better rates overall with a significantly lower chance of getting nothing
- Faster operation
- Audiovisual improvements
- Overall code improvements

In addition, a rare chance to get an even better item than what was put in has been added. This was a feature that was originally rejected, but has since become a feature that would significantly improve it.

## Why It's Good For The Game
The system that actually costs ego to use is no longer worse than the system that is pretty much free.

## Changelog
:cl:
balance: improved wishing well
/:cl:


Gboster-0:
would be nice if you could make the lists to be up-to-date in terms of format, otherwise looks fine

EgorDinamit (approved):
Eh

PR #1955, opened 2024-02-10
Another Gbee fix for Pink Midnight

InsightfulParasite:
## About The Pull Request
General Bee is escaping containment but remains in godmode?
General Bee due to her having a normal breach and a TRUE breach means that her "can breach" var is set to FALSE.
if(breach_type != BREACH_NORMAL && !can_breach) returns false in the core abnormality code. So if pink midnight calls pink breach, general bee's core abnormality code returns false instead of ungodmoding her.

Additionally attempts to fix a runtime by adding a early return to general bees fire shell proc if it ends up not having any targets listed.

## Changelog
:cl:
fix General Bee
/:cl:


Coxswain-Navigator (approved):
Nice little code improvement and fixes a potentially game breaking bug

PR #1958, opened 2024-02-11
Tweaks Naked Nest Cure Sources

InsightfulParasite:
## About The Pull Request
Moves the Naked Nest cure from something you get every time you do a good work to a purchasable EGO. Will reduce the amount of naked nest cures just laying around.
image

## Why It's Good For The Game
Ive been told in the past by players that the cure is too common. One naked nest good work is 20 PE so making the cure 20 unique PE will make it essentially the same as getting it on good work. But a new issue arrives and thats requesting the cure from EO. Now, worst case senario this system will result in a EO denying cure requests because someone wants the naked nest EGO suit or weapon. If this senario is too real then i will try to think tank of some other method that isnt my original "each infection has a random bar reagent cure that is determined upon infection" idea. That idea can be used for abnormalities that have less lethal infections and are zayins.

I tested making a new item datum and it seems to work fine. No runtimes last time i tested it.

## Changelog
:cl:
tweak: naked nest cure source.
/:cl:

### DIFF
--- a/code/datums/abnormality/_ego_datum/waw.dm
+++ b/code/datums/abnormality/_ego_datum/waw.dm
@@ -192,6 +192,12 @@
 	item_path = /obj/item/clothing/suit/armor/ego_gear/waw/exuviae
 	cost = 50
 
+/datum/ego_datum/exuviae
+	name = "Naked Nest Cure"
+	item_category = "Extract"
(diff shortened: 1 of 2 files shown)


Kitsunemitsu (approved):
Honestly I fucking forgot that we could do this

PR #1975, opened 2024-02-28
Fort Insanity fixes and improvements

Halonexus:
## About The Pull Request
Fixes the  fort insanity ai. Makes the fort insanity ai more robust in terms of pathing, damaging and destroying objects in their path, and being able to recover weapons after being disarmed.
Fort insanes can now also fire automatic guns at a decent rate and can multihit with fast melee weapons.
They can also target mechas now.
Additionally makes unarmed human attacks deal red damage instead of melee damage, and allows unarmed attack hit chance to scale with maxhealth, fist damage scales with fort and justice.
Added ego weapon stat check and attack cooldowns when attacking mechas and other objects.
Small runtime fixes for goldrush weapon and pinocchio weapon.

## Why It's Good For The Game
Fort insanes actually do stuff now instead of standing around swapping hands. Standardized the unarmed damage to work with our armor and large healthpools. 

## Changelog
:cl:
fix: fort insane ai works
tweak: ego weapons now have stat checks and attack cooldowns when attacking objects
/:cl:

### DIFF
--- a/code/datums/ai/sanity/_sanityloss_controller.dm
+++ b/code/datums/ai/sanity/_sanityloss_controller.dm
@@ -15,6 +15,8 @@
 	var/list/current_path = list()
 	var/next_smash = 0
 
+	var/timerid = null
+
 /datum/ai_controller/insane/TryPossessPawn(atom/new_pawn)
 	if(!ishuman(new_pawn))
 		return AI_CONTROLLER_INCOMPATIBLE
@@ -100,7 +102,6 @@
 	lines_type = /datum/ai_behavior/say_line/insanity_lines
 	resist_chance = 80 // Anger powered break out attempts
 	var/list/currently_scared = list()
-	var/timerid = null
 	var/interest = 3
 	var/asshole = FALSE
 
@@ -351,6 +352,8 @@
 		if(blackboard[BB_INSANE_BLACKLISTITEMS][I])
 			continue
 		if(blackboard[BB_INSANE_TEMPORARY_BLACKLISTITEMS][I])
+			continue
+		if(istype(I, /obj/item/offhand))
 			continue
 		var/item_force = GetEffectiveItemForce(I)
 		if(item_force <= INSANE_MINIMUM_WEAPON_FORCE)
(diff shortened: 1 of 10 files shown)


code/datums/ai/sanity/_sanityloss_controller.dm
```diff
@@ -155,48 +177,138 @@
...
 			pathing_attempts = 0
-
+			return
 	return TRUE
 
+/turf/proc/reachableTurftestWithMobs(caller, turf/T, ID, simulated_only)
+	if(T && !T.density && !(simulated_only && SSpathfinder.space_type_cache[T.type]) && !LinkBlockedWithAccess(T,caller, ID))
+		for(var/mob/living/L in T)
+			if(L.density)
+				return FALSE
```
LanceSmites328, line 192: More appropriate would be checking CanPass proc here from what I recall.

LanceSmites328 (changes requested):

code/datums/ai/sanity/_sanityloss_controller.dm
Halonexus: yea thats much better

code/datums/ai/sanity/_sanityloss_controller.dm
```diff
@@ -188,12 +188,7 @@
 /turf/proc/reachableTurftestWithMobs(caller, turf/T, ID, simulated_only)
 	if(T && !T.density && !(simulated_only && SSpathfinder.space_type_cache[T.type]) && !LinkBlockedWithAccess(T,caller, ID))
 		for(var/mob/living/L in T)
-			if(L.density)
-				return FALSE
-		for(var/obj/machinery/M in T)
-			if(istype(M, /obj/machinery/door))
-				continue
-			if(M.density)
+			if(!L.CanPass(caller, T))
```
LanceSmites328, line 191: This no longer counts for machinery in the path, such as doors, like this. The same proc would work for doors as and objects in general here as well. Actually it'd work for all atoms in the turf if you wanted to check that.

LanceSmites328 (changes requested):

code/datums/ai/sanity/_sanityloss_controller.dm
Halonexus: LinkBlockedWithAccess already does all the object checks

LanceSmites328 (dismissed):

Kitsunemitsu (approved):
Has been tested on the server for months.
Excellent PR.

PR #1976, opened 2024-03-01
Removes Miasma Producing Rot Component

InsightfulParasite:
## About The Pull Request
Having a component that produces miasma in the facility seemed like a drain on server resources. 
In my test the facility seems to still work as normal.

Its a bit haphazard since i simply looked up where the component was used. I'm unsure if there is some obscure area that relies on it.

## Why It's Good For The Game
Removes the main source of miasma production that is applied to every blood and gibs. This might help with lag from a abundance of gore.

## Changelog
:cl:
del: rot.dm
tweak: tweaked a few things
/:cl:


Kitsunemitsu (approved):
Approved

code/datums/components/rot.dm
```diff
@@ -1,74 +0,0 @@
...
-	var/atom/A = parent
-	
-	//SSprocessing goes off per 1 second
-	time_remaining -= delta_time * 1 SECONDS
-	if(time_remaining <= 0)
-		qdel(src)
-		return
-
-	var/turf/open/T = get_turf(A)
-	if(!istype(T) || T.planetary_atmos || T.return_air().return_pressure() > (WARNING_HIGH_PRESSURE - 10))
```
InsightfulParasite, line 30: I need to ask someone if this actually helps the processing of the game because due to the defines the component was apparently not producing any miasma but the process that is being called every decisecond is all of these procs.

PR #1986, opened 2024-03-09
Big and Will be Bad Wolf hp and work damage nerf

CrabbytheCrab:
## About The Pull Request

Lowers Big and Will be Bad Wolf's hp from 4000 to 2500 and work damage from 16 to 10
## Why It's Good For The Game

Wolf really shouldn't have the same hp as alephs(melting love, blue star, nothing there, and a few others have 4k)  while being a waw that runs away, takes normal damage at most, and can breach other stuff. So it having 2.5k would make it a bit bulkier compared to other waws hp wise without it having more hp than some alephs. Also why did it have the same work damage as nothing there when it should have been 10 like with the other red work damage waws?
## Changelog
:cl:
balance: Lowers Big and Will be Bad Wolf's hp to 2500 and work damage to 10.
fix: fixed a few things
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/waw/big_wolf.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/waw/big_wolf.dm
@@ -16,8 +16,8 @@
 	pixel_x = -16
 	base_pixel_x = -16
 
-	maxHealth = 2000
-	health = 2000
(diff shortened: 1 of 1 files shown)


Coxswain-Navigator (approved):
Much needed nerf

PR #1987, opened 2024-03-11
Buffs the shit out of Iron maiden

CrabbytheCrab:
## About The Pull Request

Buffs iron maiden's  multihits to scale with justice, made cooling it down only lowers the ramp up by 10.

Also reworks how it hurts you. Instead of doing a % of max hp  that scales when you attack, it now does red damage that scales the same way. It also now gibs you if you manage to get into crit or die from over using it. It also generates pe while getting hurt

## Why It's Good For The Game

Iron maiden just didn't work in practice due to how its multihits not scaling with justice really harmed it actually not being bad. It generating pe is silly and memey so it keeps it as a semi meme weapon while actually being able to put okish dps while doing do.

## Changelog
:cl:
balance: Made Iron Maiden an actual weapon now
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/special.dm
+++ b/code/game/objects/items/ego_weapons/special.dm
@@ -124,9 +124,9 @@
 /obj/item/ego_weapon/iron_maiden
 	name = "iron maiden"
 	desc = "Just open up the machine, step inside, and press the button to make it shut. Now everything will be just fine.."
-	special = "This weapon builds up the amount of times it hits as you attack, at maximum speed it will damage you per hit, increasing more and more, use it in hands."
+	special = "This weapon builds up the amount of times it hits as you attack, at maximum speed it will damage you per hit after attacking 5 times, increasing more and more, use it in hands to set back the speed it attacks at."
 	icon_state = "iron_maiden"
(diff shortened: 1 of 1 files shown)


Coxswain-Navigator:
How much of a buff is this? Remember that you can stack up some pretty high red defense with mimicry and knight of despair. What is the DPS?

CrabbytheCrab:
with the multihits scaling with justice its 256.833 dps
without the multihits scaling with justice its 176.25 at 130 justice

Kitsunemitsu:
Iron maiden DPS is REALLY good at max ramp.
It gets 100 DPS if you can max ramp it. 
at level 2 Ramp, you get 50 DPS (better than Waw) and then it goes 75 > 100.

It does not need a buff for an aleph weapon you get from a zayin, even if it does hurt you.

CrabbytheCrab:
justice scaling only actually effects 25% of that 100 dps which leads to stuff like crimson claw being capable to out dps it while trying to maximized the final stage's dps

CrabbytheCrab:
just want me to make a pr just to give those secondary hits justice scalling?

PR #1989, opened 2024-03-12
Reworks u-turn and voodoo

CrabbytheCrab:
## About The Pull Request

Changes Voodoo by removing its ability to do red, increased it's attack speed to 0.7, and now can be dual wielded with it self. It's a mini weapon now also

U-turn now attacks slower, no longer can charge up an aoe. It also now has a force of 40 and a reach of 4, but to counteract that insane range, it does half damage outside spear range and makes an aoe that drags stuff to the user.

## Why It's Good For The Game

Helps make these 2 ego more distinct. Also thank you kirie for suggesting me to do this.

## Changelog
:cl:
balance: Reworked u-turn and voodoo
/:cl:

### DIFF
--- a/code/datums/abnormality/_ego_datum/he.dm
+++ b/code/datums/abnormality/_ego_datum/he.dm
@@ -408,7 +408,7 @@
 	cost = 35
 
 /datum/ego_datum/weapon/voodoo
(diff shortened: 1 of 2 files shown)


Kitsunemitsu (approved):
Looks good to me.

PR #1995, opened 2024-03-19
Adds TGUI to the auxiliary console

Gboster-0:
## About The Pull Request

Makes it so instead of an UI, the auxiliary consoles give you TGUI
The UI is not removed, being kept as a failsafe in case TGUI ever fails for whatever reason. And can be toggled by players on alt+click

Le video:

https://github.com/vlggms/lobotomy-corp13/assets/82319946/7c754d88-fa73-4a05-a099-faf193d68e6a

## Why It's Good For The Game

TGUI has quite a few advantages compared to normal UI along with having these features:
- Has much more control over the looks of the console, making it nicer to look at
- Automatically updates LOB points, just a lil bit of QoL.
- The buttons change color to green/red depending on if you are able to buy the upgrades or not

And if TGUI on the console breaks, like it sometimes does. It does not prevent players from using the normal UI

## Changelog
:cl:
add: The auxiliary console can now has TGUI, you can turn it off by alt+clicking
/:cl:

### DIFF
--- a/code/game/machinery/computer/abnormality_auxiliary.dm
+++ b/code/game/machinery/computer/abnormality_auxiliary.dm
@@ -273,6 +273,19 @@
 				SSlobotomy_corp.available_core_suppressions = subtypesof(/datum/suppression)
 				update_static_data_for_all_viewers()
 
+			if("Disable Core Supression")
+				log_game("[usr] has used admin powers to disable all core supressions")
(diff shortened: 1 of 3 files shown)


Kitsunemitsu (approved):
Has been tested over the last few weeks.
Approved. has been excellent.

PR #1998, opened 2024-03-20
Makes Bread Smite make Bong Bongs into bongbread

tile4:
## About The Pull Request

What the title says. Essentially changes when somebody named “bong bong” is breaded into bongbread instead of regular bread

## Why It's Good For The Game

I think it is a neat addition and could be funny if this was the case.

## Changelog
:cl:
add: Added new things
/:cl:

### DIFF
--- a/code/modules/admin/smites/bread.dm
+++ b/code/modules/admin/smites/bread.dm
@@ -1,12 +1,15 @@
 #define BREADIFY_TIME (5 SECONDS)
 
-/// Turns the target into bread
+/// Turns the target into bread (Bongbread if they are Bong Bong)
 /datum/smite/bread
 	name = "Bread"
 
 /datum/smite/bread/effect(client/user, mob/living/target)
 	. = ..()
-	var/mutable_appearance/bread_appearance = mutable_appearance('icons/obj/food/burgerbread.dmi', "bread")
+    if target.name == "Bong Bong"
+         var/mutable_appearance/bread_appearance = mutable_appearance('icons/obj/food/burgerbread.dmi', "bongbread")
+	else
+         var/mutable_appearance/bread_appearance = mutable_appearance('icons/obj/food/burgerbread.dmi', "bread")
 	var/mutable_appearance/transform_scanline = mutable_appearance('icons/effects/effects.dmi', "transform_effect")
 	target.transformation_animation(bread_appearance,time = BREADIFY_TIME, transform_overlay=transform_scanline, reset_after=TRUE)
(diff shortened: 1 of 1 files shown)


code/modules/admin/smites/bread.dm
```diff
@@ -1,12 +1,15 @@
...
 
-/// Turns the target into bread
+/// Turns the target into bread (Bongbread if they are Bong Bong)
 /datum/smite/bread
 	name = "Bread"
 
 /datum/smite/bread/effect(client/user, mob/living/target)
 	. = ..()
-	var/mutable_appearance/bread_appearance = mutable_appearance('icons/obj/food/burgerbread.dmi', "bread")
+    if target.name == "Bong Bong"
```
Kitsunemitsu, line 9: (target.name == "Bong Bong")

Need to bracket it chief

tile4: Fixed

Kitsunemitsu:
Christ, it's all broken
give me a bit to take a look

Gboster-0:
its spaces on the added code instead of tabs, change it to tabs please

tile4:
Sorry, was on my phone when I did this which didnt have a tab button in the built in keyboard as I know. So there. Will double check if I missed any but I think that's it

code/modules/admin/smites/bread.dm
```diff
@@ -1,12 +1,15 @@
 #define BREADIFY_TIME (5 SECONDS)
 
-/// Turns the target into bread
+/// Turns the target into bread (Bongbread if they are Bong Bong)
 /datum/smite/bread
 	name = "Bread"
 
 /datum/smite/bread/effect(client/user, mob/living/target)
 	. = ..()
-	var/mutable_appearance/bread_appearance = mutable_appearance('icons/obj/food/burgerbread.dmi', "bread")
```
Gboster-0, line 9: by the way, with how DM compiles you need to have var/mutable_appearance/bread_appearance always defined as a variable before code uses it, even if its defined by the if()else() check that always gives an output
in other words, keep the line above as
```
var/mutable_appearance/bread_appearance
``` and the code should be good

tile4: So like that? Or did I misunderstand?

Gboster-0: yep, but now you have to not re-define the same variable. So remove the other instances of `var/mutable_appearance/bread_appearance` replacing them with just `bread_appearance`

tile4: Okay, there we go. Hope that is good?

Gboster-0: Yep, should work perfectly well

PR #2000, opened 2024-03-22
Returns Adventure Console Defines to Numbers

InsightfulParasite:
## About The Pull Request
Areas where the change to flags kinda broke the code in adventure_layout.dm. Adventure proc in line 102
TravelUI proc in line 164
AdventureModeReact proc in line 263

The href also returns a number solution of the define rather than the flat text so thats also messed up on the console.

## Changelog
:cl:
fix: Adventure Console
fix: running from combat used to just say you ran away but didnt actually end combat.
tweak: adventure console travel UI code
/:cl:


Kitsunemitsu (approved):
While this is worse code; we do need to keep in mind that this dev needs to work on this PR and has been having trouble with it lately.

PR #2013, opened 2024-03-29
Makes knockback a variable on all EGO weapons, attack_speed variable clean up and correcting bad descriptions

Gboster-0:
## About The Pull Request

- `knockback` is now a variable on all EGO weapons

- `attack_speed` is now a switch instead of a million if else() statements in the examine text

- Removes `return FALSE`, replacing it with `return` on some pieces of code

- Replaces the "Knocks certain enemies backwards" special knockback descriptions as a side effect of point 1, correcting the "certain enemies" part accidentally

- Changes the oberon's attacks to throw people shorter distances, but at a faster speed as a consequence of making knockbacks standard, this is the only knockback that was not in-line with others

## Why It's Good For The Game

>`knockback` is now a variable on all EGO weapons
- This allows us to do less copy-paste code and admins can set knockback on any EGO weapon they desire (this can't possibly go wrong, can it?)

>`attack_speed` is now a switch instead of a million if else() statements in the examine text
- It was supposed to be a switch, but nobody knew how to make one in the original PR and then everyone forgot about it

>Removes `return FALSE`, replacing it with `return` on some pieces of code
- return FALSE only matters for children of objects, really. and i dont think we are going to make children of EGO weapons

## Changelog
:cl:
admin: admins can now set knockback on ego weapons to any value they desire by editing the "knockback" variable
/:cl:

### DIFF
--- a/code/__DEFINES/~lobotomy_defines/weapon.dm
+++ b/code/__DEFINES/~lobotomy_defines/weapon.dm
@@ -1,0 +1,4 @@
+// Knockback defines
+#define KNOCKBACK_LIGHT (1<<0)
+#define KNOCKBACK_MEDIUM (1<<1)
+#define KNOCKBACK_HEAVY (1<<2)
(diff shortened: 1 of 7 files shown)


Kitsunemitsu (approved):
Love it.

PR #2024, opened 2024-04-03
Makes Porccubus able to dash on any tile, adds it to Rcorp

TaculoTaculo:
## About The Pull Request
Lets porccubus target atoms for the dash instead of living targets so that players can move around when possesing it, and adds it to rcorp as a EasyTank (it has full immunity to bullets).
Gives it slightly faster dash speed on combat maps.

## Why It's Good For The Game
More abnos with playables support, more abnos for Rcorp and LCL too.

## Changelog
:cl:
add: porccubus to rcorp
tweak: tweaked porccubus to let it dash on any tile instead of only on human targets
balance: gave porccubus lower dash cooldown on rcorp
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/rcorp/!abno_overwrites.dm
+++ b/ModularTegustation/tegu_items/rcorp/!abno_overwrites.dm
@@ -49,3 +49,11 @@
 	if(IsCombatMap())
 		return
 	..()
+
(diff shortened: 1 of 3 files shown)


Kitsunemitsu (approved):
Approved. Good PR.

PR #2033, opened 2024-04-04
Voices of the Slab - the cap changening

KingOfJacks:
## About The Pull Request
Changes the cap for Viva voice and Theonite slab from 50 to 100

## Why It's Good For The Game
first off, found the consistency a little silly when compared to the other two stat boosters. and second, given that you'll be taking extra damage against one of the most common damage type and one of the most dangerous damage type in black and pale respectively, even though you'll do works better and be faster/do more damage, especially in late game you can get barred from WaW+ and Aleph abnos either because you cant work them or because their breach would annihilate you.

this is also my first time messing with code and thought this would be simple enough without any actual knowledge (please dont kill me)

## Changelog
:cl:
tweak: changes viva and slab to be consistent with the other stat tools
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_tools/he/theonite_slab.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_tools/he/theonite_slab.dm
@@ -19,7 +19,7 @@
 		return //You don't need any more.
 
 	flick(icon_state, src)
-	user.adjust_attribute_buff(JUSTICE_ATTRIBUTE, 5)
+	user.adjust_attribute_buff(JUSTICE_ATTRIBUTE, 10)
 	var/datum/status_effect/stacking/slab/S = user.has_status_effect(/datum/status_effect/stacking/slab)
 	if(!(user in users))
 		users += user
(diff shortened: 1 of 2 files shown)


Gboster-0:
tbh, now that we have tougher times scaling on temperance this is more passable than back when the tool was made

i will say, preferably make it give 10 temperance/ justice to make it more in-line with other tools in its total multiplayer and to reduce the spam clicking needed

Coxswain-Navigator:
100 temperance is a lot to give players for free. Even with the scaling we're still going to end up with 100% work rates on a lot of things.

Gboster-0:
> 100 temperance is a lot to give players for free. Even with the scaling we're still going to end up with 100% work rates on a lot of things.

i will add, it is definetelly NOT for free, its 14,2 times the black damage at the moment, since you get 1.15x more black damage per 5 temperance

KingOfJacks:
changed it to give 10 stats over 5. and as an idea, could make it so they give more vulnerability so using them is more punishing for their damage types

CrabbytheCrab:
give them the same scaling as snake oil and skin proph

Coxswain-Navigator:
I still think it is better to give them less stats with a not as harsh penalty. Giving 100 temperance or 100 justice can really break game balance versus giving out 100 hp or sp

PR #2043, opened 2024-04-08
EO is now immune from EGO damage from non-insane agents.

Kitsunemitsu:
## About The Pull Request
EO and sephirah are both now immune to EGO damage from non insane agents.

## Why It's Good For The Game
People think it's funny to kill these two roles that affect their specific round negatively. 
Sephirah for randomizing abnos and EO for making extraction 5 seconds slower.
I'm removing the functionality to kill them at your whim.

## Changelog
:cl:
balance: Sane agents can no longer kill roles that are slightly harmful to their specific round.
/:cl:


Kitsunemitsu:
This was a temporary PR to get people to fuck off during an experimental EO change and is no longer expected.

PR #2044, opened 2024-04-08
Extraction Shipments

Kitsunemitsu:
## About The Pull Request
Extracting as non-EO now gives an audible message as well as takes 15 seconds to pod in.
It prefers to land on a table but if there is no table it will land on the machine itself.

## Why It's Good For The Game
People have started stealing EGO again, even when they're reserved. This should encourage them to ask the EO instead of just gunning past them.
Unlike the last PR, this one is a click and it'll arrive, not a click and sit there for 30 seconds.

Players also complain a LOT about things that can be easily addressed via communication; and I want to encourage more communication.
I feel that we're in a point in our game's life where we can start to do things like this.

The game is ultimately a multiplayer game with a focus on communication.
We need to start punishing people who don't want to communicate.

## Changelog
:cl:
add: Added Extraction shipments.
/:cl:

### DIFF
--- a/code/game/machinery/computer/abnormality_ego.dm
+++ b/code/game/machinery/computer/abnormality_ego.dm
@@ -4,7 +4,7 @@
 	resistance_flags = INDESTRUCTIBLE
 	/// Currently selected(shown) level of abnormalities whose EGO will be on the interface
 	var/selected_level = ZAYIN_LEVEL
-	var/delay = 5 SECONDS
+	var/delay = 15 SECONDS
 
 /obj/machinery/computer/ego_purchase/ui_interact(mob/user)
 	. = ..()
--- a/code/game/machinery/computer/abnormality_ego.dm
+++ b/code/game/machinery/computer/abnormality_ego.dm
@@ -63,7 +63,7 @@
 
 			else
 				addtimer(CALLBACK(src, PROC_REF(ShipOut), E.item_path),delay)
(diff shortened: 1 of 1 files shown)


code/game/machinery/computer/abnormality_ego.dm
```diff
@@ -56,16 +57,39 @@
...
+			if(usr.mind.assigned_role == "Extraction Officer")
+				new E.item_path(get_turf(src))
+				to_chat(usr, span_notice("[E.item_path] has been dispensed!"))
+
+			else
+				addtimer(CALLBACK(src, PROC_REF(ShipOut), E.item_path),delay)
+				audible_message(span_notice("[usr.name] has purchased a [E.item_path]"))
+
 			A.stored_boxes -= E.cost
-			A.current_ego += I
```
Gboster-0, line 61: why was this line removed?

Kitsunemitsu: Ah yes, so basically
This was a temporary feature to only allow X amount of EGO per round.
It was removed because to keep it would increase the shitcode of this PR by a bit, and it was never a feature that existed.

As for if I should remove it there?
I will do it in a future PR as it requires Egor's specific permission, and this PR is an initial step.

Coxswain-Navigator (approved):
This is a feature that really helps the game state, even if it's a bit of a negative for solo players.

PR #2045, opened 2024-04-09
Moves most duplicate alt refinery checks to the parent and some cleanup

Gboster-0:
## About The Pull Request

Moves the following to the alt refineries parent:
- extraction officer check
- PE check (disabled by default)
- Removing PE

Replaces timed refineries 5 lines of code with 2, using for(var/i in 1 to 5) instead of copy-pasting new() 5 times

Makes the machines tell you that you need to input in an item if they need it both on examine and trying to click them with one.
Also causes the refineries to call attack_hand() if you touch them with an item whilst they do not need any items

Also added mapload argument to the refineries initialize, because im quite sure its supposed to be there

## Why It's Good For The Game

Bit less copy-paste

## Changelog
:cl:
qol: PE refineries will now tell you if you need to insert in an item
code: Moved some code around in alt refineries
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/refinery/alt_refinery/_refiner.dm
+++ b/ModularTegustation/tegu_items/refinery/alt_refinery/_refiner.dm
@@ -47,8 +47,8 @@
 /obj/structure/altrefiner/attackby(obj/item/I, mob/living/user, params)
 	. = ..()
 	if(!requires_item)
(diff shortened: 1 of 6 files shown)


Kitsunemitsu (approved):
Approved. A quick refactor that I should have done but it escaped my notice.

PR #2055, opened 2024-04-15
Silent Girl portrait change

BLUREMOTE:
## About The Pull Request
While I fully get that the current portrait is the one found in the artbook, it is not square, so it get streched out, and honestly, having to manually redraw it on the side doesnt seem worth it. So I swapped it to one of her card art from ruina.


Coxswain-Navigator:
That image is actually ripped from the game's code and not from the artbook. It's the real deal.

Kitsunemitsu:
Closed to above, Please understand.

PR #2058, opened 2024-04-16
Minor Bullet Changes

Kitsunemitsu:
## About The Pull Request
Makes some changes to some bullets to make purchasing not HP bullets a little more desirable.

- Shield bullets now last twice as long
- Slow bullets now last 50% longer
- HP and SP bullets now cost more after the first purchase, as well as heal significantly less.

## Why It's Good For The Game
Managers pretty much always max out Healing bullets immediately, and I don't really think that this is good design.
I want shield bullets to be more desirable as well as HP/SP bullets. 

I gave general buffs to the length of both Shield and slow bullets to make those more desirable

## Changelog
:cl:
balance: rebalanced manager bullets in general
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/gadgets/manager_bullets.dm
+++ b/ModularTegustation/tegu_items/gadgets/manager_bullets.dm
@@ -33,7 +33,7 @@
 
 /datum/status_effect/interventionshield
 	id = "physical intervention shield"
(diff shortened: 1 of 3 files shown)


Coxswain-Navigator (approved):
Very good changes

PR #2056, opened 2024-04-16
Stat scaling rebalance

tile4:
## About The Pull Request
Adds a point where you can get stats at 70s and stat scaling should happen sooner.

## Why It's Good For The Game
This will make it more encouraging to join late or come back after dying and catching up.

## Changelog
:cl:
balance: rebalanced something
/:cl:

PR #2060, opened 2024-04-16
[Do not merge until #2056 is merged] Improves job outfit code

Gboster-0:
## About The Pull Request

- Replaces `. = ..()` with `return ..()` where applicable
- Replaces `.=..()` with `. = ..()`, we have the technology for SPACES
- Set lists to the standard list format
- Removes cases where attributes were being set to 0, whilst already being 0 due to it being the default
- Moved backpack contents to the end of outfit code, since they are most often changed and lists in general dont fit nicelly between other one liners
- Replaced `mob/living/carbon/human/H` with `mob/living/carbon/human/outfit_owner`, since imo it describes the variable better
- Removed backpack contents having the format `item = 1`, instead use `item` when the value is 1 since thats the default, the format `item = 1` is to be used when the value is higher than 1

## Why It's Good For The Game

moar readability

## Changelog
:cl:
code: (improved?) some code
/:cl:


Kitsunemitsu (approved):
Mostly some easy changes.
A PR will need to be reworked; but that's fine.

PR #2061, opened 2024-04-16
Makes Smile non-dense in LCL and RCA

KingSuperBOB:
## About The Pull Request
Smile gains non-density in combat maps

## Why It's Good For The Game
This is a teeny tiny buff to an abnormality that gets two shot by a stray rhino off screen 

## Changelog
:cl:
add: Added non-density to smile in combat maps
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/teth/simple_smile.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/teth/simple_smile.dm
@@ -75,9 +75,9 @@
 	src.pulled(pick(pullable))
 
 /mob/living/simple_animal/hostile/abnormality/smile/Initialize()
-  ..()
-  if(IsCombatMap())
-    density = FALSE
+	..()
+	if(IsCombatMap())
+    	density = FALSE
 
 /mob/living/simple_animal/hostile/abnormality/smile/PostWorkEffect(mob/living/carbon/human/user, work_type, pe, work_time)
 	if(lucky_counter > 3)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/abnormality/teth/simple_smile.dm
```diff
@@ -74,6 +74,10 @@
 
 	src.pulled(pick(pullable))
 
+/mob/living/simple_animal/hostile/abnormality/smile/Initialize()
+  if(IsCombatMap())
+    density = FALSE
+
```
TaculoTaculo, line 80:
```suggestion
/mob/living/simple_animal/hostile/abnormality/smile/Initialize()
  ..()
  if(IsCombatMap())
    density = FALSE

```
You need the ..() there else it breaks the abno.

TaculoTaculo: Thinking a bit more and this change would be better being moved to !abno_overwrites.dm

KingSuperBOB: Id rather keep it on smile rather than moving it to overwrites since this change is intended for both modes considering Smile was in LCL previously and has been used by admins for LCL and its also a buff directed at rca too

code/modules/mob/living/simple_animal/abnormality/teth/simple_smile.dm
```diff
@@ -74,6 +74,11 @@
 
 	src.pulled(pick(pullable))
 
+/mob/living/simple_animal/hostile/abnormality/smile/Initialize()
+  ..()
```
Kitsunemitsu, line 78: Space indentation

Kitsunemitsu:
After seeing Simple Smile on a recent RCA round, no. They are annoying enough.

PR #2068, opened 2024-04-21
Pino ai fix

Halonexus:
## About The Pull Request
Pinocchio mob will not target same faction members now.
Additionally fixes the spawn abno verb logging.

## Why It's Good For The Game
Pinocchio wont solo all the abnormalities anymore

## Changelog
:cl:
fix: fixed pinocchio targeting
/:cl:

### DIFF
--- a/code/datums/ai/sanity/sanityloss_behaviors.dm
+++ b/code/datums/ai/sanity/sanityloss_behaviors.dm
@@ -424,9 +424,6 @@
 			controller.pathing_attempts = 0
 			controller.current_path.Cut()
 			finish_action(controller, TRUE)
-			if(istype(controller, /datum/ai_controller/insane/murder))
-				var/datum/ai_controller/insane/murder/M = controller
-				M.FindEnemies()
 		return FALSE
 	// Movement
 	if(LAZYLEN(controller.current_path))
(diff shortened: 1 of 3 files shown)


code/datums/ai/sanity/sanityloss_behaviors.dm
```diff
@@ -490,17 +487,7 @@
 	// Same as the above insanity, but they look for a target between moves.
 
 /datum/ai_behavior/insanity_wander/murder_wander/PreMoveCheck(datum/ai_controller/insane/murder/controller, mob/living/living_pawn)
-	for(var/mob/living/L in livinginview(9, living_pawn))
```
Kitsunemitsu, line 493: Did you fuck this up?

Halonexus: its doing the same checks here as FindEnemies and if we want to override targetting need to make it look at the proc that we are overriding.

Halonexus: Otherwise it will stop wander behaviour even though there are no attackable enemies and be stuck there.

Kitsunemitsu (approved):
Yep, looks to be a quick and easy move. ready to go

PR #2075, opened 2024-04-22
Agent attribute bugfix

tile4:
## About The Pull Request

Fixes an issue where 19 out of 24 abnos were between two stat change intervals.

## Why It's Good For The Game

19/24 abnormalities shouldn't spawn people at 20s

## Changelog
:cl:
fix: fixed a few things
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/agent.dm
+++ b/code/modules/jobs/job_types/agent.dm
@@ -82,22 +82,22 @@
 		facility_full_percentage = 100 * (SSabnormality_queue.spawned_abnos / SSabnormality_queue.rooms_start)
 	// how full the facility is, from 0 abnormalities out of 24 cells being 0% and 24/24 cells being 100%
 	switch(facility_full_percentage)
-		if(15 to 29.99) // Shouldn't be anything more than TETHs (4 Abnormalities)
+		if(15 to 29) // Shouldn't be anything more than TETHs (4 Abnormalities)
 			set_attribute *= 1.5
 
-		if(30 to 44.99) // HEs (8 Abnormalities)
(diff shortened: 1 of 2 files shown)


code/modules/jobs/job_types/agent.dm
```diff
@@ -82,19 +82,19 @@
 		facility_full_percentage = 100 * (SSabnormality_queue.spawned_abnos / SSabnormality_queue.rooms_start)
 	// how full the facility is, from 0 abnormalities out of 24 cells being 0% and 24/24 cells being 100%
 	switch(facility_full_percentage)
-		if(15 to 29) // Shouldn't be anything more than TETHs (4 Abnormalities)
+		if(15 to 29.99) // Shouldn't be anything more than TETHs (4 Abnormalities)
```
Kitsunemitsu, line 85: You can do 15 to 29 here, and then 29 to 44 on the next one.
Sadly I can't approve this as it doesn't follow conventions </3

Kitsunemitsu (changes requested):
Requesting changes, as this is slightly more shitcode than I'm willing to approve.

code/modules/jobs/job_types/agent.dm
tile4: Oh, it does? neat! Fixed the code to what you said.

tile4:
Fixed so should be good to merge now. In fact I noticed an oversight so I fixed it too.

Gboster-0:
tbh, `var/facility_full_percentage` should just be rounded to the nearest full digit, but this should work

Kitsunemitsu (approved):
Approved. Looks great.

PR #2077, opened 2024-04-22
Wounded Courage lets you dual wield Blind Rage

BLUREMOTE:
## About The Pull Request
If you are wearing Wounded Courage (the realization of Blind Rage, Servant of Wrath's EGO), you are given the ability to use a Blind Rage weapon in each hand and attack with both at the same time in 1 attack.
**People are still vulnerable to the AOE, so be careful**
## Why It's Good For The Game
Main few reasons of this addition:
- Thematically, it's accurate to what it's refering to, as SOW monster form have her use 2 hammer shaped hands for devasting AOE attacks
- This gives a significant damage increase potential to a WAW weapon. With the dual wield feature, its slightly weaker than an ALEPH weapon, but start outdamaging it when faced with multiple enemies.
- With the potential of seemingly "infinite" dps as you have an AOE attack, you now run the risk of causing greater harms to your friends. This goes back to the first point as being thematically accurate
- Makes a somewhat "eh" realization more interesting to play. Being on the squishier side in regards to white (30) and pale somewhat (50) due to its somewhat average ability, this should bring it in par with more "usable" realizations, as the advantage granted can be quite significant in the right scenario if you are fine with the lower resistances.

### DIFF
--- a/code/game/objects/items/ego_weapons/waw.dm
+++ b/code/game/objects/items/ego_weapons/waw.dm
@@ -1031,7 +1031,6 @@
 	. = ..()
 	if(!.)
 		return FALSE
-
 	attacks++
 	attacks %= 3
 	switch(attacks)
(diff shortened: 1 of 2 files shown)


Kitsunemitsu (commented):
What other ability does it have?

BLUREMOTE:
> What other ability does it have?

Currently, Wounded Couraged has the "Justice and Balance" skill, which let the user perform 3 5x5 AOEs that deal black damage (with no friendly fire), and grant a decaying 2 justice buff for every target hit. This ability is meant to be used against groups, as it decays quite fast and hitting only a few target will give a short last justice buff.

I always find it odd doe that this skill give the armor "support" role. What are you really supporting here beside yourself, with the buff having limited applications (only becomes useful during midnight ordeal **if** its an ordeal with lots of enemies)

Kitsunemitsu:
Increase stats to 230 and I'll merge.

Kitsunemitsu (approved):
Stats were brought up to line. 
Ready to merge.

PR #2081, opened 2024-04-24
Fixes Amber Ordeal Runtime on Dungeon Map

InsightfulParasite:
## About The Pull Request
Amber ordeals were stuck picking from a empty list without turning off their own burrowing.

## Changelog
:cl:
fix: Amber Ordeals
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/amber.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/amber.dm
@@ -103,7 +103,11 @@
 
 /mob/living/simple_animal/hostile/ordeal/amber_bug/proc/BurrowIn(turf/T)
 	if(!T)
-		T = pick(GLOB.xeno_spawn)
+		if(GLOB.xeno_spawn.len)
+			T = pick(GLOB.xeno_spawn)
+		else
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/ordeal/amber.dm
```diff
@@ -103,7 +103,11 @@
 
 /mob/living/simple_animal/hostile/ordeal/amber_bug/proc/BurrowIn(turf/T)
 	if(!T)
-		T = pick(GLOB.xeno_spawn)
+		if(GLOB.xeno_spawn.len)
+			T = pick(GLOB.xeno_spawn)
+		else
+			can_burrow_solo = FALSE
+			T = get_turf(src)
```
Kitsunemitsu, line 110: Can you return here or would that cause bugs?

Kitsunemitsu (commented):
Sorry, just a question

code/modules/mob/living/simple_animal/hostile/ordeal/amber.dm
InsightfulParasite: I can return. The continue was just for the dramatic entrance.

PR #2082, opened 2024-04-24
Minor Contract buffs

Kitsunemitsu:
## About The Pull Request
Buffs contract in a few ways. 

Work rates increase by 5 across the board
4 Contracts can now be made per work type

Having a contract at all now gives you a small work damage reduction.

Now creates contracts on a neutral.

Chance to lose qliphoth on a neutral reduced.

## Why It's Good For The Game
Contract is a very interesting abno that is let down with bad work rates which compound with pale damage.
It's gimmick also only shows up on a good.
I want to rectify all 3 of these with minor buffs.

## Changelog
:cl:
balance: rebalanced A Contract, Signed
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/waw/contract.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/waw/contract.dm
@@ -90,7 +90,7 @@
 			if(user in just_havers)
 				return TRUE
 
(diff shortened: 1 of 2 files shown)


Coxswain-Navigator (approved):
This is a big improvement to the current state of the abnormality, although that's not to say I don't think the contracts could be a bit more interesting.

PR #2091, opened 2024-04-26
Mob Speed Improvements

Halonexus:
## About The Pull Request
Makes hostile mob movement speed behave the same way damage coefficients do: move_to_delay is a final value that must not be touched, speed var stores the base delay value that can be changed through ChangeMoveToDelay, ChangeMoveToDelayBy, and TemporarySpeedChange procs.
Movement modifiers affect mob move_to_delay.
Status effects that add movement speed modifiers dont change speed var but do affect move_to_delay properly.
Makes hostile mob move_to_delay match client move delay.
Allows for real speed multipliers on all mobs.

## Why It's Good For The Game
Makes it easier to deal with mob speeds.
Fixes issues with status effects that adjusted move_to_delay directly turbocharging mobs due to them also changing their own move_to_delay during the effect.

## Changelog
:cl:
tweak: can now do real speed multipliers
fix: certain mobs wont become super fast when slowdowns expire
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -981,7 +981,7 @@
 
 
 //~~~LC13 General Debuffs~~~
-#define CARBON_HALFSPEED /datum/movespeed_modifier/qliphothoverload
+#define MOB_HALFSPEED /datum/movespeed_modifier/qliphothoverload
 /datum/status_effect/qliphothoverload
 	id = "qliphoth intervention field"
 	duration = 15 SECONDS
@@ -991,12 +991,7 @@
 
 /datum/status_effect/qliphothoverload/on_apply()
 	. = ..()
-	if(ishostile(owner))
-		var/mob/living/simple_animal/hostile/L = owner
-		L.TemporarySpeedChange(4, duration)
-	if(iscarbon(owner))
-		var/mob/living/carbon/M = owner
-		M.add_movespeed_modifier(CARBON_HALFSPEED)
+	owner.add_movespeed_modifier(MOB_HALFSPEED)
 
 	var/mutable_appearance/effectvisual = mutable_appearance('icons/obj/clockwork_objects.dmi', "vanguard")
 	effectvisual.pixel_x = -owner.pixel_x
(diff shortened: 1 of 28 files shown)


Kitsunemitsu:
Need to read this.

code/datums/status_effects/debuffs.dm
```diff
@@ -991,12 +991,7 @@
 
 /datum/status_effect/qliphothoverload/on_apply()
 	. = ..()
-	if(ishostile(owner))
-		var/mob/living/simple_animal/hostile/L = owner
-		L.TemporarySpeedChange(4, duration)
-	if(iscarbon(owner))
-		var/mob/living/carbon/M = owner
-		M.add_movespeed_modifier(CARBON_HALFSPEED)
+	owner.add_movespeed_modifier(MOB_HALFSPEED)
```
InsightfulParasite, line 994: This does make 6 lines of code into one line of code. Which is always good.

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -222,21 +223,25 @@
...
-	. = ..()
-	move_to_delay = (initial(move_to_delay) + (staminaloss * 0.06))
-	UpdateSpeed()
+	if(staminaloss == 0)
+		remove_movespeed_modifier(/datum/movespeed_modifier/hostile_stamina_loss, TRUE)
+	else
+		add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/hostile_stamina_loss, TRUE, staminaloss * 0.06)
 
+///Use ChangeMoveToDelayBy instead for proper permanent speed changes
 /mob/living/simple_animal/hostile/proc/SpeedChange(amount = 0)
```
InsightfulParasite, line 232: So is this for altering the base speed without any modifiers or is this a proc that cant be reworked.

Halonexus: After initialization move_to_delay will be calculated from all the movespeed modifiers, so using this would set the speed modifier to be the sum of all modifiers combined

Kitsunemitsu (approved):
Was tested on the server; and had no issues at all.

Excellent refactor PR

PR #2095, opened 2024-04-27
Hostile ai tweak

Halonexus:
## About The Pull Request
Makes the mob ai more opportunistic about attacking other enemies while main target is out of reach.
Implements actual melee attack delay for ai (attack_cooldown in deciseconds), converts existing rapid_melee values.
AttackingTarget proc now actually uses the attacked_target parameter to allow attacking things other than the aggro holder.

https://github.com/vlggms/lobotomy-corp13/assets/23573057/f4f6d0e1-75da-461a-918e-d79c6ddbae7f

Optimised ListTargets.
Shortened patience loss time to 5 seconds, now mob has to lose target, lose patience or get enough aggro to overpower current target aggro + target_switch_resistance(about 15% maxhp worth of aggro) to switch targets instead of it being random upon taking damage. 
Simple animals now properly add aggro upon attacking, which fixes mobs always prioritizing attacking other mobs instead of humans.
Hostiles now gain patience upon being attacked by the target in melee, this lets mobs without melee attacks such as Arbiter keep the same target so long as you keep attacking them.
Added AGGRO_DAMAGE damage type for adding aggro without dealing damage.
Shield type ego now deal aggro damage, that scales with the average of fortitude and prudence, in a 3 tile radius on block(regardless of any attacks getting blocked)

## Why It's Good For The Game
Would make kiting strong enemies less effective.
Can customize mob attack cooldowns more precisely.

## Changelog
:cl:
tweak: Mob ai will attack anyone in range while main target is too far
code: AttackingTarget now uses the attacked_target parameter, mob melee attack speed is now using attack_cooldown instead of rapid_melee
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -695,15 +695,17 @@
 		target and they suddenly or are currently something
 		we dont attack.*/
 	if(QDELETED(target))
-		if(approaching_target)
-			/* Approaching target means we are currently moving menacingly
-				towards something. Otherwise we are just moving and if we
-				are investigating a location we dont want to be told to stand still. */
-			LoseTarget()
-		return FALSE
+		if(!FindTarget(possible_targets, TRUE))
+			if(approaching_target)
+				/* Approaching target means we are currently moving menacingly
+					towards something. Otherwise we are just moving and if we
+					are investigating a location we dont want to be told to stand still. */
+				LoseTarget()
+			return FALSE
 	if(!CanAttack(target))
-		LoseTarget()
-		return FALSE
+		if(!FindTarget(possible_targets, TRUE))
+			LoseTarget()
+			return FALSE
 	// The target we currently have is in our view and we must decide if we move towards it more.
 	if(target in possible_targets)
 		var/turf/T = get_turf(src)
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -24,13 +24,12 @@
 	var/list/emote_taunt = list()
 	var/taunt_chance = 0
 
-	var/rapid_melee = 1			 //Number of melee attacks between each npc pool tick. Spread evenly.
+	var/rapid_melee = 1			 //Number of melee attacks between each npc pool tick. Spread evenly. !!OBSOLETE!! It still works but use attack_cooldown instead.
 	var/melee_queue_distance = 4 //If target is close enough start preparing to hit them if we have rapid_melee enabled
 	var/melee_reach = 1			 // The range at which a mob can make melee attacks
-	var/attack_cooldown = 0
+	var/attack_cooldown = 0      //The time between attacks in deciseconds. If 0 at initialization then uses rapid_melee to calculate the initial value.
 	var/attack_is_on_cooldown = FALSE
-	var/attack_timer_id = null
-	var/old_rapid_melee = 0
+	var/old_rapid_melee = 0      //used for compatibility with old rapid_melee system to detect outside code changing the rapid_melee value and adjusts attack_cooldown accordingly
 
 	var/ranged_message = "fires" //Fluff text for ranged mobs
 	var/ranged_cooldown = 0 //What the current cooldown on ranged attacks is, generally world.time + ranged_cooldown_time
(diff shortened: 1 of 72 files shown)


code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -543,12 +554,52 @@
...
+		TryAttack()
+
+/mob/living/simple_animal/hostile/proc/TryAttack()
+	if(client || stat != CONSCIOUS || AIStatus != AI_ON || incapacitated() || !targets_from || !isturf(targets_from.loc))
+		attack_is_on_cooldown = FALSE
+		if(attack_timer_id)
+			deltimer(attack_timer_id)
+			attack_timer_id = null
+		return
+	if(target && (target.Adjacent(targets_from) || melee_reach > 1 && (target in view(melee_reach, targets_from))))
```
InsightfulParasite, line 580: Im going to test this on local to see if it get any lag or runtimes when spawning a bunch of monsters that attack in different ways.

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -122,8 +132,9 @@
 	var/list/possible_targets = ListTargets()
 	if(environment_smash)
 		EscapeConfinement()
-
 	if(AICanContinue(possible_targets))
+		if(!attack_is_on_cooldown)
+			TryAttack()
```
InsightfulParasite, line 137: Im going to test this with monsters that attack in unconventional ways like saplings to see if theres any weirdness.

InsightfulParasite: Found no issues with Wolf, Snow Whites Apple, or Parasite Tree Saplings. Those are the 3 i can think of who have unconvientional attack methods.

InsightfulParasite (approved):

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -543,12 +554,52 @@
...
+	. = ..()
+	if(!client && AIStatus == AI_ON && !attack_is_on_cooldown && CanAttack(AM))
+		TryAttack()
+
+/mob/living/simple_animal/hostile/Moved()
+	. = ..()
+	if(!client && AIStatus == AI_ON && target && !attack_is_on_cooldown)
+		TryAttack()
+
+/mob/living/simple_animal/hostile/proc/TryAttack()
```
InsightfulParasite, line 573: Found no issues with Wolf, Snow Whites Apple, or Parasite Tree Saplings. Those are the 3 i can think of who have unconvientional attack methods.

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -543,12 +554,54 @@
...
+	. = ..()
+	if(!client && AIStatus == AI_ON && !attack_is_on_cooldown && CanAttack(AM))
+		TryAttack()
+
+/mob/living/simple_animal/hostile/Moved()
+	. = ..()
+	if(!client && AIStatus == AI_ON && target && !attack_is_on_cooldown)
+		TryAttack()
+
+/mob/living/simple_animal/hostile/proc/TryAttack(called_by_timer = FALSE)
```
LanceSmites328, line 573: CheckAndAttack seems otherwise abandoned here. I believe it could and should be implemented for better range detection along with a more efficient check; the previous does not run `view` if the user does not have increased melee range to reduce calculations, the newer one I believe runs `view` regardless.

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -603,13 +656,6 @@
 
 		//This is for attacking.
 		if(target)
-			if(targets_from && isturf(targets_from.loc) && in_range)
-				//If they're next to us, attack
-				MeleeAction()
-			else
-				if(rapid_melee > 1 && target_distance <= melee_queue_distance)
-					MeleeAction(FALSE)
-				in_melee = FALSE //If we're just preparing to strike do not enter sidestep mode
```
LanceSmites328, line 612: I'm ASSUMING this works by virtue of "Attacking is done automatically" correct?

LanceSmites328 (dismissed):

code/modules/mob/living/simple_animal/hostile/hostile.dm
Halonexus: This whole proc is basically check and attack and normally if the first part of && is false the second is not evaluated

code/modules/mob/living/simple_animal/hostile/hostile.dm
Halonexus: its moved into handle automated action instead of the move code as well as a response to bump and on move

InsightfulParasite:
I didnt consider having items themselves alter aggro value. Great job.

InsightfulParasite:
I forgot test merge candidate fucks up PR's but apparently dont merge doesnt???

InsightfulParasite (approved):
Apparently was testmerged before. Looks good. Approved.

PR #2105, opened 2024-04-30
Vermillion no longer gains damage and healing from corpses and friendly naked nest serpents

Gboster-0:
## About The Pull Request

Fixes the vermillion cross gaining damage from corpses
And balances it to no longer gain damage from friendly naked nest serpents

It was always meant to be a good crowd control weapon, so everything else is fair game

## Why It's Good For The Game

Because its superior to all other weapons, a verbillion damage and healing is probably not good

## Changelog
:cl:
fix: Vermillion cross no longer gets its abilities triggered by dead corpses
balance: Vermillion cross no longer gets its abilities triggered by friendly naked nest serpents
/:cl:


Kitsunemitsu (approved):
Minor bugfix, definitely something that I want to add.

PR #2111, opened 2024-05-03
Allows for randomized gamemodes in the future

Kitsunemitsu:
## About The Pull Request
Adds in the ability to add more gamemodes in the future, on a random rotation.

## Why It's Good For The Game
We can now more easily add in random gamemodes.

## Changelog
:cl:
add: Added randomized abnormality sets
code: Moves joke abnos to their own folder.
/:cl:

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -249,16 +249,12 @@ SUBSYSTEM_DEF(ticker)
 	if(SSmaptype.maptype in SSmaptype.combatmaps)
 		if(!(istype(mode, /datum/game_mode/combat)))
 			mode = new /datum/game_mode/combat
-
-//	Among Us.
-	else if(SSmaptype.maptype == "skeld")
-	//	mode = new /datum/game_mode/traitor
-		toggle_ooc(FALSE) // Turn it off
-		CONFIG_SET(flag/norespawn, 1)
-
 	else
-		if(!istype(mode, /datum/game_mode/management))
-			mode = new /datum/game_mode/management/classic
+		var/choosingmap = pick(/datum/game_mode/management/classic, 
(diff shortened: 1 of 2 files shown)


StrangeUser:
So, let's take Artbook and Original abnos section for example. Game becomes almost... unplayable. Why? Well, it's simple answer. In terms of lategame, this pull going to suck. Since there's not much of a good abnos, and there's like only 3 alephs i can recall. It's Last Shot, DF, and GOD. There's only one zayin i believe? So called fallen amurdad, which is basically WORST zayin in the game. It's not going to refresh players, it's going to annoy them mostly. 
Another example - Limbus and Wonderlab. Two alephs. NI and Titania. How is that supposed to be played with? And considering lowpop on main LC13 mode and insane amount of ERA while only one agent persent, NI is going to be unworkable anomaly and guaranteed melt loss.
No LC abnos as well offers more painful gamestyle, but it's playable (without constant suffering), unlike those above. 
How is that considered refreshing, if you're just removing half of game content and enforce suffering to masses? 
There's 4 others options. All+jokes, LC only, all is going to be indeed balanced options.
So, in the end you have 3 out of 6 chances to get a proper game. Which is 50%. 2/6 if we dont consider only LC abnos, since this option seems very sad.
Bringing joke abnos and others from shelves is good thing to refresh game, while locking most of game content is NOT.

Kitsunemitsu:
That doesn't address the main concern that this PR will resolve.
I am looking at adding, removing and merging modes as we go on, however I believe that taking a butcher's knife to portion out the content will encourage less undesired behavior.

Please talk with the dev team about this PR, and why we are adding it.

Kitsunemitsu:
This PR now has no big gameplay changes to the players.

Coxswain-Navigator (approved):
Game mode stuff. Yippee!

PR #2113, opened 2024-05-04
Ports ghost runechat

Gboster-0:
## About The Pull Request

Ports the following:
- https://github.com/Monkestation/Monkestation2.0/pull/797 (only the ghost runechat part)

## Why It's Good For The Game

I think seeing ghost's messages is neat and helps the flow of a conversation
I do have a concern that it might be a bit too distracting but that would need to be tested in the serber.

image

## Changelog
:cl:
add: Ghosts can now speak to eachother in runechat
/:cl:


Kitsunemitsu (approved):
Approved, but if I get complaints we may have to make it a pref.

PR #2114, opened 2024-05-04
All-around Helper is now all-around weaker

Kitsunemitsu:
## About The Pull Request
- Grinder MK4's damage was reduced from 30 to 26.
- Grinder MK4's armor was changed from 5/-2/2/2 to 4/-1/2/2.
- Helper starts the spin attack 0.5 seconds faster
- Helper is now stunned for 3 seconds instead of 5.

## Why It's Good For The Game
Helper gives you way, way too much for what it is.
I want to curb it very slightly, without totally gutting everything.

## Changelog
:cl:
balance: Helper and it's EGO are now less desirable.
/:cl:


Coxswain-Navigator (approved):
It truly is a numbers game.

PR #2139, opened 2024-05-14
Tool Rebalance Project

Coxswain-Navigator:
## About The Pull Request
Improves the state of tool abnormalities regarding feasibility of use.

- Notes of a crazed researcher no longer expires until returned. Buff now scales slightly with the player's Temperance.
- Heart of aspiration's buff has gained stat scaling.
- Shelter from the 27th of March is now entered and exited significantly faster.
- Fixes a bug with Wishing well on maps where it sits too close to a wall
- Theonite Slab has had its attack changed significantly, and no longer performs attacks on players during work. In addition, **three of the stat-boosting tool abnormalities have had their penalties lowered by about a third.** The damage resistance penalty for each use has been reduced from .15 to .10 across the board, and from 0.7 to 0.6 for Theonite slab. At 10 uses this brings the maximum penalty for most permanent stat-boosting tools from 3.5x to about  2.5x.
- Luminous bracelet never made sense to me since you can just HIT YOURSELF TO KEEP YOUR HEALTH LOW! So now, it no longer cares about your health. Instead, luminous bracelet will now give you a warning at 5 minutes and kill you at 7.5 minutes of wearing. The healing has been lowered from 5 to 2 to compensate. In addition, instead of a fortitude bonus, the bracelet now lowers your max health by 40.
- Behavior adjustment receives a bit of scaling to its buff. In addition, the effect upon panic has been changed. Instead of causing death, the player's eyes are simply removed. In addition, the path has been changed to use the American spelling Behavior (vs Behaviour), the same used in Lobotomy Corporation.
- Dr. Jekyll's Tincture is an interesting tool but it's really hard to use. The ability has been changed to an at-will power, costing 50% of current sanity to activate. Having this abnormality active on your person grants a penalty of 30 maximum SP. Returning the tool while its ability is active will cause the user to instantly panic. In addition, the abnormality has had its risk level upped to WAW. The ability itself expires once the user reaches full sanity.
- Giant tree sap can now be dragged and moved around.
- Heart of aspiration goes through yet another rework. For starters, all of the work and sanity-related effects have been removed. Now, the tool is far simpler - it removes slowdown from pain and allows the players to ignore the downsides of being at critical health. As a downside, the tool causes a constant health drain starting at 50% HP. The health drain becomes significantly more intense at 25% hp, but it also provides a boost to justice at both of these HP thresholds.
- Third fate's loom now no longer requires using the tool to regenerate threads. Threads now regenerate every two minutes, and the tool can be returned at any time with no penalty. In addition, a visual element has been added to display the remaining number of threads.
- The time the player must wait before standing on You Must Be Happy has been lowered from 7 to 3 seconds.

## Why It's Good For The Game
Makes tools more fun.

## Changelog
:cl:
balance: significantly improved the usability of tool abnormalities
/:cl:


Kitsunemitsu:
I will be reading this shortly

code/modules/mob/living/simple_animal/abnormality/_tools/teth/behavior.dm
```diff
@@ -25,21 +25,25 @@
...
+	var/stat_penalty = 0
 
 /datum/status_effect/display/behavior/on_apply()
 	. = ..()
 	if(ishuman(owner))
 		var/mob/living/carbon/human/H = owner
-		H.adjust_attribute_buff(JUSTICE_ATTRIBUTE, 15)
-		H.adjust_attribute_buff(PRUDENCE_ATTRIBUTE, -10)
+		stat_bonus = (0.15 * get_attribute_level(owner, JUSTICE_ATTRIBUTE)) //15 + 15% of the user's justice is added as a bonus
+		stat_penalty = (0.1 * get_attribute_level(owner, PRUDENCE_ATTRIBUTE)) //10 + 10% of the user's temperance is added as a bonus
```
LanceSmites328, line 36:
```suggestion
		stat_penalty = (0.1 * get_attribute_level(owner, PRUDENCE_ATTRIBUTE)) //10 + 10% of the user's temperance is removed as a penalty
```

LanceSmites328 (dismissed):

Kitsunemitsu (approved):
Re-approving. All changes requested were done, and it's looking to be merged.

PR #2147, opened 2024-05-18
K-Corp SP ampules

ShinyPulsar:
## About The Pull Request
Added a K-Corp Blue Ampule, a Sanity counterpart to the HP green ampule. Obtainable through K-Corp crate in the common loot list, and through the Ampule runoff permit research in the K-Corp Representative console.

I KNOW apparently it's supposed to fit for M-Corp, but until the representative, crates, researches and all that are coded, it's good to have here. Then we can move the item to M-Corp when it's ready.

## Why It's Good For The Game
Allows quick SP healing just like HP ampules. Even though M-Corp isn't here yet, the code is at least ready and functionnal.

## Changelog
:cl:
add: Added K-Corp Blue Ampule
:cl:


Coxswain-Navigator:
Why would we add this to K-corp if it is just being removed later?

ShinyPulsar:
> Why would we add this to K-corp if it is just being removed later?

Not removed, just moved to M-Corp when it's coded.

Coxswain-Navigator:
You misunderstand. Players are going to expect to receive these from K-corp so long as it is a feature here. It will be interpreted as a removal if M-corp is ever added. And whether you agree or not, this does affect balance.

ShinyPulsar:
What do we do with that then ?

Coxswain-Navigator:
You could partially implement M-corp, or find another approach.

ShinyPulsar:
Isn't Kirie supposed to make M-Corp ? If I partially implement M-Corp I'll need more items and sprites.

Kitsunemitsu:
I would like to actually flesh out a few of our current reps, namely P-Corp, N-Corp and L-Corp before moving onto new ones.
Regarding this PR specifically, I think that K-Corp giving out SP healing is a little out of their ballpark.

We can revisit M-Corp in the future when we are ready.

PR #2154, opened 2024-05-21
No skip crimson ordeal

Halonexus:
## About The Pull Request
Normal crimson ordeals will split properly even when gibbed, timed spawn variants work the same as before.

## Why It's Good For The Game
Theres no reason why gibbing would prevent them from splitting, and it creates a weird over reliance on very particular ego to trivialise the ordeal.

## Changelog
:cl:
tweak: normal crimson ordeal mobs split when gibbed.
/:cl:


Coxswain-Navigator (approved):
I really like this, and while I was going to nerf the midnight clowns in some way to compensate for its difficulty, this was going to be part of the development anyways.

PR #2155, opened 2024-05-21
FIX : Main office rep lob points research

ShinyPulsar:
## About The Pull Request
Changes the message the manager's console gets when the research is done to the real amount of LOB points they get instead of what was originally intended. (Blame my forgetful ass for forgetting.)

## Why It's Good For The Game
So the poor manager doesn't gets confused.

## Changelog
:cl:
fix: Changed the message the Manager's console recieves when the research is done to it's real value.
:cl:


Kitsunemitsu (approved):
Approved. Looks good to me.

PR #2165, opened 2024-05-26
Apex predator can recloak

Halonexus:
## About The Pull Request
Allows apex predator to recloak after 30 seconds of not taking damage. Taking damage refreshes the timer.

## Why It's Good For The Game
The abnormality wont lose its fun gimmick for the rest of its life after taking any bit of damage anymore. Will make it more fun in RCA mode.

## Changelog
:cl:
tweak: Apex Predator can recloak after 30 seconds of not taking damage.
/:cl:


Kitsunemitsu (approved):
Sure. Looks good to me.

PR #2169, opened 2024-05-27
Ardor Blossom Moth

Kitsunemitsu:
## About The Pull Request
Adds Ardor blossom Moth.
A neutral abnormality that drops fire on the ground that burns you as you walk on it.
To save on the already intensive task of dropping the fire and refreshing them, They call Crossed() instead of checking if anyone's on the same tile of it every second.
Sprites by our lovely Mel <3

Also Decreases the amount of fire damage healed by regenerators. I was under the impression that it was much lower than the EXACT SAME as brute. This lets us use fire damage sparingly as slightly more permanent damage.

To do:
- [x] EGO weapon.

## Changelog
:cl:
add: Added Ardor Blossom Moth
/:cl:


InsightfulParasite:
May want to reduce the cooldown for restarting its patrol so it doesnt sit in a room letting its flames go to waste.

InsightfulParasite (approved):
Rebased and prepped for the server.

InsightfulParasite (dismissed):
Approved it before the rebase so it should be good now.

InsightfulParasite (approved):
Radical the checks are working again!

PR #2177, opened 2024-05-29
Fixes oversight in non-standard weapon charging and wayward having a useless attack_speed variable

Gboster-0:
## About The Pull Request

HandleCharge is supposed to universally handle charging, but it was checking specifically if it could charge via attacking and if not denying any charging at all.
My bad on the oversight

Also removes the waywards default attack speed because i just noticed it was like dat

## Why It's Good For The Game

Weapons working is gud

## Changelog
:cl:
fix: wayward passenger EGO now again charges when you move
/:cl:


Kitsunemitsu (approved):
Approved, it's a minor PR.

PR #2178, opened 2024-05-30
Adjustments and Balancing to Realizations

tile4:
## About The Pull Request
Balances Realizations, and adds a some ability additions/changes, 

- With Road to Gold, If you hold gold rush, the attack to an opponent after using the ability will massively lower the defenses of the enemy temporarily. This ability is based on the LoR page "The Road To Gold" that KoG uses which staggers the hit enemy. 
- Repentance no longer gives damage resistance when praying but instead does damage to you while praying (which will be healed anyways after you finish)
- Galaxy Child's Realization now actually should have the pebble break properly if the person who casts the ability dies.
- Adjusted Prophet and Combusting Courage to have the lower armor values akin to other realizations.

## Why It's Good For The Game
Realizations will be more desirable but not too strong. Also this should encourage more variety in what people realize. 

## Changelog
:cl:
add: Added new things
balance: rebalanced something
/:cl:


Coxswain-Navigator (approved):
I don't see any issue with these changes.

PR #2186, opened 2024-06-01
Removes the corpses inside of lockers and crates breathing

Gboster-0:
## About The Pull Request

Makes it so the closets now do not lock themselfes down if the person put inside of them is a corpse
This allows trash carts to again be used to transport corpses... althrough i will say you could just use a body bag for dat

## Why It's Good For The Game

Its what people have come to expect, and tbh its probably better
It does allow people to now again hide corpses from corpse loving abnormalities in lockers and such, but a lot of abnos don't benefit from interacting with them anyways

## Changelog
:cl:
fix: Trash carts now can again transport corpses
/:cl:

### DIFF
--- a/ModularTegustation/hiding/closet.dm
+++ b/ModularTegustation/hiding/closet.dm
@@ -24,16 +24,30 @@
 
 /obj/structure/closet/take_contents()
 	. = ..()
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (approved):
Loosk good to me.

PR #2188, opened 2024-06-03
Dummy fix and quenched blood fix

Halonexus:
## About The Pull Request
Debug dummy should work like normal hostile mobs do now so it will be affected by most things properly. Now applies damage coefficients to damage taken.
Also fixes quenched blood realized ability not dealing the promised 5% damage.
Made damage effect only check for the coefficients so that instances of 0 damage wouldnt trigger it when the mob isnt immune to that damage type.

## Why It's Good For The Game
More accurate dps dummy.
Ability works as it should.

## Changelog
:cl:
fix: fixed debug dummy, and quenched blood rapiers
code: damage effect only checks damage coefficients
/:cl:


Coxswain-Navigator (approved):
Some much-needed improvements in these trying times.

PR #2190, opened 2024-06-04
THE RETURN OF THE DEEP FRYER

Coxswain-Navigator:
## About The Pull Request
Adds fun frying, as part of the new Config Food verb for admin use. This is set up so that most non-food items cannot normally be deep fried without admin intervention. Logically, this means that deep fryers can be put back in most maps.

## Why It's Good For The Game
Removal stinks, things should be fixed instead!

## Changelog
:cl:
add: Added back deep frying
/:cl:


LanceSmites328 (approved):
Code looks solid, and if we have issues later where something important snuck past we can fix it.

PR #2208, opened 2024-06-09
Golden Apple RCA adjustments

vampirebat74:
## About The Pull Request

Golden Apple can now gain pulses from attacking a mech, provided that a pilot is inside it. Golden apple also no longer gains pulses from attacking Pinocchio carbons.

## Why It's Good For The Game

Fixing bugs is generally good for the game

## Changelog
:cl:
tweak: golden apple RCA tweak
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/golden_false_apple.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/golden_false_apple.dm
@@ -293,6 +293,17 @@
 				var/mob/living/hit = target
 				if((hit.stat == DEAD) ||!ishuman(hit))//if the target is dead or not human
 					return ..()
+				if(istype(target, /mob/living/carbon/human/species/pinocchio))
+					return ..()
+				pulse_count += 1
+			if(ismecha(target))
+				var/inhabited = FALSE
+				for(var/mob/living/L in target.contents)
+					if(L.stat == DEAD)
+						continue
+					inhabited = TRUE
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/abnormality/he/golden_false_apple.dm
```diff
@@ -293,6 +293,17 @@
 				var/mob/living/hit = target
 				if((hit.stat == DEAD) ||!ishuman(hit))//if the target is dead or not human
 					return ..()
+				if(istype(target, /mob/living/carbon/human/species/pinocchio))
```
InsightfulParasite, line 296: You could check based on faction but it might be nitpicky. Pinocchio works fine since we dont have that many carbon abnormalities.

vampirebat74: These only checks matter for a playable golden apple, since the hostile version doesn't use these pulses. Currently, there are no non-human opponents that would be encountered in RCA, so I don't think the faction check is necessary. These checks are run every single time golden apple performs an attack, so I would like to minimize unnecessary checks on it.

InsightfulParasite: Ah fair enough

InsightfulParasite (approved):
Looks good to me. If this has been tested on local then it should be good.

PR #2215, opened 2024-06-11
Apple Vines now can Grow On Floor Vents

InsightfulParasite:
## About The Pull Request
Snow Whites apple couldnt grow vines on J-777's floor vents due to her checking atmos. So now it checks reachable turfs instead. I had to additionally add a check to see if there is a door since without that they grow under doors which i felt was unfair.

If there is any noticeable lag let me know.

## Changelog
:cl:
fix: snow whites apple vines.
/:cl:

### DIFF
--- a/ModularTegustation/lc13_structures.dm
+++ b/ModularTegustation/lc13_structures.dm
@@ -37,11 +37,15 @@
 		qdel(src)
 		return FALSE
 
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (approved):
Easy approval. Looks good to me.

PR #2219, opened 2024-06-12
Fairy and Steel Runtime Fixes

InsightfulParasite:
## About The Pull Request
Fixes a runtime with Titanias fairies which are being spawned ontop of a gibbed corpse after it has been gibbed. Now they register the location of the body before it is gibbed then gib it then spawn.

Steel flying knockback has been given a return condition if the thing being thrown has been qdeleted.

## Changelog
:cl:
fix: Titania fairy spawn runtime
fix: Steel knockback runtime.
/:cl:


Kitsunemitsu (approved):
Approved. Looks good to me.

PR #2226, opened 2024-06-16
Removes sloth from LCL asset protection spawns

KingSuperBOB:
## About The Pull Request
Refer to title

## Why It's Good For The Game
Sloth did not grant an opportunity for RP, the ap ability grants players permanently dead the ability to re-enter the game and RP. Sloth would instead knockdown some folks, maybe kill 1 person and do nothing. Sloth is not a controllable mob as it moves on its own so the player would simply say a few lines then die..

## Changelog
:cl:
del: Removed Sloth from asset protection Pecca spawn
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/labs/command/assetprot.dm
+++ b/code/modules/jobs/job_types/labs/command/assetprot.dm
@@ -42,7 +42,6 @@
 	set category = "Gamemaster"
 
 	var/list/peccetulum = list(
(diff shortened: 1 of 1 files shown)


tile4 (approved):
Honestly, a fair enough removal for now, if we need to, we can work something else in or maybe add more peccatulum after the gold rework.

Kitsunemitsu (approved):
Sure. Approved. I get it.

PR #2227, opened 2024-06-16
rewrites the TSA in the woods adventure

vampirebat74:
## About The Pull Request

What it says on the tin

Tempy(temperancetempy) had mentioned that this event was very similar to the Laundry of Dreams from Distortion Detective, and I agree. This PR re-writes the lines for the failure event.

## Why It's Good For The Game

This abnormality and event was based on a dream I once had, of which this outcome was not explored. When writing this event, I simply wrote the details that came to mind at the moment, and I believe this would be an improvement.

## Changelog
:cl:
tweak: EO adventure writing TSA_woods
/:cl:


Kitsunemitsu (approved):
Approved, writing PR

PR #2231, opened 2024-06-16
Removes Abno Chem

InsightfulParasite:
## About The Pull Request
Abno Chem is already facing a rework so its a good time to clean house and get rid of the incomplete code we already have. Originally this was envisioned as one unique chemical for each abnormality. What was not considered at the time was that this would be 150+ unique chems with their own unique effects which is absurd and if actually done would result in chemicals that the clerks would actually get murdered for wasting PE on. So I'm wiping the slate clean.

The new rework that tempy is currently doing is going to be 7 sin chemicals that apply to each abnormality based on its assumed sin. This will take quite a while since making 7 chems that have unique effects that are unique and also equal in usefulness to each other will take a long time. I guess the implied gameplay of this is for clerks to just milk certain abnormalities for the sin they want even though it is not up to the clerks what sins they get and instead up to the agents who fully understand abnormalities for understanding. The general flaw of abno chem is its inaccessibility and the fact that the starter crate in the extraction cargo does not have all the tools needed for abnormality fluid extraction. You in fact need both the crate and the spout which will cost around 700 PE. The effects we already had to abnormality reagents were laughable at best such as how bald was genuinely useless since it just make the drinker bald. Wellcheers zero was too unpredictable to be useful.

Abno chem never really looked at what the average player would have wanted as a temporary buff nor does it have the foundation for people to quickly swig or inject chemicals before or during combat.

## Changelog
:cl:
remove: abno chem
/:cl:


InsightfulParasite:
Im sorry this was from a inappropriate mixture of frustration and anxiety.

PR #2232, opened 2024-06-17
Lets admins turn people into distortions

Coxswain-Navigator:
## About The Pull Request
An expansion of the mob distortion system.
Adds a smite that transforms a mob into a distortion, with a couple parameters. Also adds an admin proc that turns every mob into a random distortion instantly.

Also contains some outstanding new distortion code for handling the distortion of a human being.
1. Any mob can now be converted to a distortion, which is called by the distort smite. This allows for the creation of black boxes, and natural(?) distortions by future coders.
2. Humans who become a distortion now stay as the same person when unmanifesting.
3. Humans who unmanifest from being a distortion now gain the minimum attributes required to use their new E.G.O.

## Why It's Good For The Game
This will absolutely be abused by admins.

## Changelog
:cl:
add: Added distortion for admins
/:cl:


code/modules/admin/smites/distort.dm
```diff
@@ -0,0 +1,34 @@
...
+	var/instant = FALSE
+	var/forced = FALSE
+	var/mob/living/simple_animal/hostile/distortion/chosen_distortion = null
+	if(istype(target, /mob/living/simple_animal/hostile/distortion))
+		switch(tgui_alert(user,"Target is already a distortion. Force distortion again?","GET DISTORTED",list("Yes", "No"), 0))
+			if("Yes")
+				forced = TRUE
+			if("No")
+				return
+	switch(tgui_alert(user,"Choose distortion?","GET DISTORTED",list("Yes", "No"), 0))
```
Gboster-0, line 22: could be turned into an IF check, since we do effectivelly nothing on "no" anyways
```suggestion
	if(tgui_alert(user,"Choose distortion?","GET DISTORTED",list("Yes", "No"), 0)) == "Yes")
		chosen_distortion = tgui_input_list(user,"Which one?","Select a mob", subtypesof(/mob/living/simple_animal/hostile/distortion))
```

code/modules/admin/smites/distort.dm
```diff
@@ -0,0 +1,34 @@
...
+		if("Yes")
+			chosen_distortion = tgui_input_list(user,"Which one?","Select a mob", subtypesof(/mob/living/simple_animal/hostile/distortion))
+		if("No")
+			chosen_distortion = null //randomly picked later
+	switch(tgui_alert(user,"Instant distortion?","GET DISTORTED",list("Yes", "No"), 0))
```
Gboster-0, line 27: same here as above, since instant is FALSE on default in the same way that chosen_distortion is null by default

code/modules/mob/living/simple_animal/distortion/_distortion.dm
```diff
@@ -170,22 +189,84 @@
...
-	return
+			return
 
 /mob/living/simple_animal/hostile/distortion/proc/PostUnmanifest(mob/living/carbon/human/egoist)
 	return
 
+//Turn into a distortion
+/mob/proc/BecomeDistortion(mob/living/simple_animal/hostile/distortion/chosenmob = null, instant = FALSE, forced = FALSE)
+	var/mob/themob = null
+	var/list/message_list = list(
```
Gboster-0, line 216: items in a multi-line lists should be elevated one indentation up, so instead of
```
var/list/a_list = list(
item,
item2,
)
```
it should be
```
var/list/a_list = list(
	item,
	item2,
)
```

code/modules/mob/living/simple_animal/distortion/_distortion.dm
```diff
@@ -170,22 +189,84 @@
...
+	C.screen += T
+	animate(T, alpha = 255, time = 10)
+	var/display_text = text
+	T.maptext = "<br><span style=\"[style]\">[display_text]</span><br>"
+	addtimer(CALLBACK(GLOBAL_PROC, GLOBAL_PROC_REF(fade_blurb), C, T, 25), 40) //fade_blurb qdels the object
+
+/mob/living/simple_animal/hostile/distortion/BecomeDistortion(mob/living/simple_animal/hostile/distortion/chosenmob = null, instant = FALSE, forced = FALSE)
+	if(!forced)
+		return FALSE //Already a distortion
+	. = ..()
```
Gboster-0, line 268: quite sure you can replace this with a `return ..()`, but should be checked locally before pushing since it rarelly breaks stuff
```suggestion
	return ..()
```

Gboster-0 (commented):
le review, not too bad but could use changes

Gboster-0 (approved):
Was TM'd for a long time and now its basically as good as it can be code quality wise

PR #2233, opened 2024-06-17
Fluffy Tongue Removal

tile4:
## About The Pull Request
Removes all the Fluffy Tongue related code.

## Why It's Good For The Game
It isn't exactly a good thing when almost every person who uses it doesn't play the quirk's lore correctly, and the fact some fluffy tongues take this role yet use a command role or other role with loud speak just isn't healthy enough, and would cause WAY less conflicts between players if it was outright removed instead....

## Changelog
:cl:
del: Removed old things
/:cl:


Gboster-0:
Yeaaa, the better solution imo would be making it so taking the quirk prevents you from taking any command roles.
But at this point it went too far and removing it is well warranted

Kitsunemitsu:
Assigning to @InsightfulParasite for practice.

PR #2247, opened 2024-06-20
Fixes bottle status effect

Gboster-0:
## About The Pull Request

Turns bottle into a proper stacking status effect as its supposed to be
tested locally, should work just fine

## Why It's Good For The Game

Because players accidentally getting themselfes permanent -40 stats on a shift is not gud

## Changelog
:cl:
fix: Bottle's status effect now correctly stacks, and does not leave you with permanent negative stats if you take it more than 1 time per status effect
/:cl:


InsightfulParasite (approved):
Looks good. If tested locally and works then ill trust you.

PR #2251, opened 2024-06-21
I got pissed at inconsistent tool behavior, so i decided to make the warnings more visible -- A salt PR

Gboster-0:
## About The Pull Request

This is a salt PR technically
It does the following:
- All tools (bracelet and researchers notes) now use the span "hypnophrase" for their near-death warnings, making them much easier to notice, heart specifically also had another message added thats an exclamation mark with span_narsie
- Bracelet now gives a heart sound effect when you are nearly dead, as while researchers note spams you, heart does not.
- Made lists on tools more consistent, and in-line with the list standards
- Removed some unnecessary indentation

## Why It's Good For The Game

> All tools (bracelet and researchers notes) now use the span "hypnophrase" for their near-death warnings, making them much easier to notice
- The warnings are supposed to be VISIBLE, yet they (specifically bracelet) are basically impossible to differentiate from combat messages when you are fighting abnormalities. That is not gud

> Made lists on tools more consistent, and in-line with the list standards
- They already should be up to standards

> Removed some unnecessary indentation
- Its easier to read the code that way, really

## Changelog
:cl:
tweak: Tool near-death warnings are now more visible to the user
tweak: The heart of aspiration will no longer damage you if you are in hard-crit for 30 seconds
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_tools/zayin/wishwell.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_tools/zayin/wishwell.dm
@@ -181,9 +181,9 @@
 
 //This proc removes the need to copypaste every single armor and weapon into a list.
 /obj/structure/toolabnormality/wishwell/Initialize()
-	return ..()
+	. = ..()
 
-//Sorts them into their lists
+	//Sorts them into their lists
 	for(var/path in subtypesof(/datum/ego_datum))
 		var/datum/ego_datum/ego = path
 		switch(initial(ego.cost))
--- a/code/modules/mob/living/simple_animal/abnormality/_tools/zayin/wishwell.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_tools/zayin/wishwell.dm
@@ -8,16 +8,17 @@
 
 	ego_list = list(
 		/datum/ego_datum/weapon/bucket,
-		/datum/ego_datum/armor/bucket
-		)
+		/datum/ego_datum/armor/bucket,
+	)
 
 //loot lists
 	var/list/superEGO = list( //do NOT put this in the loot lists ever. SuperEGO is for inputs only so people can throw away twilight for no good reason.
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/simple_animal/abnormality/_tools/zayin/wishwell.dm
```diff
@@ -297,41 +309,44 @@
...
-	if(M != user)
-		to_chat(user, span_warning("You start pulling [M] into the well."))
-		if(do_after(user, 7 SECONDS, target = M)) //If you're going to throw someone else, they have to be dead first.
-			if(M.stat == DEAD)
-				to_chat(user, span_notice("You throw [M] in the well!"))
-				buckle_mob(M, check_loc = check_loc)
-			else
-				to_chat(user, span_warning("How could you be so cruel? [M] is still alive!"))
-		return
+	if(M == user)
```
InsightfulParasite, line 331: Does this file change the functioning of wishwell or is it just formatting correction?

Gboster-0: Actually, yes it does change it very minorly.
Whilst its mostly a formatting change i moved the `if(M.stat != DEAD) {return}` check from the do_after to be in front of the do_after. So instead of trying to stuff a living person for 7 seconds into the well and failing, you get told instantly that they are alive and you should probably start doing CPR you blind fool

InsightfulParasite: Okay thats a thumbs up from me.

InsightfulParasite:
image
Bit big but it does its job. The heart thing works because its a heartbeat but the luminous bracelet having a big ! may be a bit too much.

InsightfulParasite (approved):
Tested locally and the code looks good.

PR #2255, opened 2024-06-22
Fixes hurtinturf hitting objects twice

Halonexus:
## About The Pull Request
It was dealing damage to objects twice which caused taking damage after deletion runtimes.
Also added tiny fans and request consoles to the exclude list as those are decorative objects.

## Why It's Good For The Game
Fixes thing.

## Changelog
:cl:
fix: fixed hurtinturf damaging objects twice
/:cl:


InsightfulParasite (approved):
Looks good. Also strange the take_damage proc was there twice.

PR #2260, opened 2024-06-25
Fixes my silly little accident

Litberries:
## About The Pull Request

SO i forgot to do the math for my minigun and it ended up being ALEPH+ give or take, so I'm gonna unfuck that now

## Changelog
:cl:
balance: rebalanced the shitgun to be normal WAW-ish, like 40-90 DPS, spread making it hard to check though

/:cl:


Coxswain-Navigator (approved):
Should probably be fine but for real this time.

PR #2262, opened 2024-06-26
Makes a parent for Snake oil/Skin prophecy/Viva la voce/Theonite slab. They can always and only be used 10 times now

Gboster-0:
## About The Pull Request

- They are really similar in how they function, so made a parent for them. Now you can legally say "Viva la voce's daddy"
- The tool abnos can only now be used 10 times, regardless of the bonus attributes
- Bracelet no longer gibs people if they have oxygen damage, instead it deals 75% of their current health
- Tools will no longer allow you to do a do_after bar to tell you "thou cannot use this tool". They do it before instead

## Why It's Good For The Game

>They are really similar in how they function, so made a parent for them.
- Less copy-paste is gud

>The tool abnos can only now be used 10 times, regardless of the bonus attributes
- Training suppression or skin prophet, need i say more?
- Actually, yeah i do. You could have even gotten a minus in one of your attributes via a gift and used them again...

>Bracelet no longer gibs people if they have oxygen damage, instead it deals 75% of their current health
- Anemiacs will have oxygen damage 95% of the round, and lets be honest no one will use blood bags to fix it. So lets deal damage to them instead, a lot of it

> Tools will no longer allow you to do a do_after bar to tell you "thou cannot use this tool". They do it before instead
- Feels a bit weird to make a player do an interaction just to refuse working

## Changelog
:cl:
balance: Tool abnormalities can now only be used 10 times, no more. Regardless of your bonus
balance: bracelet no longer kills anemiacs
balance: bracelet now properly takes 40 max health, instead of giving it
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_tools/basic_attribute_giver.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_tools/basic_attribute_giver.dm
@@ -1,0 +1,36 @@
+/obj/structure/toolabnormality/attribute_giver
+	name = "all-natural snake oil"
+	desc = "A purported panacea that will supposedly treat anything from minor scratches to Alzheimer's."
+	icon_state = "snake_oil"
+	/// Users that used this tool
+	var/list/mob/captured_users = list()
+	/// The maximum boost that users can get from using this tool
+	var/max_boost = 100
+	/// The attribute thats given to the user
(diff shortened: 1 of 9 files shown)


Coxswain-Navigator (approved):
This is a nice change overall. I don't see any more outstanding issues, so I'll give this a merge.

PR #2272, opened 2024-07-01
Brute is not a real type of armor, looks like. Public is shocked!

Gboster-0:
## About The Pull Request

Thought brute was a real armor type, but nope!
This was causing runtimes any time a person was being checked for BRUTE damage (you must become strong), not optimal

## Why It's Good For The Game

no moar runtimes

## Changelog
:cl:
fix: Putting your hands in You must become strong no longer causes runtimes
/:cl:


Coxswain-Navigator (approved):
How could I say no to a runtime fix?

PR #2273, opened 2024-07-01
Removes Attack Speed Variable = 1 Runtime[DONE]

InsightfulParasite:
## About The Pull Request
A runtime is occuring in line 110 of _ego_weapon.dm due to the inclusion of a unused attack_speed var.

Also changes default attack_speed from null to 1

The purpose of this crash message was because attack speed = 1 was redundant and attack speed = 0 defaulted to 1 anyways. However making this crash message or altering the resto f the code to remove 1 would be difficult since alot of attack_speed buffs and alterations are made on the assumption that the weapon has attack_speed = 1. Maybe in the future ill replace the attack speed alterations with a proc that defaults 0 to 1 but that would be alot.

The reason that so many shields have their attack_speed overrided is because the default attack speed of the shield subtype is 3.

## Changelog
:cl:

tweak: _ego_weapon.dm
fix: _ego_weapon.dm

/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/_ego_weapon.dm
+++ b/code/game/objects/items/ego_weapons/_ego_weapon.dm
@@ -9,7 +9,7 @@
 	slot_flags = ITEM_SLOT_BELT
 	drag_slowdown = 1
 	var/list/attribute_requirements = list()
-	var/attack_speed
+	var/attack_speed = 1
 	var/special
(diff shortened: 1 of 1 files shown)


Gboster-0 (approved):
looks alright, merge when ya think its right

Kitsunemitsu (approved):
Approved. Looks good to me.

PR #2278, opened 2024-07-04
Makes more abnormalities ignore other Z levels for checks

tile4:
## About The Pull Request
Makes the rest of the death counter abnormalities that count deaths from other z levels, don't count those deaths. Also makes Hammer require the breached abnos to be on the same z level.

## Why It's Good For The Game
Bugfixes are good for the game

## Changelog
:cl:
fix: Z Level Checks
/:cl:


InsightfulParasite (approved):
Looks safe enough.

PR #2285, opened 2024-07-05
Pushes back the closet hiding variables to just open turfs

Gboster-0:
## About The Pull Request

Pushes back hiding variables into open turfs, rather than just floors

## Why It's Good For The Game

One less runtime to think about

## Changelog
:cl:
fix: Trash carts can no longer eat items
/:cl:


Coxswain-Navigator (approved):
This fixes a runtime, good.

PR #2291, opened 2024-07-08
Evangelion Elevator Scene (manager elevator now takes 53 seconds)

Kitsunemitsu:
## About The Pull Request
The Manager Elevator now takes 53 seconds instead of 1.
This is the exact amount of time as the Neon Genesis Evangelion elevator scene.

Optional To-Do
- [ ] Make Elevator music from scratch

## Why It's Good For The Game
People have been using the manager floor for immunity since time immemorial.
While it does not bother me, I am the only person that has the ability to do this without taking like a whole day (It only took me 3 hrs)

This serves a few things:
1) Less clerks want to go to the manager's floor, hopefully pushing them to instead use shelter
2) When people still want to go to the manager's floor, they still can it will just take fuckign forever.

Hopefully, players won't be running up and down like fools becuase they odn't want to take the elevator except when necessary.

## Changelog
:cl:
balance: The Elevator now takes 53 seconds.
/:cl:

### DIFF
--- a/_maps/shuttles/manager_elevator.dmm
+++ b/_maps/shuttles/manager_elevator.dmm
@@ -16,10 +16,10 @@
 /area/shuttle/mining)
 "m" = (
 /obj/machinery/computer/shuttle/mining{
+	dir = 4;
 	name = "elevator console";
-	shuttleId = "manager";
 	possible_destinations = "elevator_home;elevator_away";
-	dir = 4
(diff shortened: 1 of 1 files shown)


Kitsunemitsu:
I don't really like how this PR interacts with Seph and Rep; it has lowered Roleplay interactions across the board.

PR #2294, opened 2024-07-10
Paycheck Code

InsightfulParasite:
## About The Pull Request
Using the economy system i made a proc for paying employees on a 20min interval. Honestly id just start from scratch rather than using this EXTREMELY bloated economy system.

## Why It's Good For The Game
Requested by my fellow maintainers. Kind of unsure about this code since when i asked "Why" i was told that money wasnt used for much of anything currently.

Its proof of concept that i can make a payroll system. Though implimenting this now would make the money system worse because money has even less of a purpose in Lobotomy Corp. The real Currency is PE boxes in the Lobotomy Corp subsystem since that is what buys items and equipment from cargo. Money from ID cards is a holdover from SS13. 

## Changelog
:cl:
add: paychecks
/:cl:


Coxswain-Navigator:
The "Why" is pretty much so that we can put an end to rewarding players with ahn for some of the random jobs they can do. Like being unlucky with a gacha. There are actually a few things I've proposed making players use their personal funds for, like life insurance just as an example. Think of this as a foundation.

InsightfulParasite:
> The "Why" is pretty much so that we can put an end to rewarding players with ahn for some of the random jobs they can do. Like being unlucky with a gacha. There are actually a few things I've proposed making players use their personal funds for, like life insurance just as an example. Think of this as a foundation.

Makes sense. Ill polish the code later when im more awake.

PR #2295, opened 2024-07-10
A beginning for customizable TGUI: Auxiliary console preference

Gboster-0:
## About The Pull Request

Don't you just hate it when you as a ghost want to see the managers upgrades but they use the old UI that dosen't allow ghost interaction?
Aren't you annoyed that you prefer normal UI over TGUI. But it resets everyone round?

If eighter are true. This PR is for you

Now the console being TGUI or not is a user preference thats settable in the UI itself
It carries over the rounds and immediatelly saves

## Why It's Good For The Game

Bit less work to do each round if you like UI rather than TGUI

## Changelog
:cl:
code: Auxiliary console TGUI is now per-player instead of being universal to all players
/:cl:


Kitsunemitsu (approved):
Approved. Looks excellent to me.

PR #2297, opened 2024-07-12
Drifting fox rework and bugfixes

Coxswain-Navigator:
## About The Pull Request
Fixes Basilisoup stealing Drifting Fox's information
Minor code improvement to space lady that I happened to find
"Finishes" Drifting fox, implementing the unimplemented features and giving the abnormality a balance pass.
- Reduced Drifting Fox's attack speed from 3 to 2 - drifting fox was dealing far more damage than its movement speed would imply. This should make it no longer extremely dangerous in comparison to similar abnormalities.
- Drifting fox now periodically summons umbrellas.
- The more umbrellas are alive, the faster drifting fox moves.
- Killing an umbrella will slow drifting fox down, and deal damage to it.
- The umbrellas attack with a telegraph and apply black weakness on hit.

Reworks the sunshower E.G.O. to be a bit more interesting, fixes some numbers on basilisoup E.G.O.

- Sunshower is now both a poise AND a combo weapon
- Unlike other combo weapons, sunshower starts out slow but attacks faster on subsequent combo hits. This makes it feel a bit clunky to use, to balance out its high damage output due to poise critical hits. However, as a consequence of attacking slower the crits now hit a lot harder. Attacks that are not part of a combo are punished more harshly due to dealing a lot less damage.

## Changelog
:cl:
balance: rebalanced something
fix: fixed a few things
/:cl:


code/modules/mob/living/simple_animal/abnormality/aleph/space_lady.dm
```diff
@@ -13,7 +13,6 @@
 	can_breach = TRUE
 	threat_level = ALEPH_LEVEL
 	start_qliphoth = 2
-	ranged = 1
```
InsightfulParasite, line 16: Is this a accidental removal or part of the fix?

Coxswain-Navigator: This part of the fix. Space lady  has ranged = TRUE a few lines below.

InsightfulParasite (approved):
Its alot of stuff in one PR so it will be difficult to keep track of all of it but if you tested it local and it works then it should be a good approval.

PR #2303, opened 2024-07-17
Adds abnormalities in a seperate tab on the orbit menu

Gboster-0:
## About The Pull Request

Makes abnos show up below the living section of the orbit menu in their own little corner
If the abnormality is contained, it will have a green color
if its not, then it will be red
if its a tutorial abno, it will just show up in NPC's because who gives a shit
image

## Why It's Good For The Game

Bit weird to have them be in the same sections as all the other cleanbots and such, this is much clearer

## Changelog
:cl:
add: Abnormality orbit menu
/:cl:


InsightfulParasite (approved):
Nice to have QOL for ghosts.

PR #2305, opened 2024-07-18
Un-Lava-ing Kappa Corp Map and Theta Map Fix 2.0

ScarSkit:
Slight map changes to theta and kappa

## About The Pull Request

Removed var-editted lava tiles that seem to be the root cause of the lag and crashes in Kappa Corp Map. Also, removed the meme deep fryer area filled with hidden syndicate turrets. Added chasms to the abyss maps as well as a few decorative objs.

## Why It's Good For The Game

H corp is called Abyss, it should have Abyss tiles. J corp Arena did not need lava tiles disguised as iron sand open turfs.

## Changelog
:cl:
add: A few new aesthetic things in H Corp Abyss
del: Removed all the lava tiles in J Corp
/:cl:


Kitsunemitsu (approved):
Approving. Should be an easy fix

PR #2306, opened 2024-07-18
Re-adds kitchen in Zeta Corp

ScarSkit:
Re-adds kitchen utilities from other maps to Zeta Corp (F Corp Island)

## About The Pull Request

Makes Dead pop or very low pop with no dedicated fixer less painful by having a kitchen with prestocked food.

## Why It's Good For The Game

Makes deadpop or lowpop less painful as players arent required to fish for food. Also makes bassilisoup less of a pain.

## Changelog
:cl:
add: Added kitchen utilities to zeta corp
delete: Removed fishing stuff from cc and relocate them to training since they were clogging up south of cc.
/:cl:

### DIFF
--- a/_maps/map_files/Zeta/zetacorp.dmm
+++ b/_maps/map_files/Zeta/zetacorp.dmm
@@ -58,13 +58,12 @@
 	},
 /area/facility_hallway/south)
 "an" = (
-/obj/machinery/grill,
-/obj/structure/railing{
-	desc = "Indestructible railing created in a vain attempt of fool-proofing the wing's workplace.";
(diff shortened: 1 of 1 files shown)


Kitsunemitsu:
There's always vending machine food, should you need it.

ScarSkit:
that is true, I added vending machines in that map for that purpose, but also, clerks who like cooking will be more satisfied with a full package kitchen that lets them make not only fish based food but everything else
also trying to feed ballisoup dead pop with only vending machines is a pain in the ass.
also some players hate having to rely on only junk food
and fishing takes a while to do

PR #2313, opened 2024-07-22
Minor change to ERA buff description

Litberries:
## About The Pull Request

Changes it from saying +20 stat buff to -20 stat requirement

## Why It's Good For The Game

Because for **_some_** reason, most ERAs think that they're STRONGER when wearing gacha gear, when they're really just wearing it with lower stats

## Changelog
:cl:

tweak: tweaked ERA buff thingamababa
/:cl:


Litberries:
I could also be wrong and they might literally get stats buffed, but from my ERA hours, it doesnt seem that way

Gboster-0:
Yeah no, looking at the code it is just the ability to wear them at lower stats, weird that it was with the wrong description for so long

Kitsunemitsu (approved):
Approved. Looks good to me.

PR #2316, opened 2024-07-23
Theta Corp map fix 3

ScarSkit:
Abyss tiles now have density to prevent out of bounds shenanigans

## About The Pull Request

Sets density of fake chasm tiles seen outside the facility to 1, this prevents lets say helper from charging out of the facility and into the void.

## Why It's Good For The Game

Map fix, and a pretty important one

## Changelog
:cl:
fix: fixed an map oversight where density was not set properly leading to clowns and other simplemobs from getting out of bounds
/:cl:


Coxswain-Navigator (approved):
Important fix, approved.

PR #2319, opened 2024-07-25
Fallen Amurdad Crops Changes

tile4:
## About The Pull Request
Makes some changes so Amurdad soil doesnt produce varients that don't have repeatable harvest, but just prevents repeat harvest on the soil, changes some abnochem chemicals to other chemicals, including one made specifically for 2 Amurdad plants were also buffed.

## Why It's Good For The Game
Abnochem is planned to be revamped, so this will prepare for it too. Also,

## Changelog
:cl:
add: Added a new reagent
tweak: tweaked how Amurdad prevents repeatable growth, as well as replacing the abnochem reagents with other ones.
balance: Improved some plants you could get
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/ambrosia.dm
+++ b/code/modules/hydroponics/grown/ambrosia.dm
@@ -61,7 +61,7 @@
 	plantname = "Ambrosia Gaia"
 	product = /obj/item/food/grown/ambrosia/gaia
 	mutatelist = list(/obj/item/seeds/ambrosia/deus)
-	reagents_add = list(/datum/reagent/consumable/nutriment = 0.06, /datum/reagent/consumable/nutriment/vitamin = 0.05)
+	reagents_add = list(/datum/reagent/consumable/nutriment = 0.06, /datum/reagent/consumable/nutriment/vitamin = 0.05, /datum/reagent/medicine/earthsblood = 0.05)
 	rarity = 30 //These are some pretty good plants right here
 	genes = list()
 	weed_rate = 4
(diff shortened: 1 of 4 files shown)


code/modules/hydroponics/grown/ambrosia.dm
```diff
@@ -61,7 +61,7 @@
 	plantname = "Ambrosia Gaia"
 	product = /obj/item/food/grown/ambrosia/gaia
 	mutatelist = list(/obj/item/seeds/ambrosia/deus)
-	reagents_add = list(/datum/reagent/consumable/nutriment = 0.06, /datum/reagent/consumable/nutriment/vitamin = 0.05)
+	reagents_add = list(/datum/reagent/consumable/nutriment = 0.06, /datum/reagent/consumable/nutriment/vitamin = 0.05, /datum/reagent/medicine/earthsblood = 0.05)
```
InsightfulParasite, line 64: This is the only thing im a bit iffy about but everything else looks good.

tile4: I figured it wouldn't be too strong as 1. It does brain damage which you need another person so you can deal with it, 2. It pacifies when while in the person's body, which means no suppressing abnormalities when healing up with it, and 3. Usually teamwork like this isn't used. I can't count the number of times agents don't use healing items that clerks set up for them. But if it is that strong, I could always adjust the values if needed.

InsightfulParasite (approved):
Its not too bad of a change. Approved.

InsightfulParasite (approved):
This seems harmless enough.

PR #2323, opened 2024-07-27
Disabling Atmospherics Subsystem Hotfix

InsightfulParasite:
## About The Pull Request
Air was still being removed from turf due to assume_air. Bit embaressing that i didnt notice this during testing.
Im not sure what i removed initially but my goal right now is to prevent air being removed from turfs while allowing it to be taken from internal tanks.

Testing went smoothly air is no longer removed from the enviorment from mob/life() or lungs. Internal tanks still function if we ever want to utilize those. If this wasnt a hotfix id refactor simple animal air breathing since simple mobs dont take oxygen but register how much is in the turf.

Since crit damage is connected to breathing i had to check that too and noticed it took a very long time to expire from crit damage.
After returning to a code before my PR i found that crit is just very slow.

## Changelog
:cl:
fix: carbon/life
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -65,11 +65,11 @@
 
 /mob/living/carbon/human/handle_mutations_and_radiation()
 	if(!dna || !dna.species.handle_mutations_and_radiation(src))
-		..()
+		return ..()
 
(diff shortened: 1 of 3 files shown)


Kitsunemitsu (approved):
Approving

PR #2332, opened 2024-07-29
Makes Quota Calculated From Only Alive People

tile4:
## About The Pull Request
Increases the quota requirement modifier per person, but only includes alive people in the calculation. That way quota can be calculated with more accurate numbers of who will be playing at the start.

## Why It's Good For The Game
No more rounds with like 3 people playing for the whole entire round and over 10k PE for quota.

## Changelog
:cl:
balance: rebalanced quota calculations
/:cl:


Kitsunemitsu (approved):
Approved.

PR #2343, opened 2024-07-30
Reduces Blood Refinery PE Cost

tile4:
## About The Pull Request
Makes the Blood Refinery cost 75 PE to use instead of 100 PE.

## Why It's Good For The Game
As is it is worse than the regular refinery in all possible ways, as it takes a good amount of time to heal and is more expensive than using the refinery with filters. With this change, it might make the blood refinery more worth it. 

## Changelog
:cl:
balance: rebalanced blood refinery
/:cl:


Gboster-0 (approved):
I do definetelly think this is good for the game, the blood refinery sucks in every single way.
Approving but not merging, i'll wait till another maintainer also agrees on this and merges it.

Kitsunemitsu:
I'm agreeing, the alt refineries pr had the blood refinery be way too good in it's original interation. 

It was overnerfed.

PR #2344, opened 2024-07-31
Optimizes most LC13 uses of Range to incorporate urange and livinginrange

Kitsunemitsu:
## About The Pull Request
Fits all uses of our range procs to their more optimal range procs.

## Why It's Good For The Game
livinginrange is much more intensive than urange or range for distances lower than 15, and urange is better than range for above 8.

This just makes it so that we have much cleaner code, and is a rather small change.

## Changelog
:cl:
code: Made most LC13 things use what I think is the most efficient range proc
/:cl:

### DIFF
--- a/code/datums/ai/sanity/_sanityloss_controller.dm
+++ b/code/datums/ai/sanity/_sanityloss_controller.dm
@@ -377,7 +377,7 @@
 		if(living_enemy.status_flags & GODMODE)
 			blackboard[BB_INSANE_CURRENT_ATTACK_TARGET] = null
 			return
-		if(!(living_enemy in livinginrange(10, living_pawn)))
+		if(!(living_enemy in range(10, living_pawn)))
 			blackboard[BB_INSANE_CURRENT_ATTACK_TARGET] = null
 			return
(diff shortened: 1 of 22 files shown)


InsightfulParasite:
If this is based on the amount of tiles checked shouldnt we have a unified proc that automates which it checks for?
like
if(tile_Range > 8)
return range(tile_Range)

Kitsunemitsu:
> If this is based on the amount of tiles checked shouldnt we have a unified proc that automates which it checks for? like if(tile_Range > 8) return range(tile_Range)

I would agree, but it is slower than manually calling it, and these sometimes run many times a second.

InsightfulParasite:
> > If this is based on the amount of tiles checked shouldnt we have a unified proc that automates which it checks for? like if(tile_Range > 8) return range(tile_Range)
> 
> I would agree, but it is slower than manually calling it, and these sometimes run many times a second.

Fair

InsightfulParasite:
Checking Urange functionality locally because i havent seen the urange proc before.

InsightfulParasite (approved):
Tested locally with urange and it seems to work. Approved.

PR #2346, opened 2024-07-31
[KAR] Fixes Quirk Runtime

LanceSmites328:
## About The Pull Request
Quirks through an empty string into `to_chat` if they had no Gain or Lose text. This is bad and causes a runtime. This PR simply adds an `if` statement checking for the text before trying to send it.

## Why It's Good For The Game
Fixes a runtime.

## Changelog
:cl:
fix: a runtime.
/:cl:


InsightfulParasite (approved):
This makes sense as a runtime fix.

PR #2351, opened 2024-08-02
[TM ONLY] Abno month abno disable list

Kitsunemitsu:
## About The Pull Request
Disables certain abnormalities for abnormality month.


Kitsunemitsu:
Did what it needed to do

PR #2354, opened 2024-08-02
Turns eta-corp wastefull var-edited turfs into actual coded ones

Gboster-0:
## About The Pull Request

It was wastefull to have so many turfs on the map be var-edited, not to mention some having the wrong description/not certain variables on them whilst others did

This PR cleans up duplicate lattices, wrong vars, and makes them into actual code

Also it adds windows where there previously were pixel shifted windows that you can walk though, because mappers

## Why It's Good For The Game

Its probably better for game performance, also storage, a lot


InsightfulParasite (approved):
Map fix

PR #2360, opened 2024-08-05
Makes the spoon and Mail satchel show its true damage on examine, instead of hiding it

Gboster-0:
## About The Pull Request

Spoon and the mail satchel now return to their initial force at the beginning of the `attack()` proc instead of at the end, letting people examine it to see how much damage its doing
Also removed a useless piece of code that set satchel to white damage each time it swung, because thats copy-pasta from the spoon code

## Why It's Good For The Game

I think its good to inform players that a weapon will deal X amount of damage, and not tell them only the base amount without unique multipliers

## Changelog
:cl:
tweak: The protagonist spoon and mail satchel EGO weapons can now be examined after hitting an enemy to see their true damage
/:cl:

### DIFF
--- a/code/game/objects/items/ego_weapons/special.dm
+++ b/code/game/objects/items/ego_weapons/special.dm
@@ -1,5 +1,14 @@
 //Abnormality rewards
 //Bottle of Tears
+/**
+ * Here's how it works. It scales with Fortitude.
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (approved):
Approved. Looks good to me, just moving shit around

PR #2362, opened 2024-08-06
Makes the cherry tree no longer stop you from working or hold your gained attributes hostage for 10 seconds upon breaching

Gboster-0:
## About The Pull Request

Gives the cherry blossom an async calling the proc instead of directly calling it
Cleans up some code to use `for(var/i in 1 to number of executions)` instead of `for(var/i=5,  i>=1, i--)`

## Why It's Good For The Game

Gives the cherry blossom an async calling the proc instead of directly calling it
> It just feels jank that you don't immediatelly get your stats upon completing its work, and have to wait for it to start cherrying
> This does mean you can immediatelly start working it again instead of waiting those dreadfull 10 seconds, so kinda a buff in that regard

Cleans up some code to use `for(var/i in 1 to number of executions)` instead of `for(var/i=5,  i>=1, i--)`
> Its a lot more clear to see what its doing, imo the format of the method thats being used rn is just... bad

## Changelog
:cl:
tweak: The Grave of Cherry Blossoms no longer makes you wait 10 seconds to give its attributes from work if it starts the blossom season
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/teth/cherry_blossoms.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/teth/cherry_blossoms.dm
@@ -39,13 +39,12 @@
 		There is beauty even in great and terrible things. <br>\
 		Even the bodies underneath this tree would agree with you."
 
-	var/numbermarked = 5
+	var/number_of_marks = 5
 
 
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (approved):
Approved.

PR #2366, opened 2024-08-06
Fixes runtime with metal fixer's projectiles calling the wrong arguments in a proc + cleanup

Gboster-0:
## About The Pull Request

A very small code cleanup
Also fixes a runtime because metal fixer's bullets were calling a proc that wanted 1 argument with 2 when hitting him, so tis was a lot of runtimes because of dat

## Why It's Good For The Game

Bugs bad

## Changelog
:cl:
code: Metal fixer no longer causes runtimes when he gets hit by his own bullets
/:cl:


LanceSmites328 (approved):
0ROOv

PR #2373, opened 2024-08-07
Airlock De Densification[WIP]

InsightfulParasite:
## About The Pull Request
Replaces the density variable in doors into closed_door variable. This seems to block all things including projectiles.
Go_to proc paths around structures that are dense. If a door is defined as not dense then AI will attempt to enter through the door rather than bang on the window.

TESTING CHECKLIST

- [x] Projectiles are blocked by closed doors and pass through open doors.
- [x] Monsters actually path around obstacles through doors.
- [ ] Bolted Doors cannot be opened.
- [ ] Welded doors cant be opened
- [ ] People standing on doors are not immune to effects.

## Why It's Good For The Game
AI can path through doors if the doors are not defined as dense.

## Changelog
:cl:
tweak: Door density is replaced with closed_door variable
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -109,7 +109,7 @@
 	var/obj/machinery/door/airlock/cyclelinkedairlock
 	var/shuttledocked = 0
 	var/delayed_close_requested = FALSE // TRUE means the door will automatically close the next time it's opened.
-	var/air_tight = FALSE	//TRUE means density will be set as soon as the door begins to close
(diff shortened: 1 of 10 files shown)


InsightfulParasite:
This has might more consequences than i expected. Removing dense from doors will result in some dashing effects going right through them. Ill have to figure out some other way to make monsters smarter.

InsightfulParasite:
This has might more consequences than i expected. Removing dense from doors will result in some dashing effects going right through them. Ill have to figure out some other way to make monsters smarter.

PR #2374, opened 2024-08-07
Fixes even more metal fixer runtimes

Gboster-0:
## About The Pull Request

i did not see that the projectiles also interacted with the metal statues the same wrong way as with the metal fixer in the original PR

## Why It's Good For The Game

Less runtimes


Coxswain-Navigator (approved):
This is a lot cleaner than the original code

PR #2375, opened 2024-08-07
Removes stat equalizers from ERA's/DO's, stats automatically get assigned when available instead

Gboster-0:
## About The Pull Request

ERA's / DO's now have their stats automatically adjusted to what they should be whenever enough abnos exist

## Why It's Good For The Game

I hate using the device in my hand every 30 seconds to see if i can have higher stats or not

## Changelog
:cl:
tweak: ERA's / DO's now automatically have their stats adjusted according to what they should be, the stat equalizer is now a training officer only tool
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/mentor/trainingofficer.dm
+++ b/code/modules/jobs/job_types/mentor/trainingofficer.dm
@@ -89,11 +89,7 @@
 		if(79 to 100) // ALEPHs around here (20 Abnormalities)
 			set_attribute *= 4
 
-	if(user?.mind?.assigned_role == "Records Agent") //only give these if you're the guy that's allowed to have this, it's here so the Rec Agent can use it on interns
-		set_attribute += GetFacilityUpgradeValue(UPGRADE_AGENT_STATS) + SSlobotomy_corp.ordeal_stats
-
-	else
-		return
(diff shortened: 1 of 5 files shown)


code/modules/jobs/job_types/mentor/trainingofficer.dm
```diff
@@ -39,26 +39,67 @@
...
+			set_attribute *= 3.5
+
+		if(79 to 100) // ALEPHs around here (20 Abnormalities)
+			set_attribute *= 4
+
+	if(user?.mind?.assigned_role == "Records Agent") //only give these if you're the guy that's allowed to have this, it's here so the Rec Agent can use it on interns
+		set_attribute += GetFacilityUpgradeValue(UPGRADE_AGENT_STATS) + SSlobotomy_corp.ordeal_stats
+
+	else
+		return
```
LanceSmites328, line 96: Reasonably you can't even reach here without being a "Records Agent" due to previous checks, but this check should be higher to prevent redundant operations. Also inverted as part of that to simply be
```
if(user?.mind?.assigned_role != "Records Agent")
	return
```

LanceSmites328 (changes requested):

LanceSmites328 (approved):

PR #2376, opened 2024-08-07
Fixes the training officer having a "Newbie Agent" alt-title

Gboster-0:
## About The Pull Request

Fixes the training officer having a "Newbie Agent" alt-title, shrimple as that
tis because on agent it has the alt title defined, and we didnt previous override it in the Training Officer role, whilst it is a subtype of agent

## Why It's Good For The Game

its weird that it can have that

## Changelog
:cl:
fix: Training officer no longer has the "Newbie Agent" alt-title
/:cl:


Coxswain-Navigator (approved):
Very clean, very good.

PR #2377, opened 2024-08-07
Makes abnormalities where you avoid goods give more understanding on Neutrals

tile4:
## About The Pull Request
For abnormalities that quiploth drop from goods or can't get goods realistically, this PR makes them able to get understandings goods would give on their normal works.

## Why It's Good For The Game
Convenience for higher pops as with these abnormalities, as well as this was a thing that was planned to be added in the past. 

## Changelog
:cl:
add: Added a new variable
balance: rebalanced good work haters understanding rates
/:cl:


Kitsunemitsu (approved):
Approving this. It looks like a great change.

PR #2378, opened 2024-08-08
Fixes memories temperance check

Gboster-0:
## About The Pull Request

Makes memories check of `usertemp >= 60` into `usertemp > 60`

## Why It's Good For The Game

Its a level 3 temperance check that fails if the user is at preciselly level 3, not great

## Changelog
:cl:
fix: Memories from a better time no longer fails the temperance check if the user is at preciselly 60 temperance
/:cl:


InsightfulParasite (approved):
As the initial creator of Better Memories i approve this message!

PR #2380, opened 2024-08-09
Improves Simple Smile's Work Rates

tile4:
## About The Pull Request
Changes Simple Smile's workrates to have an 85% base chance and decrease by 20% instead of 80% and 30% respectively

## Why It's Good For The Game
Simple Smile's workrates have been quite outdated throughout the changes we have gotten. With the slower pace of the game, it would be way less plausable for people to have 1 in all. Hopefully this will help make Simple smile better.

## Changelog
:cl:
balance: rebalanced Gone With a Simple Smile
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/teth/simple_smile.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/teth/simple_smile.dm
@@ -101,7 +101,7 @@
 
 	for(var/attribute in stats)
 		if(get_attribute_level(user, attribute)>= 40)
-			chance_modifier *= 0.85
+			chance_modifier *= 0.8
 			lucky_counter += 1
 
 	return chance * chance_modifier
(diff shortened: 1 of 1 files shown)


Coxswain-Navigator:
My only issue with this PR is that even the worst case scenario work rates become right about the average for an abnormality of this risk level. I guess its breach is annoying enough though.

tile4:
> My only issue with this PR is that even the worst case scenario work rates become right about the average for an abnormality of this risk level. I guess its breach is annoying enough though.

Actually, I took a look and I guess I misread what Smile's old values are. I will adjust that and have everything properly updated in info to be just .8 scaling instead of .85 because .7 is too awful overall.

Kitsunemitsu (approved):
Approving this minor change, we'll see what happens.

PR #2381, opened 2024-08-10
Soup code deshittification

Gboster-0:
## About The Pull Request

This was only meant to be a change where you could see its workrates with a work chance injector, god what has it become?

- Some variables had comments added saying what they did
- Redundant indentation was removed in a lot of places
- `Move()` no longer overrides its vars
- Basilsoup mob now stores the datum of soup
- The literally useless `SpawnConnectedStructure()` override was removed (it was literally the same as the proc)
- Code that handled work chances was moved from `AttemptWork()` to `WorkChance()`, making people see its real chances.
- Soup lists were made up to our standard
- Switches/if statements that checked from -INFINITY to 0, now only check 0.
- `attack_hand()` No longer puts 2 fingerprints of a person on itself
- You no longer get told that you need a person to be dead before trying to put them into soup
- Fixes `return . = ..()` because ???

## Why It's Good For The Game

- Some variables had comments added saying what they did
> I saw that some of them had comments, so those that did were moved to /// instead of // and had some wording changed to better describe the vars

- Redundant indentation was removed in a lot of places
- `Move()` no longer overrides its vars
- The literally useless `SpawnConnectedStructure()` override was removed (it was literally the same as the proc)
- Soup lists were made up to our standard
> muh code quality

- `attack_hand()` No longer puts 2 fingerprints of a person on itself
> The parent call already put a fingerprint onto the soup anyway, we don't need to do it twice

- Code that handled work chances was moved from `AttemptWork()` to `WorkChance()`, making people see its real chances.
> Consistency, KHz also does this and whilst im a bit icky about letting people with work chance injectors see that the soup is poisoned, i think it'll be fine

- Basilsoup mob now stores the datum of soup
> Should be better, since it reduces redundant `locate()` calls

- Switches/if statements that checked from -INFINITY to 0, now only check 0.
> We clamp soup levels earlier to be between 0 and 100, if it reaches less than 0 something is REALLY wrong

- You no longer get told that you need a person to be dead before trying to put them into soup
> Nobody likes to do a `do_after()` just to be told "yeah no, you dont meet the requirements :)"

## Changelog
:cl:
balance: Work chance injectors now show the true workrates on basilsoup
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/basilisoup.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/basilisoup.dm
@@ -108,8 +108,7 @@
 	if(!locate(/obj/structure/basilisoup_pot) in datum_reference.connected_structures)
 		connected_soup = SpawnConnectedStructure(/obj/structure/basilisoup_pot)
 
-	for(var/atom/movable/A in datum_reference.connected_structures)
-		A.forceMove(get_step(src, NORTH))
+	connected_soup.forceMove(get_step(src, NORTH))
 
 // Work Mechanics
 /mob/living/simple_animal/hostile/abnormality/basilisoup/AttemptWork(mob/living/carbon/human/user, work_type)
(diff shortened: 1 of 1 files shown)


Coxswain-Navigator (approved):
Tested this locally and didn't find any issues. Well, except if someone sets the soup level below zero. Nobody will ever do that. Probably.

PR #2382, opened 2024-08-11
Adds melee damage to guns

Coxswain-Navigator:
## About The Pull Request

MAJOR code refractors : 
- attack_speed has been moved up to the /item level, which means that guns can now have a set attack speed
- force_multiplier has been moved up to the /item level. There's pretty much no reason not to. This allows the melee damage of guns to be upgraded through various means.
- Things that previously only improved bullet damage multipliers for guns now also update the force multiplier.

Gun Changes :

- All ego guns now deal approximately 70% of the melee damage you would expect from a melee weapon of the same tier.
- All ego guns now deal a damage type matching the bullet they fire.
- Pistols have a 0.5 attack speed by default.
- All guns now have melee damage statistics as part of their examine text.
- All automatic weapons now check if you are able to attack before firing. This prevents people from firing automatic weapons while simultaneously bashing something with them.

Individual weapon changes :

- Banquet is now a medium sized gun, and has lifesteal on melee hits. It now only consumes health to reload rather than per shot.

## Why It's Good For The Game
Guns have to reload and players don't like being a useless wet noodle when that happens.

## Changelog
:cl:
tweak: all guns now have a melee damage type
balance: all guns now have actual melee damage numbers
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/extraction/egosurgery.dm
+++ b/ModularTegustation/tegu_items/extraction/egosurgery.dm
@@ -9,7 +9,7 @@
 	var/maximum_energy = 5
 	var/target_item = null
 	var/current_progress = 0
-	var/random_sound_list = list( //Random surgery sound for every step
+	var/random_sound_list = list( // Random surgery sound for every step
 	'sound/items/jaws_pry.ogg',
 	'sound/items/drill_use.ogg',
 	'sound/items/welder.ogg',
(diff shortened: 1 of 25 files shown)


Kitsunemitsu:
Also technically on freeze as it's a global weapon change and not pertaining to any specific abno

Gboster-0 (approved):
Been testmerged enough, can't find anything major in a review.

PR #2383, opened 2024-08-13
Snow Queen Arena Tweaks and Fixes[DONE]

InsightfulParasite:
## About The Pull Request
Fixes a issue where snow queen would teleport out of her if she melts down during a duel.
Adds some extra special text to the ice sword so people know they can deflect damage with it.
Added some additional red defense to the ice sword so it is more effective in arena combat.
Fixed a bug where the person trapped in ice would not lose the immobalization traits added by ice.
Adds sound effects for removing sword from the ice and hitting the ice block to melt it.
Due to issues with snow queen final observation the frost crown was changed to her default gift rather than her having no gift type.

## Changelog
:cl:
fix: Snow Queen breaching during a duel.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/snow_queen.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/snow_queen.dm
@@ -59,7 +59,7 @@
 		/datum/ego_datum/weapon/frostsplinter,
 		/datum/ego_datum/armor/frostsplinter,
 	)
-	gift_type = null
+	gift_type = /datum/ego_gifts/frostcrown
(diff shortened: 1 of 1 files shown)


Gboster-0:
Gonna put balance on this, since it does bump up the red protection on da ice queen sword from 40 to 80, making parrying much more effective

InsightfulParasite:
> Gonna put balance on this, since it does bump up the red protection on da ice queen sword from 40 to 80, making parrying much more effective

This is true i forgot i did that.

Kitsunemitsu (approved):
Approving. Was tested with one minor bug that was fixed after.

PR #2387, opened 2024-08-14
Makes bottle of tears no longer consider your temperance buff, when taking away your justice

Gboster-0:
## About The Pull Request

- Makes it so bottle takes in your temperance + buff in the check (what it has always done), and only your temperance without buffs on the actual justice buff part
- Removes the message i earlier inserted about becoming a part of an endless sea

## Why It's Good For The Game

- Makes it so bottle takes in your temperance + buff in the check (what it has always done), and only your temperance without buffs on the actual justice buff part
> Its supposed to leave you at 20 temperance in the end, if you had viva you instead get left at 0 and get a massive buff thats WAY over what it should be. Ending at a negative justice buff
- Removes the message i earlier inserted about becoming a part of an endless sea
> It was supposed to show that you failed the check, it does not do that clear enough. My bad

## Changelog
:cl:
fix: Bottle no longer looks at your temperance buff and goes "yeah, lets buff his justice 3 times more because he had viva"
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/zayin/bottle.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/zayin/bottle.dm
@@ -162,9 +162,9 @@
 	if(temperance >= (fortitude + prudence + justice) / 1.5) // If your temperance is at least twice your average stat, you aren't hurt, but lose temperance.
 		var/raw_temperance = get_raw_level(user, TEMPERANCE_ATTRIBUTE)
 		to_chat(user, span_userdanger("The room is filling with water... but you feel oddly unconcerned."))
-		user.adjust_attribute_level(TEMPERANCE_ATTRIBUTE, 20 - raw_temperance)
+		user.adjust_attribute_level(TEMPERANCE_ATTRIBUTE, 20 - floor(raw_temperance))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/abnormality/zayin/bottle.dm
```diff
@@ -176,22 +176,21 @@
 	user.AdjustSleeping(10 SECONDS)
 	if(user.stat == DEAD)
 		animate(user, alpha = 0, time = 2 SECONDS)
-		to_chat(user.client, span_userdanger("You have died from the bottle, and your body is now a part of the endless sea."))
```
Coxswain-Navigator, line 179: Why not replace it with anything if removing it?

Gboster-0: i just feel like its more consistent to have the user die without explanation, as happens with other abnormalities at the current moment.
When it was made i was more-so pivoting towards "let's tell them that they didn't have enough HP/Fort/Temp, and died due to that" but the message effectivelly served to do no explanation nor give flavor as i worded it wrongly.
If ya want i could still add the messages for all the different death reasons from bottle of tears tho

Coxswain-Navigator (approved):
Fixes attribute weirdness, and does exactly what it says on the tin.

PR #2393, opened 2024-08-18
Gives shep a simple second, faster attack

Kitsunemitsu:
## About The Pull Request
Gives shep a second, faster attack on the player's position. It's still dodgable

## Why It's Good For The Game
Gives shep some variance on if they attack you or aroudn it.

## Changelog
:cl:
balance: added a new attack to blue shep
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/blue_shepherd.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/blue_shepherd.dm
@@ -222,7 +222,10 @@
 		slash_current = slash_cooldown
 		say(pick(combat_lines))
 		slashing = TRUE
-		slash()
+		if(prob(50))
+			slash()
+		else
+			targetslash()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/abnormality/he/blue_shepherd.dm
```diff
@@ -287,13 +291,23 @@
 		return FALSE
 	return ..()
 
+/mob/living/simple_animal/hostile/abnormality/blue_shepherd/proc/targetslash()
```
Gboster-0, line 294: isn't the only different line the `addtimer()`?
If so, i dont really see a point making an entirelly different proc, and it should be a var in the proc instead like
```
/mob/living/simple_animal/hostile/abnormality/blue_shepherd/proc/slash(fast = FALSE)
```

PR #2395, opened 2024-08-20
Removes Box of L corp Debug tools from J Corp Lootbox pool

tile4:
## About The Pull Request
Removes a quite big oversight of J Corp Lootboxes and prevents the debug box itself from being impossible to get from it.

## Why It's Good For The Game
Some of the LC13 debug tools can break the games in multiple ways. This was intended but I forgot to include the box itself.

## Changelog
:cl:
fix: removed LC13 debug box from being possible to get from lootboxes
/:cl:


Gboster-0 (approved):
Imo i don't like the debug items being blacklisted because its funny to me if people get them.
But a needed change.

PR #2397, opened 2024-08-21
Allow Wcorp Vest to be Removed from Armor Fridges

InsightfulParasite:
## About The Pull Request
W corp vest was unable to be removed from the armor fridge because of the \improper in its name. The improper is removed for functionality.

In vv i went to edit the variable for name to see if that was the issue. 
Screenshot 2024-08-21 173355
the mproper was still in the name but i couldnt see it in the variable.

One day we will have smart fridges that work on item type rather than name.

## Changelog
:cl:
fix: W corp vest name.
/:cl:


Coxswain-Navigator (approved):
Yeah, sure. Fixes a bug.

PR #2408, opened 2024-09-02
Makes the shrimp tongue no longer change "krill" into "kreel"

Gboster-0:
## About The Pull Request

Makes kill get turned into "%1", that then gets turned into "krill"
This protects any "kill" sections from being turned into "keel", without using any if() statements or double-regexing

## Why It's Good For The Game

Its probably the best solution for this, that allows both "ill" and "kill" to be turned correctly in a single message
Also its annoying to see it actually replace krill

## Changelog
:cl:
fix: The shrimp tongue is now properly telling people to krill themselfes
/:cl:


Coxswain-Navigator (approved):
Nice fix to a major feature

PR #2411, opened 2024-09-08
Abnormality Portrait Refactor

LanceSmites328:
## About The Pull Request
Changes abnormality portrait definition from manual to automatic. Simply placing the file in the folder is now sufficient to access it via abnormality files. 
Also moves the proc from it's own file to Abnormality Helpers
Abnormality Portraits paths are now stored in a global list and not generated repeatedly
Abnormalities without portraits should now simply exclude abnos without a portrait as opposed to showing a missing image file.

## Why It's Good For The Game
Less work :+1:

## Changelog
:cl:
refactor: moves `get_abnormality_portraits()` into `create_portrait_paths()`
refactor: Abno portrait paths are now stored in a global list instead of searching the files repeatedly.
/:cl:


Coxswain-Navigator (approved):
Testmerged for a couple days, and nobody complained. No news is good news?

PR #2413, opened 2024-09-08
Reindeers no longer kill everyone when crossing beams

Kitsunemitsu:
## About The Pull Request
Reindeer beams now gib the user instead of calling explosion

## Why It's Good For The Game
Reindeer beam crossing has been the reason why R-Corp loses literally 90% of the time.

Almost every single loss from R-Corp has been because reindeer have killed 10+ people in a push.
While extremely funny, stupid reindeer players do lose the entire round singlehandedly.

## Changelog
:cl:
balance: Reindeers now just kill each other.
/:cl:


Coxswain-Navigator (approved):
YES! YES! A thousand times, yes!

PR #2422, opened 2024-09-12
Turns the records core code inside out to hopefully no longer make spiders cuddle with shrimps

Gboster-0:
## About The Pull Request

- Makes the records core swap the physical places of all the objects instead of their vars, as thats probably better in every way shape and form
- Abnormality datums now contain the camera and door they are using, specifically for records
- Makes basilsoup properly use `SpawnConnectedStructure` instead of forcemoving its structure immediatelly after spawning it
- Records will now always teleport at least 1 abno per meltdown, if it can

## Why It's Good For The Game

- Makes the records core swap the physical places of all the objects instead of their vars, as thats probably better in every way shape and form
> If in the future if new vars get added to doors or any of the objets, records will now no longer need to be updated alongside them
> This does raise a concern of the door's colors also being now teleported so abnos are easier to categorize that way. But if its a concern we can make records clean doors

- Abnormality datums now contain the camera and door they are using, specifically for records
> Looking for them each time we want to teleport is probably a bit wastefull, so just made them save on the datum for us to use in any way we want

- Makes basilsoup properly use `SpawnConnectedStructure()` instead of forcemoving its structure immediatelly after spawning it
> It was buggy, and didnt work for records. No reason to keep it

- Records will now always teleport at least 1 abno per meltdown, if it can
> I feel like it would be good for both early and late-game thingies to teleport

## Changelog
:cl:
fix: records core should no longer have 2 abnos in the same cell
/:cl:


Kitsunemitsu (approved):
Pretty easy PR merge. if there's any serious issues I can revert it

PR #2426, opened 2024-09-14
Body Preservation Unit

EnderMage99:
## About The Pull Request

Adds a "Body Preservation Unit" (BPU) to the City Fixers/CoL maps. Civilians can use the BPU by paying a certain amount of Ahn (Their total states combined * 5) to create a record in the unit. This record will contain all of their attributes at the time of payment. They are also able to update their record by spending more ahn. Afterwards, Once they die they will be able to recreate their body at the BPU after 2 minutes. This new body will have a small penalty of -6 to all attributes, and will not remember their previous skills.
Body Fab

## Why It's Good For The Game

This will be very helpful to City Fixers/CoL since it will give a chance for players to return to the game, without having to wait half an hour. It could also help those players who get killed all alone in the backstreets and no one tries/can save them. Also, It will not replace the doctor since the "BPU" makes them lose a good bit of progress.

## Changelog
:cl:
add: Adds the "BPU" to the City Fixers/CoL maps
add: Adds the "Torso Fabricator" to the City Fixers/CoL maps
tweak: tweaked the City Fixers/CoL maps to include some spare outfits near the "BPU"
/:cl:


Kitsunemitsu (approved):
Been TMed for a while, couple bugs but whatever tbh

PR #2429, opened 2024-09-16
Players can no longer join with the same name.

Kitsunemitsu:
## About The Pull Request
Players can no longer infinitely duplicate themselves.

## Changelog
:cl:
tweak: players can no longer join if a player with the same name is in the game.
/:cl:


InsightfulParasite:
Alternatively we could randomize it to Alpha 1 clerk names if people want to join anyways.

PR #2431, opened 2024-09-17
Dealer of the damned final observation

X3ros:
## About The Pull Request

Adds dealer's final observation and fixes its description to match its current design

## Why It's Good For The Game

Ensures that Dealer has a Final observation that's more interesting than a placeholder and prevents confusion when examining them.

## Changelog
:cl:
add: Final Observation for DotD
fix: Fixed Dealer's examine text.
:cl:


code/modules/projectiles/projectile/ego_bullets/zayin.dm
```diff
@@ -57,3 +57,12 @@
...
 	damage_type = WHITE_DAMAGE
+
+/obj/projectile/ego_bullet/ego_dud
+	name = "dud"
+	damage = 1
+	damage_type = RED_DAMAGE
+
+/obj/projectile/ego_bullet/ego_dud/Initialize()
+	qdel(src)
+	..()
```
Gboster-0, line 68: NEVER do this with Initialize() man
eighter do `. = ..()` or `return ..()` as otherwise we get silent errors from `..()`
```suggestion
	return ..()
```

X3ros: Gotcha. Thankfully, this bullet was never actually implemented due to it breaking the guns it uses somehow, but I'll try out the suggested code to see if it works.

X3ros: Tested both suggestions, neither seem to actually fix the "permanently broken gun" problem. Gonna just scrap the "dud round" idea for Death dealer until I'm more experienced with BYOND witchcraft.

PR #2455, opened 2024-10-01
Liu section 1 and section 2 swap places in their roaming routes.

KingSuperBOB:
## About The Pull Request
Swaps Liu 1 and Liu 2 making Liu 2 rare and Liu 1 uncommon.

## Why It's Good For The Game
The rarity system is meant to make less players spawn as overall strong offices, the Liu1 roaming has normal Liu armor and a fire fist, the Liu2 roaming has veteran armor which is better in everyway and the fire spear which not only has more range but also does more damage than the fire fist.

## Changelog
:cl:
tweak: swaps liu2 and liu1 in rarity
/:cl:


Kitsunemitsu (approved):
Still mixed on this; but I'll pass it through.

PR #2461, opened 2024-10-05
Reworks how the possession blacklist is handled, Admins can now disable possessing on individual abnormalities.

Gboster-0:
## About The Pull Request

- Changes the global forbidden abnormality possession blacklist into a var on every abnormality that gets checked instead
- Moves some if() checks around into better spots
- Removes the admin message if someone tries to possess whilst possession is not enabled

## Why It's Good For The Game

> Changes the global forbidden abnormality possession blacklist into a var on every abnormality that gets checked instead
- This way, admins can disable any abnormality they are preparing for adminbus and dont want a ghost entering into. Also cleaner
> Moves some if() checks around into better spots
- Please, let me bury the history of me making the code check if possession is enabled in SSlobotomy subsystem EVERY time it checks if a mob is valid to possess, god the inefficiency
> Removes the admin message if someone tries to possess whilst possession is not enabled
- It's not really needed, no harm is being done

## Changelog
:cl:
admin: you can now disable possession on individual abnormality's by changing their `do_not_possess` var
/:cl:

### DIFF
--- a/ModularTegustation/misc_procs.dm
+++ b/ModularTegustation/misc_procs.dm
@@ -9,7 +9,6 @@
 /datum/proc/try_take_abnormality(mob/dead/observer/possessing_player, mob/abnormality)
 	if(!SSlobotomy_corp.enable_possession) // uhhhh, how did you even access this proc?
 		to_chat(usr, span_userdanger("Abnormality possession is not enabled!"))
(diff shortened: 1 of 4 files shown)


Kitsunemitsu (approved):
Approved. Love it

PR #2466, opened 2024-10-08
Fixes K-corp healing drones runtiming every time they hit someone, and them not healing anyone

Gboster-0:
## About The Pull Request

- Fixes K-corp drones runtiming due to checking for stamina armor that does not exist every hit
- Fixes K-corp drones never succeeding in their is_human() check due to checking a non-existant entity
- Adds a feedback message about your wounds healing when a drone jabs you
- Adds a defective toggle to K-corp healing drones that makes them mimic their old broken behavior, for fun

## Why It's Good For The Game

> Fixes K-corp drones never succeeding in their is_human() check due to checking a non-existant entity
- The proc is never passed anything under normal circumstances so we were checking the wrong thing, what we want to check is our `target` var with the actual person we are trying to heal

> Fixes K-corp drones runtiming due to checking for stamina armor that does not exist every hit
- Bug fix gud

> Adds a feedback message about your wounds healing when a drone jabs you
- When i fixed the stamina thingy by directly applying the stamina damage instead of checking for armor, we lost the big red text about being jammed in favor of a small blue one, that is very un-noticable
- So now they give a green message about your wounds painfully closing

## Changelog
:cl:
fix: K-corp healing drones no longer break everything.
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/representative/items/kcorp.dm
+++ b/ModularTegustation/tegu_items/representative/items/kcorp.dm
@@ -108,8 +108,10 @@
 		var/mob/living/H = target
 		H.adjustStaminaLoss(rand(14, 18))
 		if(!defective)
-			to_chat(H, span_nicegreen("You feel your injuries painfully close!"))
 			H.adjustBruteLoss(-20)
+			to_chat(H, span_nicegreen("You feel your injuries painfully close!"))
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (approved):
Nice bugfix! Approved.

PR #2471, opened 2024-10-09
Organs now rot significantly slower.

Kitsunemitsu:
## About The Pull Request
Organs now all rot by 60 minutes, except the heart which is at 30 (becuase the heart has a super easy to remember surgery)
Max brain traumas gained from rot is always healable by surgery

## Why It's Good For The Game
It sucks rotting on the ground for 60 minutes only to then need another 60 minutes of surgery.
This makes rotting less common.

## Changelog
:cl:
balance: Organs rot slower
balance: Brains no longer get permanent traumas
/:cl:


InsightfulParasite (approved):
Makes organs more good for eatings approved!

PR #2474, opened 2024-10-10
Moves the brain attributes into the body fabricator file for modularity, saves it on death instead of constantly re-saving it every life tick

Gboster-0:
## About The Pull Request

- Moves the brain attributes into the torso fabricator file
- Makes the brain attributes only save when the user is killed, instead of saving every life tick

## Why It's Good For The Game

> Moves the brain attributes into the torso fabricator file
- I like modularity :3
> Makes the brain attributes only save when the user is killed, instead of saving every life tick
- Saving it every life tick is wastefull, its better to just do so when the user is killed. Afterall the torso fabricator should **NEVER** have a brain inserted into it before that brain's host dies

## Changelog
:cl:
code: the brain now only saves attributes to itself when the user dies, since it does not need to do it before the user dies.
/:cl:

### DIFF
--- a/ModularTegustation/lc13_machinery.dm
+++ b/ModularTegustation/lc13_machinery.dm
@@ -329,23 +329,20 @@
 	var/stored_temperance = 0
 	var/stored_justice = 0
 
-/mob/living/carbon/human/death(gibbed)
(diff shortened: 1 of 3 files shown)


Kitsunemitsu (approved):
Deshits our code

PR #2475, opened 2024-10-10
Nerf Gold Dawn Beanstalk Corrosion Damage

tile4:
## About The Pull Request
Reduces the damage that beanstalk corrosions do by 3

## Why It's Good For The Game
Considering they had the same damage output as Steel dawns but tended to spawn in higher numbers, they were too strong with that and having a range of 2 to the point where they kind of made overshadow the actual bosses of the ordeal.

## Changelog
:cl:
balance: rebalanced beanstalk corrosion damage
/:cl:


Coxswain-Navigator (approved):
Sure, why not.

PR #2486, opened 2024-10-15
Fixes faelantern lowering qliphoth if you are below 2 temperance regardless of the work

Gboster-0:
## About The Pull Request

- Fixes faelantern lowering its qliphoth if you are below 2 temperance, on any work
(The original intent was that you can only do repression if you have 1 temperance)

- Fixes a wrong check, making it so if you are exactly on point with your temperance at 40 you can work repression without lowering faelantern's qliphoth

## Why It's Good For The Game

> Fixes faelantern lowering its qliphoth if you are below 2 temperance, on any work
> Fixes a wrong check, making it so if you are exactly on point with your temperance at 40 you can work repression without lowering faelantern's qliphoth
- Bug bad, very bad in this case

## Changelog
:cl:
fix: faelantern no longer lowers its qliphoth on all works if you are below 2 temperance
/:cl:


Kitsunemitsu (approved):
Approved.

PR #2488, opened 2024-10-15
FISHING 5.0 (I say morning, cause it's not a good morning unless my chakra is aligned)

Kitsunemitsu:
Note: all of this only applies to the Cityfixers and City Maptypes.

# Major Fishing Expansion.
## NEW PLANETARY ALIGNMENT SYSTEM
There's 8 planetary bodies. Each of these planets moves in orbit around the sun along with earth.
If they align with earth, then all fishermen get a buff to a specific skill (Can be found on the wiki, or mentioned when it happens)
Each planet moves in different speeds, with the moon taking 4 cycles as well.

## NEW FISHING GODS
There's 7 fishing gods to devote yourself to at a fishing shrine, one for each planetary alignment aside from the moon
Get devotion by sacrificing fish, fishing, and praying when your planet is the highest in the sky.

## NEW DYNAMIC FISHING ENVIRONMENT
Your fishing options and buffs expand as more planets reach alignment.
While you are devoted to a god, you get various passive buffs to one specific portion of fishing.

## FISHING SKILLS
12 BRAND NEW FISHING SKILLS
These skills focus on mostly the fishing tech tree, with each fishing skill interacting with your devotion, the phases of the moon, or the planetary alignment that use your devotion as a resource.
The idea is that you get a bunch of people vying for the alignment of their planet over others.

### LEVEL 1 FISHING SKILLS
Scry - Detect which planets are currently in alignment
Planet - Move each planet forwards by one phase
Lunar Prayer - A healing skill that gets better the more in alignment the moon is.
Sacred Word - Pray to your god and get devotion should your planet be in alignment
Detect Fish - Detect if you can fish nearby

### LEVEL 2 FISHING SKILLS
Smite the Heretics - Deal damage to everything nearby based off devotion. Does no damage to players of the same god, deals more damage to atheists
Lunar Might - Increase Justice based off the current phase of the moon
Awe the Weak - Immobilize any atheists nearby
Chakra Misalignment - Deals damage to any player with a god, who's planet is not currently in alignment. If your planet has been destroyed, you will die.

### LEVEL 4 FISHING SKILLS
Alignment - Align your god's planet with earth
A Moment in Time - Freeze all planets until the moon is full again
Supernova - Destroy one planet. It will never go into alignment again.

For more specifications, see the wiki page. https://wiki.lc13.net/view/Guide_to_Fishing

## Why It's Good For The Game
Fishing has always been a pretty hit or miss on if you like it or not.
I want to add in more esotericism, and more unique mechanics.
None of this will affect LC13, and just affect CF and COL.

This is a content expansion to have fishermen join the rest of the game in enjoyment, while allowing for both a chilled back and supportive playstyle.

For those otherwise, this should add in some strife to a playstyle that sees much love but not much change. I expect players to be fighting over planet alignment and more.

## Changelog
:cl:
add: Added Fishing Skills
add: Added Planetary Alignment System
add: Added Fish god system
/:cl:


ModularTegustation/fishing/code/fish_market.dm
```diff
@@ -152,3 +157,61 @@
 		if((FISH_RARITY_VERY_RARE + 1) to FISH_RARITY_RARE)
 			fish_worth = 10
 	return round(fish_worth * fish_worth_mod)
+
+
+/obj/machinery/fish_market/proc/check_city()
+	if(SSmaptype.maptype in SSmaptype.citymaps)
```
Gboster-0, line 163: imo, should de-hardcode it
How about making the if() check where the proc is called instead of here, then renaming the proc to something like `udpate_stock()`. That would allow admins to call it manually on any map, and i think that'd be neat

Kitsunemitsu: It's actually hardcoded because the other god stuff isn't supposed to work outside of those maps.

Kitsunemitsu:
It's good to go~!

InsightfulParasite (dismissed):
Approval. The fishing begins.

InsightfulParasite (approved):
Approved for the fishing

PR #2495, opened 2024-10-21
Removes some items from J corp Lootboxes

tile4:
## About The Pull Request
Removes more debug and admin items, along with some other things that would break the game.

## Why It's Good For The Game
As funny as I thought it would be for these to be in, it would be better as a whole to have them removed as they can lead to problems

## Changelog
:cl:
balance: removed some items from lootbox pool
/:cl:


Kitsunemitsu (commented):
We can keep adding to this as we can!
Great work, Tile4

Kitsunemitsu (approved):

PR #2523, opened 2024-10-30
Fixes Bugs of Distorted Form

scionalu:
## About The Pull Request

Fixes DF from going invisible after transforming into bald
Makes DF also do the Unfreezemob on death, in case of gib etc
The list also gets cleared on Unfreezemob so it doesn't trigger twice

## Why It's Good For The Game

Fixes Known bugs of DF

## Changelog
:cl:
fix: DF invisibility and timestop bug
/:cl:


InsightfulParasite (dismissed):
Bug FIX!!! Approved.

scionalu:
AH, I thought it wasn't reviewed yet since not merged, so I just combined it with the new bugfix that we found today
Sorry about that

Gboster-0 (approved):
Tested locally and all works well.

PR #2525, opened 2024-10-31
Fixes HR only telling people the current core suppression and next ordeal if an ordeal is currently in progress

Gboster-0:
## About The Pull Request

- Fixes my mistakes:tm:

## Why It's Good For The Game

> Fixes my mistakes:tm:
- Both of those in fact should be 1 layer of indentation up, they were 1 layer down due to me previously only having these two being under the if() check, 100% my bad

## Changelog
:cl:
fix: Fixes me lobotomizing HR
/:cl:


Kitsunemitsu (approved):
Approved

PR #2529, opened 2024-11-02
removes the not in the abnocheck for sentience potions

Thundershade09:
## About The Pull Request

makes sentience potions blacklist abnos instead of everything except abnos

## Why It's Good For The Game

this is the intended code


Gboster-0 (approved):
Yup, this is the intent.

PR #2541, opened 2024-11-10
Endermage Rcorp Updates: Removal of Friendly Fire from Abnormalities [Done]

EnderMage99:
## About The Pull Request

This PR does 2 very small things.
1. It makes it so abnormalities are unable to melee attack their allies in R-Corp.
2. Whenever you possess an abnormality, That abnormality will do an emote which signals that it's player-controlled.

## Why It's Good For The Game

This PR helps R-Corp by removing a bit of clunkiness from melee fighting as an abno. I don't think having to worry about hitting your allies or yourself is a good way of making playing as an abno more challenging.

Being able to see when your allies become player-controlled can be useful to summoners like ML or CENSORED

## Changelog
:cl:
tweak: hostile mobs will become unable to attack themselves/allies in R-Corp.
tweak: abnos emote when they get possessed
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
@@ -140,6 +140,12 @@
 	// rcorp stuff
 	var/rcorp_team
 
+/mob/living/simple_animal/hostile/abnormality/Login()
+	. = ..()
+	if(!. || !client)
+		return FALSE
+	manual_emote("awakens...")
+
 /mob/living/simple_animal/hostile/abnormality/Initialize(mapload)
 	SHOULD_CALL_PARENT(TRUE)
 	. = ..()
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -673,6 +673,13 @@
 	walk_to(src, target, minimum_distance, delay)
 
 /mob/living/simple_animal/hostile/proc/AttackingTarget(atom/attacked_target)
+	if(client)
```
Kitsunemitsu, line 676: This is necessary for re-saning. FF is a feature, not a bug

code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
```diff
@@ -129,6 +129,12 @@
 	var/observation_success_message = "Final Observation Success!"
 	var/observation_fail_message = "Final Observation Failed!"
 
+/mob/living/simple_animal/hostile/abnormality/Login()
+	. = ..()
+	if(!. || !client)
+		return FALSE
+	manual_emote("awakens...")
```
Kitsunemitsu, line 136: This I like tho

code/modules/mob/living/simple_animal/hostile/hostile.dm
EnderMage99: This is only for abnormalities, I don't think abnormalities should be able to re-sane in R-Corp.

PR #2542, opened 2024-11-10
Endermage Rcorp Updates: Shrimp VIP Buff [DONE]

EnderMage99:
## About The Pull Request

This PR will change the the Shrimp VIP in a few ways:
1. It will give him 2 new abilities, Barricade and Airstrike. Barricade will let him place down up to 6 weak barricades which can be used to fortify an area, has a 15 second cooldown. Airstrike will let him select an area to constantly rain down missiles for a few seconds, has a 1 minute cooldown. [DONE]
2. It gives the shrimp as passive protection, where for each ally they can see, they will take 25% less damage. Up to 100% less damage. This will let the Shrimp VIP to take a much more active role in abnormality fights. And they will not need to just sit alone and hide. [DONE]
3. Makes it so if the VIP loses sight of at least 2 allies, a 5 minute timer will start. If they are able to get at least 2 allies back into their view, the timer is stopped. However, If they don't regain sight of at least 2 allies, they will take damage equal to 25% of their health as a unknown sniper fires at them. [DONE]

## Why It's Good For The Game

1. These abilities will let them be more proactive in a fight, Making it more interesting to play as them compare to what they are now.
2. This will let them stay with the team without having to worry about a random bullet firing in their direction and then dying. So they will not have to resort to a boring playstyle where they will be forced to sit in one corner in fear of a raven/rat picking them off.
3. This will stop the strategy where the Shrimp VIP hides in one corner of the map all alone, so humans become unable to find them for a long time. That old strategy is unfun for humans and the VIP. The Humans have to search to whole map even after they kill of the the abnormities, while the VIP just sits in one spot and basically AFKS.

## Changelog
:cl:
balance: reworks the shrimp VIP to be more fun to play as
/:cl:


Kitsunemitsu (approved):
Approved. Looks good to me.

PR #2546, opened 2024-11-10
REDUCE LAG AND DESTROY AUTO PATROL CLEANBOTS [DONE]

InsightfulParasite:
## About The Pull Request
Reduces lag by making mobs a bit more nearsighted on city maps by reducing their vision range to 7-10 and disabling wander for all creatures. I hate wander. Bot vision range was reduced from 7 to 5 and cleanbot patrol was turned off.

Quick testing showed no runtimes.

## Why It's Good For The Game
Doing this during the last city map seemed to improve preformance.

## Changelog
:cl:
tweak: _DEFINES\robots.dm DEFAULT_SCAN_RANGE
tweak: set cleanbot auto patrol to FALSE
tweak: auto set vision range to 8 on city maps if more than 7
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -38,7 +38,7 @@
 
 
 //These vars are related to how mobs locate and target
-	var/robust_searching = 0 //By default, mobs have a simple searching method, set this to 1 for the more scrutinous searching (stat_attack, stat_exclusive, etc), should be disabled on most mobs
+	var/robust_searching = FALSE //By default, mobs have a simple searching method, set this to 1 for the more scrutinous searching (stat_attack, stat_exclusive, etc), should be disabled on most mobs
 	var/vision_range = 9 //How big of an area to search for targets in, a vision of 9 attempts to find targets as soon as they walk into screen view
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -83,6 +83,9 @@
 		move_to_delay - 2 = speed. */
 	UpdateSpeed()
 	. = ..()
+	if(SSmaptype.maptype in SSmaptype.citymaps)
+		if(vision_range > 7)
+			vision_range = 8
```
Gboster-0, line 88:
```suggestion
		vision_range = min(vision_range, 8)
```

InsightfulParasite: Tested and seems to work just as well. Ill do a quick second test on a non city map to make sure it doesnt effect normal LC.

InsightfulParasite: Okay works flawlessly.

Gboster-0: well, because its basically the same
min() takes in some arguments, and puts out whatever's the lowest. Thus if the vision_range is higher than 8 its set to it

Kitsunemitsu (approved):
While I like wander as a way to make some abnos bump into doors, I am totally willing to approve this PR

PR #2547, opened 2024-11-10
Endermage Rcorp Updates: Buffing/Reworking R-Corp Abnormalities [DONE]

EnderMage99:
## About The Pull Request

At the time of creation, this PR will buff/rework the following Abnormalities:
- Red Blooded American: They now deal much more damage with their shotgun, gains a bit of health, and gains a reload mechanic where they can spam their shot in bursts, however, they will need to wait longer to regain bullets to fire again. They can also regain bullets by melee attacking.
- Scarecrow: Fixed a bug where they could not reuse dead bodies, and gives them a new ability called "Hungering for Wisdom." When they trigger this ability, they will enter an enraged state where they gain movement speed, deal more damage, and heal 5% of their health on hit. However, if they don't hit a human target every 10 seconds, they will enter a state of weakness where they take damage and slow down for a short amount of time.
- Clouded Monk: Makes it so their charge attack performs a 5 by 5 aoe at the end of each rush, dealing 50 RED damage.
- Ppodae: Now, They have 2 different forms. A "Cute"/"Buff" form. They can switch forms using their ability, which has a 5-second cooldown. While they are in the "Cute" form, they have their massive speed boost, and they are able to crawl under living mobs and mechs, but can't do their slam attack. While they are in the "Buff" form, they have a reduced speed, but take 75% less damage from all attacks and their slam attack can damage objects. In both forms, they can remove limbs from targets, which heals them 10% of their HP per limb removed.
- Porccubus: Their movement is now based on charges. They regain one every 2 seconds and can hold up to 3. Each leap has no cooldown but consumes a charge each time they leap. They can't leap without charges. Also, their health is doubled to account for the melee damage buff.
- Eris: Removed the nerf she has gotten where she could only heal abnos under 50% health.
- KoG: Changed the charge to deal more damage than longer they charged for.
- R-Corp Nerf: This will reduce the damage of the grenades by 100, and make them deal more damage to their user.

[This pr is done, more abno will be updated in the next PR, Endermage Rcorp Updates: Buffing/Reworking R-Corp Abnormalities Part 2.]

## Why It's Good For The Game

Overall, each of these abnormalities are currently really unfun to play as. Giving these buffs to them will make them have a place in their respective team, while making them require more thought to play in exchange for contributing much more to the fight.
- Red Blooded American is currently so pathetic and can't their team in any way. They are a damage deal who deals the least amount of damage, deal even less them some support abnos. They have to way away from their team like KoD, but doesn't even have close to the same payoff. This buff will make them have some worth in the team, and will require some more though into playing them, as they will need to manage their ammo.
- Scarecrow is also as pathetic as Red Blooded American, they are a tank who relies on bodies. However they gain a very low amount of health from draining a body which makes them the weakest tank of the easy team. This rework will let them play a more aggressive playstyle where they want to be in the enemy team to deal the most damage/gain healing, however they need to be careful to not get too deep into the r-corp team, as they will still get shredded.
- Monk is okay-ish but their charge is really difficult to hit, which is strange since his charges are a limited resources. We got KoG who has a charge that can be use unlimited times, has 0 start up, hits multiple people and has 3 by 3 range to hit. Giving Monk a bit of more range with his charge will make his charges have much more impact, which is important due to their limited amount. Along with that, it will be at least possible to hit with his charges, since they have a decently long delay.
- Ppodae is currently a non-threat to R-Corp. They do a very low amount of damage and their slam attack stops them, which makes them easy to kill since they have the lowest amount of health of all abnos in all of R-Corp. This rework lets them become a much more active threat, which can make R-Corp pay attention to them. The form switching can be a fun way of balancing their movement speed along with their better DPS, since the player will need to think of when they should switch in or not.
- Porccubus: This is a QoL change, which makes them a bit more comfortable to play, since it will improve their mobility without making them impossible to catch.
- Eris: Turns out, she was a really weak heal so the nerf was underserved.
- KoG: The last nerf basically made her unplayable due to how long humans had to react to her. This will make it so KoG is able to do something if they are right next to her, but they still don't have the ability to instantly kill at melee range.
- R-Corp nerfs: Currently, grenades deal a MASSIVE amount of damage, being able to 4 shot most easy team abnos and humans are able to rush at abnos while keeping them in their hand which makes it really difficult for abnos to counter it.

## Changelog
:cl:
balance: rebalanced some rcorp abnormalities
balance: nerfed some rcorp weapons/tools
/:cl:


Kitsunemitsu (approved):
Was tested over the course of weeks. Approved.

PR #2554, opened 2024-11-15
Fix Lobotomy Corp Achivement Icons [testmerge whenever]

InsightfulParasite:
## About The Pull Request
Achievement icons for LC13 were not working and it turned out that making a icon for a achievement is a stupid two step process that involves adding them to a list.
All achievements and achievement icons were added to a new file for LC13 achievements.

Worst news is that for some ungodly reason achivements do not work when i local test. Its super stupid so i need to testmerge this eventually.
## Changelog
:cl:
fix: achivement icons
/:cl:


Kitsunemitsu:
They actually don't work on a local because they require a database haha

InsightfulParasite:
> They actually don't work on a local because they require a database haha

Suffering

Kitsunemitsu (approved):
Bug is confirmed fixed.

PR #2555, opened 2024-11-15
Fixes some init hints and amber bugs hard deleting themselfes saving a LOT of hard-delete time

Gboster-0:
## About The Pull Request

- Makes clerkbot constructors go inside the clerkbot and pop out when the clerkbot is killed instead of constantly being destroyed/spawned
- Fixes clerkbot initialize hint
- Fixes KHz corrosion initialize hint
- Makes amber bugs hold a reference to their spawning mob, if one exists. And delete themselfes from the spawned_mobs list.
- Fixes some instances of using `.len` instead of `length()`
- Makes single-letter vars in KHz's corrosion into actual var names
- Deletes the linked_subsystem var from fish planets and makes them call SSfishing instead

## Why It's Good For The Game

> Makes clerkbot constructors go inside the clerkbot and pop out when the clerkbot is killed instead of constantly being destroyed/spawned
- Its most likelly cheaper this way, and a bit more how you'd expect it to work in-code
> Fixes clerkbot initialize hint
- guh
> Fixes KHz corrosion initialize hint
- bah
> Makes amber bugs hold a reference to their spawning mob, if one exists. And delete themselfes from the spawned_mobs list.
- Cheaper, they constantly hard-deleted themselfes because the spawning mobs hold a reference to them. Let them rest and save us the like 4-5 times the time they should take to delete.
> Fixes some instances of using `.len` instead of `length()`
> Makes single-letter vars in KHz's corrosion into actual var names
- Shitcode my beloved
> Deletes the linked_subsystem var from fish planets and makes them call SSfishing instead
- If an admin creates a planet, the linked_subsystem is probably never getting set so this is a good chunk safer

## Changelog
:cl:
code: amber bugs now hold a reference to their creating parent, and delete themselfes from the list instead of clearing nulls.
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/gadgets/unpowered.dm
+++ b/ModularTegustation/tegu_items/gadgets/unpowered.dm
@@ -598,11 +598,11 @@
 	if(!istype(user) || !(user?.mind?.assigned_role in GLOB.service_positions))
 		to_chat(user, span_notice("The Gadget's light flashes red. You aren't a clerk. Check the label before use."))
 		return
-	new /mob/living/simple_animal/hostile/clerkbot(get_turf(user))
+	var/mob/living/simple_animal/hostile/clerkbot/bot = new(get_turf(user))
+	forceMove(bot)
 	to_chat(user, span_nicegreen("The Gadget turns warm and sparks."))
-	qdel(src)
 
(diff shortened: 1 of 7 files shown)


Kitsunemitsu (approved):
Important bugfixes

PR #2565, opened 2024-11-19
Weaving Lindworm

Kitsunemitsu:
## About The Pull Request
Adds Weaving Lindworm.
This is an abnormality that has the theme of "Shedding armor" so that it loses it's defenses.
Instead of dying, lindworm releases a damage turf range. It also does this on attack.
It then sheds it's skin and heals to full. It deals this one time per layer, and starts with 5 layers

Lindworm actually lowers the defenses of the person who worked on it, based off what was worked on it.
To encourage this, Lindworm has quite high work rates; with each work, and lowers it's own layers per work.

It's EGO gear however, is very strong. While it has a low attack rate, and stuns you on attack, it also rends both red and white.
All on top of 4 range, making it an excellent support weapon.

## Changelog
:cl:
add: Added Weaving Lindworm
/:cl:


code/modules/clothing/suits/ego_gear/waw.dm
Tokanova, line 1: Only Be Becoming? Was that supposed to be Only by becoming?

code/modules/mob/living/simple_animal/abnormality/waw/lindworm.dm
```diff
@@ -0,0 +1,210 @@
...
+
+/mob/living/simple_animal/hostile/abnormality/lindworm/AttackingTarget()
+	..()
+	for(var/turf/open/T in view(1, target))
+		if(locate(/obj/structure/turf_confetti) in T)
+			for(var/obj/structure/turf_confetti/floor_fire in T)
+				qdel(floor_fire)
+		new /obj/structure/turf_confetti(T)
+
+	if(ishuman(target))
```
InsightfulParasite, line 103: Your going to want to put a CanTarget override for non humans. Currently the stat attack applies to simplemobs and causes the lindworm to attack ordeal corpses.
Essentially just put in 
/mob/living/simple_animal/hostile/abnormality/lindworm/CanAttack(atom/the_target)
	if(isliving(the_target))
		var/mob/living/carbon/host = the_target
		if(host.stat == DEAD && !ishuman(the_target))
			return FALSE
	return ..()

InsightfulParasite: Actually it might be easier to do it backwards and remove the stat_attack and just put in a can_attack override for dead humans.

Kitsunemitsu: Was fixed!

tile4 (changes requested):
The paperwork says the counter decreases on bad works but I don't see it in the actual code for the abno.

PR #2572, opened 2024-11-23
Kikimora Ego [DONE]

InsightfulParasite:
A teth dress and armor for the Kikimora

## About The Pull Request
Adds Kikimora Ego
image

## Why It's Good For The Game
Atomized Ego

## Changelog
:cl:
add: Kikimora Ego
/:cl:


InsightfulParasite:
Descriptions for EGO are sort of placeholder.
I was trying to brainstorm a story for this creature and only came up with the idea the Kikimora was originally a shapeshifter who lived in a persons home. Some element of the story would have the idea of not being understood even amongst people you lived with your entire life. She would be trapped in her current form due to a bear trap taking her other leg. But i couldnt really~ figure out a way to express that story in her broom and dress. The shapeshifter part of the story might be a bit too much so we may change it later on to a different kind of creature.

My dread is that people could call the Kikimora a abberation of Beauty and the Beast since they both have elements of a curse. For the kikimora however the curse spreads outwardly rather than taking on another host.

PR #2577, opened 2024-11-30
Endermage Rcorp Updates: Buffing/Reworking R-Corp Abnormalities Part 2 [DONE]

EnderMage99:
## About The Pull Request

This PR will change the following abnos in R-Corp:
- Blue Shepherd [DONE]:  I will be adding 2 new abilities to them, Evade and Parry. Both can be triggered with a hotkey, (Spacebar for Evade and Right Click for Parry). Evade causes the Blue Shepard to lose density and gains a speed boost for 1 second. After that speed boost, they are slowed down for 1 second. There is no cooldown for this ability, they just can’t use it while slowed down or sped up. Parry can be turned on and off with a toggle. Then they trigger the parry, if they take damage within the next 1 second, they will instantly perform a 5x5 AoE which will knock down all carbons. On success or failure, it does on a 5 second cooldown after use.
- Der Freischütz [DONE]: While they have sniper vision active, they are able to place a portal anywhere where they can see. Placing down a portal has a 10 second cooldown. They are able to have a max of 3 portals. Then, they are able to change their perspective to any of their created portals. While viewing through a portal, they are able to fire through the portal. Also, The time to fire normally and through the portals is reduced from 1.5 seconds to 1 second. Portals have 1000 HP, they have no density and have normal resistances to everything.
- All Round Helper [DONE]: While they are dashing, they will be able to change the direction they are dashing without a cooldown, and they inflict bleed to targets hit by the dash. Also, they can't change their direction after hitting a human for 1 second.
- Drifting Fox [DONE]: Then more umbrellas they have active, then faster they become.
- Yin [DONE]: Their counter will be replaced with a faster pulse which deals more damage, and their beam will become an attack they can use by clicking on a tile. It will have a 8 second cooldown. Also, their movement speed will be buffed.
- Clown Smiling at Me [DONE]: One, I will reduce the cooldown of throwing their knife from 4 seconds, to 0.5 seconds. Also, hitting a human with a knife will inflict them with 8 “Bleed”. Bleed deals Brute damage each time the target moves equal to its value and then reduces itself by half. Also, clown’s knives now go through allies. Letting them use them behind safety. Also, Their finishing attack is buffed, it used to deal 5 WHITE damage 7 times within 1.4 seconds before finishing by dealing 30 WHITE damage. Now it will deal 12 WHITE damage 7 times within 1.4 seconds before finishing by dealing 80 WHITE damage, dealing 164 WHITE damage, rabbits take half damage so it will deal 82 WHITE damage in total. Also, Their death explosion will inflict 30 “Bleed” to nearby Humans.
- Warm-Hearted Woodsman [WORK IN PROGRESS]: Warm-Hearted Woodsman will gain a new ability which lets them throw their axe, which is attached to them with a chain. If it hits a human, it will give them a “Chained” statues effect and there will be a visible chain between them and the Woodsman. While they are chained, they are pulled 2 sqrs closer to the woodsman every 0.75 seconds. After every 2 pulls, he will wait an additional 0.25 seconds for the next pull, where he will pull them 4 sqrs closer to them instead of 2. This continues until either they break line of sight, or they are right next to the Woodsman.

## Why It's Good For The Game
- Blue Shepherd: Currently, They are suffering from having no gimmicks. They only have a basic AoE which forces them to stand still for a decent amount of time and deal a low-decent amount of damage. They just feel basic, and not in a good way.

- Der Freischütz:  They are similar to KQE as in they need the correct map to function at their peak, however KQE has other things other than their ability like tankness and a melee AoE. Der Freischütz has absolutely nothing if they don’t have the right map, they can’t melee attack, low health, slow and their payoff is surprisingly low for the amount of things they need that need to go right for them to function.

- All Round Helper: The only ability they have harms them much more than help them. The dash deals a very low amount of damage compared to all other dashes, and stuns him as much as KoG. Also, They are surprisingly fragile for an abno that is supposed to be charging into the rabbits and can be easily stopped by a rhino being in the way.

- Drifting Fox: This is a bug fix, their speed was changing in the code but the update speed proc was never called.

- Yin: Tied with Jangsan Tiger for the slowest abnormality in the game, they don’t have anything that deserves this downside. They are supposed to be a tank, yet their slow speed doesn’t give them a chance to position themselves in an area which would block damage for their team. Also, They are supposed to be an area denying abno, yet all of R-Corp is able to move past it without repercussions since it can’t attack if it is not attacked back and it is too slow to chase after them.

- Clown Smiling at Me: For a hard team combat abno, they are quite weak at fighting R-Corp. Their knife throws have a decently long cooldown and will hit rhinos most of the time, which makes the slowdown it applies useless since it can’t hit humans. And the payoff they get from using dead bodies is the smallest when compared to the rest of the body eating abnos since the amount of WHITE damage it deals is low. (Better than Woodsman by a small margin.)

- Warm-Hearted Woodsman: Some problems as Blue Shepard but worse. Their AoE has an even bigger delay then Blue Shepard and their gimmick of needing to eat bodies is by far the weakest in the whole game, since it only buffs the damage of his AoE which is already predictable as hell.

## Changelog
:cl:
balance: rebalanced a few r-corp abnos.
/:cl:


Kitsunemitsu:
After fucking forever, I feel okay merging this,

Kitsunemitsu (approved):
Balance is good.
Took me forever due to the long, long balance process

PR #2583, opened 2024-12-02
Mobwave Spawner Fix

InsightfulParasite:
## About The Pull Request
Fixes some broken code in the mobwave spawner that was causing the wave commander to dissapear and delete from the list that was supposed to be static.

## Changelog
:cl:
fix: mobwavespawner structure
/:cl:


Gboster-0 (approved):
looks gud

PR #2588, opened 2024-12-05
Fix: Move default to config/example, ignore config

Cupax3:
## About The Pull Request
Local configs should be able to be static between branches and be local!! >:(

## Why It's Good For The Game
N/A

## Changelog
:cl:
tweak: Default config now lives in /config/example
tweak: /config/ is now gitignored
/:cl:


Kitsunemitsu (approved):
I'll approve it. Does make things a little more annoying however.

PR #2604, opened 2024-12-09
SoW Hermit Fix

scionalu:
## About The Pull Request

Fix the bug where Azure Hermit still deal damage even though you're out of his AoE range
Also gives a more clear indication when the attack finishes

## Why It's Good For The Game

The only good bug is Gregor or smth

## Changelog
:cl:
fix: fixed SoW bug
:cl:


Kitsunemitsu (approved):
Approved. Looks good to me.

PR #2609, opened 2024-12-14
Fix Mermaid Water Tiles

InsightfulParasite:
## About The Pull Request
Fixes mermaid AI. Mermaid on spawn is placed on deep water tiles that are labeled as dense but coded to be non dense in order to force mobs to avoid them. Mermaids AI thinks that she is in a 3by3 wall due to this. I changed obsessing water subtype to nondense. to fix this. 
Turns out i made the wrong watertype non dense but ill keep obsessing water changed too since the density is to prevent AI from pathfinding into water tiles they can sink into.
## Changelog
:cl:
tweak: obsessing water and safe saltwater is now non dense.
/:cl:


Kitsunemitsu (approved):
Approved, minor fix.

PR #2614, opened 2024-12-16
Fix Warbanner and Warcry

scionalu:
## About The Pull Request

Agent captains have been unknowingly killing everyone by using their "buff"
Said buff doesn't delete its list of affected people and subsequent uses add the same people into the list,
making it permanently take away their resistance on every use which stacks
Also minor cleanup for both codes

## Why It's Good For The Game

Removes the secret strongest Level 4 skill in gone from CoL/CF
Infinite Debuff

## Changelog
:cl:
fix: fixed the lv4 team fixer skills
:cl:


Kitsunemitsu (approved):
Approved. Thanks for the fix!

PR #2620, opened 2024-12-18
Holopads now use ACCESS_COMMAND and not ACCESS_KEYCARD_AUTH

Kitsunemitsu:
## About The Pull Request
Holopads now use Access command and not keycard auth.
This is a bug preventing the manager from calling down to the floor.

## Changelog
:cl:
fix: fixed holopad access.
/:cl:


InsightfulParasite (approved):
Phone no longer asks for the ID of a local police officer. Approved.

PR #2621, opened 2024-12-18
Roundtime fix

Halonexus:
## About The Pull Request
This fixes roundtime timestamp on my local, should probably work for the server too.
I assume some new byond version changed how those procs worked and the old implementation relied on some quirky stuff.

## Why It's Good For The Game
Correct time.

## Changelog
:cl:
fix: Round time timestamp fixed.
/:cl:


Gboster-0 (approved):
Ayup, thats how TG does it. Nothing to think about

PR #2626, opened 2024-12-19
Adds icons to City of lag/City fishers backstreets map spawners

Gboster-0:
## About The Pull Request

- Adds icons to every room spawner that backstreets use to generate, they encompass the entire area they take to spawn with yellow/black lines that are 0.5 width of a tile, so you can still see whats below them. A pink arrow shows where their predicted enterance should be

## Why It's Good For The Game

> Adds icons to every room spawner that backstreets use to generate, they encompass the entire area they take to spawn with yellow/black lines that are 0.5 width of a tile, so you can still see whats below them. A pink arrow shows where their predicted enterance should be
- It'll majorly help speed up wrongly placed spawners, and move them where they should be

## Changelog
:cl:
code: backstreets spawners now have proper icons
/:cl:


Kitsunemitsu (approved):
All icon changes, approved.

PR #2631, opened 2024-12-21
Representative Stock Market [Should be OKAY]

InsightfulParasite:
## About The Pull Request
Requested by Kirie. 
So this has been quite a ride, first i attempted to port TG stock market to LC13 but that ended up being a nightmare because the code was just everywhere. So i decided to remake it in a language im familiar with: Adventure Console.
This likely was irrational and real coders would go "why would you reinvent the wheel" but i found this much easier to handle.

The Stonk Console, has much of the same code as the TG Stock Console but has been infected with GENERAL_BUTTON defines and UI regulated to their own procs. To ease its creation ive made a few changes

- Stock companies are now predefined instead of procedurally generated. I understand people felt alot of connection to randomly generated events that said the company was revealing a new product made from orangutan meat but it felt a bit floated.
- The stock updates every 3 minutes with a "catch-up" code implimented for when someone checks after a while. Essentially if no one touches a stonk console anywhere then the company code will not update, but the moment a person uses a stonk console it will update the market by 1 market cycle per 3 minutes. 12 minutes will result in the market updating 4 times, hopefully this isnt resource intensive since its maxed at 8 cycles or 24 minutes processed each time a person checks it. I know its complicated and a bit insane to turn the stock market into a sort of shrodingers cat that only exists when you look at it.
- In order to prevent stocks from shooting into the millions i had to impliment a decay that effects stocks past 600 value. It happens 4 times until the decay is 100% and doesnt reset until the stock falls.

Main complains im on the watch for is if the console is "too slow" to update stocks or if the prices of stocks are not equal to the LC13 ahn economy since i dont even know how much a chocolate bar from the vendor is.

If bankruptsy is too frequent ill make a change to how bankruptsy occurs.

## Why It's Good For The Game
Hypothetically can be used by Representative for ahn loss/gain. 
## Changelog
:cl:
add: stockexchange computer
/:cl:


Gboster-0:
I will say, i am quite against the PR at its current state. Its code is all over the place
- it has several procs defined globally that should not exist (look at the proper TG `VOWELS` define, and then at `/proc/vowels()`
- It uses the non-good `for(var/i = 1, i++, i>4)` syntax instead of `for(var/i in 1 to 4)`
- And it has un-necessary `while()` loops that could EASILY be `for(var/i in 1 to 4)`

InsightfulParasite:
> I will say, i am quite against the PR at its current state. Its code is all over the place
> 
> * it has several procs defined globally that should not exist (look at the proper TG `VOWELS` define, and then at `/proc/vowel()`
> * It uses the non-good `for(var/i = 1, i++, i>4)` syntax instead of `for(var/i in 1 to 4)`
> * And it has un-necessary `while()` loops that could EASILY be `for(var/i in 1 to 4)`

Yeah it needs to cook a bit more. Im honestly looking into porting it so i dont have to code stocks.
Ill tweak it a bit more.

InsightfulParasite:
ughhhh im finding alot of flaws like the stocks being global and i think i have to gut and rework this entire system.

Gboster-0:
> ughhhh im finding alot of flaws like the stocks being global and i think i have to gut and rework this entire system.

yeaaaaaaaa i think thats for the best

Kitsunemitsu (approved):
This actually does not show up in game yet. Will be added in an update soon.

PR #2636, opened 2024-12-23
Actually fixes FAN's observation

Anonmare:
## About The Pull Request
Don't look at me, I'm embarassed I missed this for so long

## Why It's Good For The Game
Fixing boneheaded mistakes is good

## Changelog
:cl:
fix: FAN's final observation is actually fixed for real now
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/fan.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/fan.dm
@@ -52,7 +52,7 @@
 		if("Set it to 2")
 			observation_fail_message = observation_fail_message_3
 		else
-			observation_fail_message = initial(observation_success_message)
+			observation_success_message = initial(observation_success_message)
 	return ..()
 
 /mob/living/simple_animal/hostile/abnormality/fan/examine(mob/user)


code/modules/mob/living/simple_animal/abnormality/he/fan.dm
```diff
@@ -52,7 +52,7 @@
 		if("Set it to 2")
 			observation_fail_message = observation_fail_message_3
 		else
-			observation_fail_message = initial(observation_success_message)
+			observation_success_message = initial(observation_success_message)
```
Gboster-0, line 55: is this right? Ya sure you dont want this instead:
```suggestion
			 observation_fail_message = initial(observation_fail_message)
```

Gboster-0: originally it was setting the fail message to the success message, the success message itself is never changed from what i can see in the code so i dont really see a reason in reseting it

Gboster-0: Tested locally, nope. Does not work. Eighter way.
`observation_success_message` and `observation_fail_message` are both purelly cosmetic and serve to tell the user some text, and are NOT checked for success or failure.
ya can check it at https://github.com/vlggms/lobotomy-corp13/blob/master/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm#L561
Not a clue why it does not work and i blame observation code for this. Mark as resolved when ya read this, and please play-test your code.

PR #2637, opened 2024-12-26
How i lost my sanity painfully changing 159 files one by one -- Reworks observation code to be better for any amount of answers

Gboster-0:
## About The Pull Request

_Please end my suffering_
- Makes it so you can assign any amount of wrong or right answers to observations
- Makes every observation answer capable of having custom text, without extra proc overrides
- Makes buttons be randomly shuffled every time someone does an observation
- Fixed misspellings on the correct_answer list making abnormalities unable to be observed by removing it
- Fixed FAN observation
- Closes https://github.com/vlggms/lobotomy-corp13/pull/2636 , this fixes the same issue with the new framework

## Why It's Good For The Game

> Makes it so you can assign any amount of wrong or right answers to observations
- Ayup, we now have a framework for that, FAN enjoyers rejoice

> Makes every observation answer capable of having custom text, without extra proc overrides
- Yeah, thats now a thing

> Makes buttons be randomly shuffled every time someone does an observation
- This is a lil controversial of a change imo, but its for the better. Now all the correct answers are at the top of the list so we can organize a lil bit more neatly. Shouldnt have much of an effect on gameplay tho

> Fixed misspellings on the correct_answer list making abnormalities unable to be observed by removing it
- Now the format is `"Choice" = list(TRUE or FALSE [depending on if the answer is correct], "Response")` making it MUCH harder to make observations unachieveable since we now have to mis-spell "true" for "false" or vice-versa instead of a single letter

> Fixed FAN observation
- before we used the `in` keyword to search for the correct answer, i think that caused jank with abnos that have a lot of buttons (only FAN). This made it so we directly just yoink the value if its the correct answer or not instead of relying on `in`

## Changelog
:cl:
tweak: Observations now have randomly shuffled buttons
fix: FAN's observation now works
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
@@ -123,13 +123,19 @@
 	/// Offset for secret skins in the Y axis
 	var/secret_vertical_offset = 0
 
-	/// Final Observation details
-	var/observation_in_progress = FALSE
+	/// Final Observation stuffs
+	/// The prompt we get alongside our choices for observing it
 	var/observation_prompt = "The abnormality is watching you. What will you do?"
-	var/list/observation_choices = list("Approach", "Leave")
-	var/list/correct_choices = list("Approach", "Leave")
-	var/observation_success_message = "Final Observation Success!"
-	var/observation_fail_message = "Final Observation Failed!"
+	/**
+	 * observation_choices is made in the format of:
+	 * "Choice" = list(TRUE or FALSE [depending on if the answer is correct], "Response"),
+	 */
+	var/list/observation_choices = list(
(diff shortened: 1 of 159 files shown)


Kitsunemitsu:
I don't want to give this a review, I'm tossing this shit onto an intern @Blimwe0

Gboster-0:
> I don't want to give this a review, I'm tossing this shit onto an intern @Blimwe0

I think this might be the most cruel thing you have ever done

Kitsunemitsu:
It's just copy pasta

Kitsunemitsu (approved):
Just about all of this is formatting. I read the base _abnormality code and it seems good to go. I trust that there's no typos, and if there is we'll catch it in production :clueless:

PR #2638, opened 2024-12-27
Button changes -- Buttoning now takes 60 seconds instead of 45, vendors unmoveable, SOME doors unhackable

Gboster-0:
## About The Pull Request

- Buttoning now takes 60 seconds instead of 45
- The button counter:tm: now only resets after you button, instead of when you start attempting to harm button
- Vendors now cannot be un-wrenched from the floor
- Dont touch me and records doors were given the unhackable flag

## Why It's Good For The Game

> Buttoning now takes 60 seconds instead of 45
- I like a rounder number, should be that way imo
> The button counter:tm: now only resets after you button, instead of when you start attempting to harm button
- Always was meant to be that way, but i couldn't be bothered to earlier. This PR was a nice opportunity to fix it
> Vendors now cannot be un-wrenched from the floor
- Yeah no, lets not do that
> Dont touch me and records doors were given the unhackable flag
- You can destroy the doors, but it takes a while. Now you cannot hack the button doors directly nor the records doors leading to them, some doors in the departments where the button rests also got this flag (for example, 2 disposal doors) so its easier to get to the button aswell

## Changelog
:cl:
balance: The button is now takes 60 seconds instead of 45 to activate, the button was given several more safety measures
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/_tools/zayin/touch.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/_tools/zayin/touch.dm
@@ -26,55 +26,57 @@
 		to_chat(user, span_userdanger("THE BUTTON REJECTS YOU."))
 		return
 
-	cooldown = world.time + 45 SECONDS // Spam prevention
+	cooldown = world.time + 60 SECONDS // Spam prevention
 	for(var/mob/M in GLOB.player_list)
 		to_chat(M, span_userdanger("[uppertext(user.real_name)] WILL PUSH DON'T TOUCH ME[round_end ? "" : " TO BREACH ABNORMALITIES"]."))
 
 	if(round_end)
-		SSpersistence.rounds_since_button_pressed = ROUNDCOUNT_BUTTON_PRESSED
-		for(var/obj/structure/sign/button_counter/sign as anything in GLOB.map_button_counters)
-			sign.update_count(ROUNDCOUNT_BUTTON_PRESSED)
+		bastards += user.ckey
 		RoundEndEffect(user)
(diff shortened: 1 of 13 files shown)


code/modules/mob/living/simple_animal/abnormality/_tools/zayin/touch.dm
```diff
@@ -26,55 +26,57 @@
...
-		for(var/datum/abnormality/A in SSlobotomy_corp.all_abnormality_datums)
-			A.qliphoth_change(-A.qliphoth_meter) // Down to 0
+			flash_color(M, flash_color = COLOR_RED, flash_time = 150)
+			M.playsound_local(M, pick('sound/abnormalities/donttouch/kill.ogg', 'sound/abnormalities/donttouch/kill2.ogg'), 50, FALSE)
+			if(ishuman(M))
+				var/mob/living/carbon/human/H = M
+				H.gib()
+	else
+		for(var/mob/M in GLOB.player_list)
+			if(isnewplayer(M))
```
Kitsunemitsu, line 62: What is the point of this?
Shouldn't it check beforehand if they're a new player?

Kitsunemitsu (approved):
Looks just fine, important shit-clearing PR

code/modules/mob/living/simple_animal/abnormality/_tools/zayin/touch.dm
Gboster-0: explained in discord, isnewplayer() checks if the person is in the lobby mob.

PR #2658, opened 2024-12-31
City Gamemodes Update: Bloodfiends [DONE]

EnderMage99:
## About The Pull Request

This PR will add 3 new mobs, which replace the spawners in the ruined La Mancha Land backstreets layout.

Bloodbags: They are mobs with decent HP which take damage each time they attack. Each time they take damage, they will spawn a pool of blood around them with a short cooldown. Also they inflict bleed on hit. On death, they will explode after a short delay, creating pools of blood in a 3x3 area and inflict bleed to all humans in that area.

Bloodfiends: They have decent HP, and are able to absorb nearby blood to gain stacks of "Bloodfeast." Once they have enough bloodfeast, they will leap towards their target, inflict a good amount of bleed on hit. Then they will preform 2 short dashes towards the target. They also inflict bleed on hit.

Royal Bloodfiend: They work like normal bloodfiends, but they have more HP, and spend bloodfeast in a different way. One, Once they reach max bloodfeast, they freeze and will start marking 3 random directions around the target. The target will be able to see those directions. Also, the target will gain the city faction for the duration of this attack. Then they will leap to those direction in the order they created them, and preform an AoE around the target, leaving only 1 space open which is the opposite direction from which they are preforming the AoE. The attack deals low damage but inflicts a massive amount of bleed. If they miss all 3 of their attacks, they will be stunned for 3 seconds and will take double damage from all attacks. Finally, Once they drop bellow 75% HP, they will create 2 bloodbags around them and spend bloodfeast. They double the amount of bloodbags spawned after dropping bellow 25% HP.

Bleed: It is a stacking statues effect, which deal HP damage equal to it's stack after the target moves then it halves itself. It will also create pools of blood bellow the target if they have enough bleed. Also, If the target does not move for around 5-10 seconds and doesn't gain any new stacks of bleed while standing still, they will clear the stack completely.

Also, This PR updates most rooms in the backstreets, making them more enclosed in order to make them more safe.
## Why It's Good For The Game

Bloodfiends are more interactive foes that players can face in the backstreets, which should make the backstreets more fun. 

Also, rooms being more enclosed is useful since it makes exploring the backstreets more reliable.

## Changelog
:cl:
add: Added 3 new backstreets mobs, and a new statues effect "bleed"
add: Added new outfits, which the bloodfiends drop
reworked: changed how most rooms in the backstreets look
/:cl:


Kitsunemitsu:
I am actually looking to make bloodfiends an antag.

Not only this, I think tat this PR is good just unsure about spawning them like Gcorp and such

EnderMage99:
> I am actually looking to make bloodfiends an antag.
> 
> Not only this, I think that this PR is good just unsure about spawning them like Gcorp and such. It's something that would be nice for rooms and that bloodfiend template however

Okay, I can remove bloodfiends from spawning in like G-Corp.

Kitsunemitsu:
Honestly? If the La Manchaland map is merged, you can make it so that ONLY bloodfiends spawn there and just hard-map it

ModularTegustation/tegu_items/associations/cityspawners.dm
```diff
@@ -99,10 +100,15 @@ GLOBAL_VAR_INIT(city_east_enemies, FALSE)
 			spawning = /mob/living/simple_animal/hostile/ordeal/green_bot
 			if(prob(30))
 				spawning = /mob/living/simple_animal/hostile/ordeal/green_bot_big
-	new spawning(get_turf(src))
-
 
+		if("worms")
+			spawning = /mob/living/simple_animal/hostile/ordeal/amber_bug
+			if(prob(30))
```
Kitsunemitsu, line 106: These teleport around, which is why they have not been added.

EnderMage99: I gave them a variable which turns off their teleport during city games.

Kitsunemitsu: Ah, makes sense

Kitsunemitsu:
This PR does too much. I will request atomization between Bloodfiends and map updates (not ma lancha land), as map updates will be more scrutinized whereas bloodfiends are good to go.

Kitsunemitsu (approved):
Maps were removed. Looks good to me.

PR #2668, opened 2025-01-05
Recodes a good chunk of galaxy child's code to be cleaner and more efficient

Gboster-0:
## About The Pull Request

- Galaxy child now uses its own calculated delta_time to always have accurate healing per second
- Fixed a runtime with the child of galaxy cancelling its pebble gifting
- Added a lot of safety checks for inputs in child of galaxy's abilities to prevent long-distance pebbling
- Added a confirmation button to child's pebble breaking ability
- Galaxy child's abilities can now be used regardless of the map, but are only given during limbus labs.
- Galaxy child's abilities now have 5 seconds of cooldown instead of 60, technically a bug and not a balance change
- Galaxy child now activelly removes qdeleted and dead people from its friend list, alongside destroying their pebbles (but NOT reducing its healing/damage)
- Galaxy child's pebble giving and pebble breaking were made into procs for easier admins for admins/less re-used code for abilities.
- Replaces instances of Lazylist defines with normal ones

## Why It's Good For The Game

> Galaxy child now uses its own calculated delta_time to always have accurate healing per second
- Helps if we get a lot of lag on a shift for some reason
> Fixed a runtime with the child of galaxy cancelling its pebble gifting
- inputs can always return null, thats life
> Added a lot of safety checks for inputs in child of galaxy's abilities to prevent long-distance pebbling
- Im quite sure you could just telepathically ask someone to get a pebble 50 tiles away as long as you had the initial hook on them, thats not good
> Added a confirmation button to child's pebble breaking ability
- I feel like it could be pressed accidentally
> Galaxy child's abilities can now be used regardless of the map, but are only given during limbus labs.
- This makes it so admins can give the child of galaxy the abilities if they want to without the abilities refusing to work
> Galaxy child's abilities now have 5 seconds of cooldown instead of 60, technically a bug and not a balance change
- Yeaaaa
> Galaxy child now activelly removes qdeleted and dead people from its friend list, alongside destroying their pebbles (but NOT reducing its healing/damage)
- If we are holding a reference to them, that feels like it could very easily cause hard-deletes alongside un-necessary qdeleted human references that we have to sift through later when galaxy child inevitably breaches
> Galaxy child's pebble giving and pebble breaking were made into procs for easier admins for admins/less re-used code for abilities.
- Gotta love some cleaning
> Replaces instances of Lazylist defines with normal ones
- Yes i know LAZYLEN is the same as length(), but lazy defines are supposed to be used in lazy lists. This could confuse a developer when they see it.

## Changelog
:cl:
balance: child of galaxy now calculates its healing using delta_time, meaning its healing will ALWAYS be accurate regardless of lag
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/galaxy_child.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/galaxy_child.dm
@@ -1,5 +1,4 @@
 #define STATUS_EFFECT_FRIENDSHIP /datum/status_effect/display/friendship
-#define GALAXY_COOLDOWN (60 SECONDS)
 /mob/living/simple_animal/hostile/abnormality/galaxy_child
 	name = "Child of the Galaxy"
 	desc = "A young, lost child."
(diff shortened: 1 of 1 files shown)


InsightfulParasite (approved):
Code makes sense to me. Approve'd

PR #2678, opened 2025-01-13
Better brown dawn

Coxswain-Navigator:
## About The Pull Request
Reduces the HP of stage 1 peccatulum by roughly half across the board, reworks sloth/gloom to play better with other gamemodes and to make it a better fit for the later ordeals. Also contains some visual and audio updates for sloth and gloom peccatulum.

## Why It's Good For The Game
I actually have a plan for the harder versions of this ordeal now, and this is more in line with that plan. Also, this should be a bit more in line with the other dawns in terms of difficulty now.

## Changelog
:cl:
tweak: made sloth peccatulum fight a bit differently
balance: made the brown dawn ordeal a bit easier
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/brown/dawn.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/brown/dawn.dm
@@ -7,11 +7,11 @@
 	icon_living = "sinrock"
 	icon_dead = "sin_dead"
 	faction = list("brown_ordeal")
-	maxHealth = 200
-	health = 200
+	maxHealth = 80
+	health = 80
 	melee_damage_type = BLACK_DAMAGE
-	melee_damage_lower = 14
-	melee_damage_upper = 24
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/ordeal/brown/dawn.dm
```diff
@@ -241,27 +161,49 @@
 	guaranteed_butcher_results = list(/obj/item/food/meat/slab/human/mutant/slime = 1)
 	is_flying_animal = TRUE
 	ranged = TRUE
+	ranged_cooldown_time = 50 // Fires a laser dealing 10 white damage every 5 seconds
```
Gboster-0, line 164: Why not just include a define here if this is being timed
```suggestion
	ranged_cooldown_time = 5 SECONDS // Fires a laser dealing 10 white damage
```

PR #2679, opened 2025-01-13
LC13 Bleed Statues effect

EnderMage99:
## About The Pull Request

This PR adds the bleed statues effect, Which deals damage equal to it's stack each time the mob moves, then halves it. It deals * 4 more damage to mobs. If they have a high stack, they will also create blood piles bellow the target as they move. Also, If they stand still for 5 seconds, it will be automatically removed.

## Why It's Good For The Game

It will be useful for other PRs

## Changelog
:cl:
add: added new bleed statues effect
/:cl:


Kitsunemitsu (approved):
Approved!

PR #2687, opened 2025-01-20
Makes Lances Hit Multi-Tile Mobs

Halonexus:
## About The Pull Request
Makes lance charge attack hit the actual mob upon bumping into its blocker dummies.

## Why It's Good For The Game
Lances properly do damage when they're supposed to.

## Changelog
:cl:
fix: Lance charge attack can now damage multi tile mobs properly.
/:cl:


Gboster-0 (approved):
Works

PR #2688, opened 2025-01-20
In the name of Love and Hate healing is now a flat number

BLUREMOTE:
## About The Pull Request

QoH weapon now heal for flat amounts (10 HP, 10SP or 5HP/SP) instead of taking 20% or 10% of the projectile damage

## Why It's Good For The Game

If you got your hands on ways to increase the damage of the bullet (ranged class, wink wink), you could buff the healing output of the weapon. Healing stuff is balanced around their numbers for specific reasons, so increasing them would be funky. This also safekeep the healing in case something else further down the line can buff the damage of the projectile.


Kitsunemitsu:
I mean just because it's not intended doesn't mean we can't keep it

Gboster-0:
taking away the fix label, this more-so fits balance imo

Kitsunemitsu (approved):
Approving.

PR #2721, opened 2025-02-10
Backstreets Balance Update [DONE]

EnderMage99:
## About The Pull Request

This PR adds a few changes to the backstreets to make them more fair for players.

~**ROOM CHANGES**
All of the deadly rooms in the backstreets have been changed to have some sort of barricade between the players and the mobs, giving them a chance to see the threat, before jumping in to fight them. Also some rooms with especially deadly mobs have been given a return to origin component so those mobs don't go out and harass other players for one messing up.~ [REMOVED]

~**MOB CHANGES**
Many dawns in the backstreets are far too deadly for grade 9-8 fixers (green dawn can early take 45% of a fixers HP even while they are using healing skills), and with them being the most common common enemy types, this causes a lot of frustration from players due to them being kicked down while they are low. This issue is being fixed by reducing the damage of many dawn level mobs.~ [REMOVED]

**AMPULE CHANGES**
New players often don't know about how leveling up too fast can be a bad thing, since it can lock them out of skills. So, this update makes ampules give a warning to players if they are about to level up from using an ampule, giving them more information on when to use them.

One more changes which might be controversial is the nerf of fishing ampules. Now they can only be used until you are grade 8. At that point they will no longer work. I am making this change for 2 reasons. One, fishing has become a crutch for players who are too scared to fight in the backstreets. That way, it's difficult to balance the backstreets since players could be entering it at VERY different starting levels. Second, Fishing is a mechanic with no risk or gameplay. So it feels wrong when players are able to progress into the endgame effectively by being AFK. The best way of progression should be the most engaging one. Along with that, I am making fish no longer sellable. Since that would let the fishers to by pass this system and still level up from fishing.

~**FISHING CHANGES**
Instead of determining which fishing skills you can learn based on your attributes, the fishing skill books now check your fishing level instead. Along with that, fishing skills no longer counter towards the limited number of fixer skills you can have and have their own limits. (3 Level 1 Fishing Skills, 1 Level 2 Fishing Skill, and 1 Level 3 Fishing Skill *[All level 4 fishing skills have now become level 3 for convenience sake]*) Also, you are able to learn lower level fishing skills even if your fishing skill (the stat that grows as you fish) is very high. Lastly, there is a new cheap device which you can buy at the fishing market which lets you check your fishing skill.~ [REMOVED]

**PALE DAMAGE CHANGES**
Currently as it stands, Pale damage mods objectively suck. There is only a single mob where they are the best weapons for taking them down [Amber Dusk], and all of the Pale damage mods reduce the base force making them weaker then every weapon; and for what reward? Dealing more damage to players? I think that is very weak for a rare workshop mod, so I am changing all backstreets mobs pale resistance to be equal to their weakest resistance. That way, Pale damage can be seen as a all-rounder damage type. It will always be the second most effective damage type on all mobs, but always one step behind the best due to all of the pale damage mods reducing it's base damage.

**WORKSHOP WEAPON CHANGES**
The AoE in workshop weapons can now be toggled between targeting only hostile mobs (with will reduce their damage by 20%), and all mobs (letting them keep their original damage), letting them be used within teams. I feel like they are currently underpowered and anti-teamwork which is something I want to avoid. For that reason I am adding a toggle to let them work with a team, but if they are willing to put more effort into using this weapon in a team, they have an option.

## Why It's Good For The Game

Read the "About The Pull Request" section.

## Changelog
:cl:
tweak: changing backstreet rooms and mobs
balance: rebalanced fishing ampules
/:cl:


Kitsunemitsu:
Regarding mapping:
**Players that are not the best should die. Players should REQUIRE 2-3 other players with them.**

Making things easier by blocking off rooms encourages players to play alone. They should NOT be playing alone, and if they do, they should die. 
If they die as a team, then that's more people to encourage medical to pick up their corpses.

Kitsunemitsu (changes requested):
See previous comment.

Kitsunemitsu:
Speaking with this developer for this PR

Kitsunemitsu:
TBH, the only thing I'm worried about is a fishing change, can we atomize that so I can immediately merge this?

Kitsunemitsu (approved):
Approving. Mostly an update

PR #2734, opened 2025-02-27
Para's Devious Trap Update

InsightfulParasite:
## About The Pull Request
- ONE TILE TRIPWIRE TRAPS:
Tripwire traps are a structure ive just been considering for a while. Tripwire traps can be mapload or crafted.
If generated at mapload any /obj/item in their loc will be used as the attack. When Crossed() the item will be forcemoved to the turf and throw_impact will be called on tthe crosser. If a grenade is loaded into the tripwire trap it will arm the grenade. When a tripwire trap breaks it produces 1 cloth. This is to ensure a net loss in material if someone wants to mass produce them. Functionally they would be similar to Dwarf Fortress rock traps that drop a single rock item onto someone.
- PRESSURE PLATE AUTOHIDE: Pressure plates are items that are hidden under the floor like ss13 traitor smuggler satchels. When crossed they send a signal using the signaller system. I managed to make a subtype of pressure plates called "autohide" which hides them automatically upon spawn. 
- SYSTEM SIGNALLERS:
The only mechanic in ss13 that utilizes signalers to effect a naturally occurring event is ss13 anomalies. Those pyroclastic anomalies can be disabled with the correct combination of frequency and code on a signaler. How this worked was that each anomaly core inside a anomaly is a signaler subtype that calls disable on every anomaly on its turf when supplied with a signal. System signaler are like that but with projectile launchers. System signalers are also silent and do not beep when activated.
- PROJECTILE LAUCNHER:
A structure that fires a defined projectile in the direction its facing upon receiving a signal. 

image

## HOW TO USE SIGNALLER TRAPS
For a simple pressure plate signal trap that shoots a beanbag shotgun slug into a poor sods face you need to do the following in StrongDMM.
> place a pressurep_plate/autohide
> Pick a number above 100 and change the roundstart_signaler_code variable to the aforementioned number. 
> Place sigsystem/projectile_launcher and change its signaller_code to the pressure plates number.
> Leave it where fools tread.

## Why It's Good For The Game
It would be helpful for making custom traps in city maps instead of relying on contact mines. 

## Changelog
🆑
add: One Tile Tripwire Traps
add: projectile launcher
add: system signallers
add: backstreets large_south room "shotgun_exam" that utilizes a few of the new features. 
tweak: pressure plates
/🆑


Kitsunemitsu (approved):
Love it. Thanks for the simple system!

InsightfulParasite:
> Love it. Thanks for the simple system!

Ill do a follow up PR for a device that spawns atoms on signal. That seems like it would be fun.

PR #2739, opened 2025-03-01
Minor Cargo Update

Kitsunemitsu:
## About The Pull Request
Makes some minor changes to Cargo.
- Buying things with Cargo no longer lowers the quota. Just takes from available to spend boxes.
- Filtering PE boxes gives you a 30% chance for a double payout
- Filter cost has been halved.

## Why It's Good For The Game
Cargo as a mechanic right now is heavily underutilized system because it takes away from quota, and the blackjack minigame was underutilized. 
I believe that by encouraging players to use both Cargo and filters would be good for the game; two underutilized mechanics

This may require us to rebalance cargo in the near future.

## Changelog
:cl:
balance: Cargo no longer lowers the quota, just available boxes
balance: Filter cost has been halved, and filtering may double output.
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/refinery/refinery.dm
+++ b/ModularTegustation/tegu_items/refinery/refinery.dm
@@ -74,7 +74,7 @@
 		loaded = FALSE
 	else if(blackjack == 0)
 		timeleft -= round(refine_timer/3)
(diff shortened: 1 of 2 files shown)


InsightfulParasite (approved):
Looks good enough and Kirie is usually the moderator of Cargo Code.

PR #2743, opened 2025-03-02
Station traits Rework/Rebalance

Kitsunemitsu:
## About The Pull Request
Station Traits now only appear in 60% of all rounds instead of 89%.

-Legacy Pale Trait has been removed.
-Selection Mishap has been removed. While some people like this one, and it's not as hated, it's quite polarizing so it gets the axe
-Abno Blitz weight has been reduced from 5 to 3 weight. It's very polarizing so it's important to keep it low. Some people love it, and some people hate it.
-Joke Abnormalities now Guarantees alternate joke sprites
-Wounds should actually works in the callback trait. This trait is also rather polarizing so it may be removed.
-Working clerks had it's rate reduced from 3 to 2
-Added experimental departmental agents station trait, people have been bitching about it for forever, so testing it as a trait is a good idea. May or may not stick around

## Why It's Good For The Game
A small rebalance of Station Traits.
Initial feedback is generally positive, however the system as a whole needs changing to have players like it more.

Removing less liked traits, and making tweaks to other traits.

## Changelog
:cl:
balance: rebalanced station traits
/:cl:


code/modules/mob/living/simple_animal/abnormality/_abnormality.dm
```diff
@@ -177,6 +177,9 @@
 	if(secret_chance && (prob(1)))
 		InitializeSecretIcon()
 
+	if(SSmaptype.chosen_trait == FACILITY_TRAIT_JOKE_ABNOS && secret_chance)
+		InitializeSecretIcon()
```
Gboster-0, line 181: Tbh i think this would be cleaner
```suggestion
	if(secret_chance && (prob(1) || SSmaptype.chosen_trait == FACILITY_TRAIT_JOKE_ABNOS))
		InitializeSecretIcon()
```

InsightfulParasite (approved):
Looks good to me. Approved'd

PR #2745, opened 2025-03-02
Balances Quirk Points

tile4:
## About The Pull Request
Changed how many points some quirks gave, according to how impactful they are in LC13.

## Why It's Good For The Game
Some quirks costed points that were harmless, and others were too cheap for what they did. A balancing doesn't hurt.

## Changelog
:cl:
balance: balanced quirk amounts
/:cl:

### DIFF
--- a/code/datums/quirks/good.dm
+++ b/code/datums/quirks/good.dm
@@ -4,10 +4,10 @@
 /datum/quirk/resistant
 	name = "Damage Resistant"
 	desc = "You take 5% less damage from all sources."
-	value = 4
+	value = 6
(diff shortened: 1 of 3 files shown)


Gboster-0:
Don't really like this tbh, since brain tumor can be healed with a lobotomiser/pills if you are mid-combat.
Parapalegic can actually be perma-cured by having new legs stapled onto you with surgery
100% agree with RDS point nerf tho, its cost was from when doors had a chance to hallucination shock you

Coxswain-Navigator (approved):
I gave it a quick lookover, it seems fine to me.

PR #2746, opened 2025-03-03
Nobody Is Improvements/Nothing there fixes

Coxswain-Navigator:
## About The Pull Request
This PR adds the new CopyHumanAppearance() Proc, which creates a dummy mob with the same appearance as a target then copies that dummy. This prevents things like overlays from panic halos or buffs from making the copied appearance look weird. This proc can also be used for other things in the future. Also fixes some major bugs.

TL;DR VERSION

- Fixes weirdness when Nobody Is Shells someone
- Fixes the strangle attack dealing negative damage, and thus healing people.
- Properly implements the voice copying feature I never actually finished.
- Fixes pathing issues and NI occasionally derping out and not doing anything.
- Fixes Nothing There being invisible if someone somehow actually gibs while being turned into a shell
- Slightly changes examine text on humans used as shells.

## Why It's Good For The Game
No more Oberon running around with a giant holographic brain floating over his head, being invisible, and other sprite shenanigans. Hopefully.

## Changelog
:cl:
fix: fixed Nobody Is sprite stealing issues
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/aleph/nobody_is.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/aleph/nobody_is.dm
@@ -56,7 +56,6 @@
 
 	//Contained Variables
 	var/reflect_timer
-	var/mob/living/disguise = null
 	var/shelled = FALSE
 	var/mob/living/carbon/human/chosen = null
 	var/can_act = TRUE
@@ -189,12 +188,12 @@
 		HD.update_limb()
 		chosen.hairstyle = oldhair
 		chosen.facial_hairstyle = oldbeard
-	HD.update_limb()
 	headicon = new(get_turf(src))
 	headicon.add_overlay(HD.get_limb_icon(TRUE,TRUE))
 	headicon.pixel_y -= 5
 	headicon.alpha = 150
 	headicon.desc = "It looks like [chosen] is reflected in the mirror."
+	HD.update_limb()
 	//Handles connected structure part
 	datum_reference.connected_structures = list(headicon = list(0,-5))
 
@@ -232,7 +231,7 @@
 	datum_reference.qliphoth_change(-1)
 
 /mob/living/simple_animal/hostile/abnormality/nobody_is/BreachEffect(mob/living/carbon/human/user, breach_type)
-	if(current_stage > 1)
+	if(!(status_flags & GODMODE) && breach_type != BREACH_MINING) // Already breaching
 		return
 	if(reflect_timer)
 		deltimer(reflect_timer)
@@ -241,7 +240,7 @@
 		return
 	CheckMirrorIcon() //Clear overlays
 	next_stage()
-	if(breach_type == BREACH_MINING)
+	if(breach_type == BREACH_MINING) // Keeps it from teleporting off when uncovered
 		return
 	// Teleport us somewhere where nobody will see us at first
 	var/list/priority_list = list()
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/abnormality/aleph/nobody_is.dm
```diff
@@ -576,7 +578,7 @@
 			grab_victim.deal_damage(strangle_damage, BLACK_DAMAGE)
 		else	//Apply ramping damage
 			playsound(get_turf(src), 'sound/effects/wounds/crackandbleed.ogg', 200, 0, 7)
-			grab_victim.deal_damage((strangle_damage * (3 - count)), BLACK_DAMAGE)
+			grab_victim.deal_damage((strangle_damage * (count - 3)), BLACK_DAMAGE)
```
InsightfulParasite, line 581: Will this cause the overall damage to be in the negative? because if so you may want to use the CLAMP proc to keep it above 0.

InsightfulParasite:
Does this fix https://github.com/vlggms/lobotomy-corp13/issues/2747 ?

Coxswain-Navigator:
> Does this fix #2747 ?
No, I made the issue because I couldn't figure out how to fix it.

code/__HELPERS/icons.dm
```diff
@@ -1328,3 +1328,54 @@ GLOBAL_LIST_EMPTY(transformation_animation_objects)
...
+
+	else
+		//even if target isn't a carbon, if they have a client we can make the
+		//dummy look like what their human would look like based on their prefs
+
+		target?.client?.prefs?.copy_to(copycat, icon_updates=TRUE, roundstart_checks=FALSE, character_setup=TRUE)
+	SLEEP_CHECK_DEATH(1)
+	appearance = copycat.appearance
+	qdel(copycat)
+	return
```
Gboster-0, line 1381: don't put return's at the end of procs
```suggestion
```

code/__HELPERS/icons.dm
```diff
@@ -1328,3 +1328,54 @@ GLOBAL_LIST_EMPTY(transformation_animation_objects)
 	final_color_code[2] = FLOOR(final_color_code[2]/total, 1)
 	final_color_code[3] = FLOOR(final_color_code[3]/total, 1)
 	return "#[num2hex(final_color_code[1], 2)][num2hex(final_color_code[2], 2)][num2hex(final_color_code[3], 2)]"
+
+/*
+ * Make a copy of a human and copy their appearance without copying any special overlays. Notice that this currently doesn't include held items!
+ */
+/mob/proc/CopyHumanAppearance(mob/target)
```
Gboster-0, line 1335: Why is this a proc on the mob level? Is there ANY situation where ghosts/undead mobs should be using this
If not, i'd recommend putting it on the abnormality zone

Gboster-0 (commented):
Tbh this feels a bit wastefull as a concept, though i quess it could work

code/__HELPERS/icons.dm
Coxswain-Navigator: Probably not, but there isn't really a good reason to put it at the abnormality level either. There's always adminbus, or if someone wants to make a distortion that is some sort of doppelganger. I'm more concerned about how it currently  doesn't grab held items, but for the purpose of this PR specifically, that doesn't matter.

Gboster-0: Eh, if someone wants to make an abno that uses this they can always bump it down a level, imo this is unnecessary

Coxswain-Navigator:
Discovered some major bugs during live testing - will leave this on "draft" until everything is ironed out, but this isn't a PR that is easy to test locally.

Coxswain-Navigator:
All of the previously mentioned bugs in this PR have been fixed either directly or indirectly.

Kitsunemitsu (approved):
Mostly a bugfix.

PR #2755, opened 2025-03-10
Fix Snow Queen Sword being Stuck in Belts and Doorstuck

InsightfulParasite:
## About The Pull Request
Fixes snow queen sword being stuck in belt slots.
The issue came from it being a cursed weapon and retaining its storage slots. So i set the storage slots to null. Preemptively adds this fix to hammer of light to avoid any bugs since it is also a cursed weapon. Also fixed another bug where if your standing in the snow queen door while it closed you would be trapped forever.

## Changelog
:cl:
fix: Snow Queens Sword and Hammer of Light being put in belt slots despite being Cursed
fix: Snow Queens door trapping people forever if it closes on them.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock_types.dm
+++ b/code/game/machinery/doors/airlock_types.dm
@@ -572,7 +572,7 @@
 /*
 	LC13 Airlocks
 */
-
+//One day ill replace this with a structure similar to the necropolis gate. -IP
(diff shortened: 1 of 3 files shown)


Gboster-0 (approved):
Looks good

PR #2764, opened 2025-03-15
Tuning up Thumb firearms

41adushek:
## About The Pull Request

- Buffs up the entirety of CoL variant Thumb guns to deal a bit more damage.

## Why It's Good For The Game

- Soldatos, Capos and Sottocapo are now bigger of a force to both low to mid grade fixers and a considerable advesary to high grade fixers. As the Thumb should be for a SotC level threat.

## Changelog
:cl:
bal: soldato rifles deal 18 RED instead of 10 RED.
capo rifles deal 35 Red instead of 20 RED.
sottocapo shotgun deals 11 RED instead of 7 RED, but with a drawback in form of reload time (7 seconds instead of 5 seconds)
tweak: sottocapo shotgun is a medium weight weapon now (can be used one handed, can't be sheathed into EGO belts) to make up for being a handgun.
/:cl:

### DIFF
--- a/ModularTegustation/ego_weapons/melee/non_abnormality/thumb.dm
+++ b/ModularTegustation/ego_weapons/melee/non_abnormality/thumb.dm
@@ -44,11 +44,12 @@
 //Sottoacpo
 /obj/item/ego_weapon/ranged/city/thumb/sottocapo
 	name = "thumb sottocapo shotgun"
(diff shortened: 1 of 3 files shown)


41adushek:
i screwed up on the PR part, need to redo it the RIGHT way

PR #2766, opened 2025-03-16
Para's Devious Trap Update: Tile Remake [DONE]

InsightfulParasite:
## About The Pull Request
Sequel to https://github.com/vlggms/lobotomy-corp13/pull/2734 this entirely reworks the system to be more reliant on directional tiles.

Most structures have been made invisible to players and less reliant on signallers.
Im sorry if i explain these poorly.

-Signal Tower-
image
Signal towers receive pings from other towers. When you ping a signal tower it checks a static list of all existing towers which should max be around like 20. When the signal tower is pinged by another tower it will do a directional ping to whatever is in front of it. Main purpose of signal towers to facilitate communication between visible components or across long distances.

Keypads, buttons, and and_gates have been remade into subtypes of the sigsystem/tower so they can easily send out and recieve signals. Unless they are keypads and buttons then they don't receive signals.

-Connectors-
image
Function as simple directional tiles that ping in a line. They have variables to change them into delay, repeat, or timers.
Can be stacked for multiple directions but if prudent there is a split subtype.

-Activators-
image
Activators are placed under sigsystem crates or shutters to unlock or open the structure. These can be stacked on top of connectors to allow a seamless flow.

-Signal Nodes-
image
Functions as the inbetween of non sigsystem structures like the crate and sigsystems. If a crate wanted to send out a signal it would have to have a node underneath it.

-Spawners and Deployers-
image
image
Function similarly to how they did before this PR, instead spawners will spawn their atom and deployers will fire their projectiles upon receiving a ping from another tile.

-Pressure Plates-
image
Work like other types pressure plates where they send out a ping in the direction of the gear when crossed.

-Signaller-
image
A horrid combo of 2 signallers and a tile. Work similarly to radio towers but instead of pinging the tower list they instead ping with signallers.
PS: Only exists because a previous map i made is reliant on signallers so it kind of bridges systems.

image
In the above image a sigsystem chest sends a signal to the two nodes underneath it when it opens. One node activates a announcer tile while the other node is sent to two spawners and a door opening device. The delay and repeat variables of some of the connectors has been altered to make the spawned enemies appear one at a time. The second connector leading into the device activator in the top left is delayed for 3 minutes since it takes 3 minutes for all 6 of the enemies to spawn.
image

## Why It's Good For The Game
Allows mapmakers to be more complex and devious possibly leaving their entire room inert until someone enters.

## Changelog
:cl:
add: sigsystem atom spawner
add: sigsystem keypad & and-circuit
add: sigsystem button
add: 4 demostrative backstreets rooms
/:cl:


EnderMage99:
I have an idea for how to use it for my next away mission, it will be fun!

PR #2776, opened 2025-03-21
Pinocchio is now immune to glass shards

Anonmare:
## About The Pull Request
Pinocchio will no longer take damage from walking on glass shards. Technically doesn't need the shoes but I won't steal his drip from RCA. Also prevents him from leaving footprints from decals but that's got no relevant balance effects.

## Why It's Good For The Game
You gotta at least shove Pino against the wall yourself if you want to bully him, don't outsource your labour to broken glass.

## Changelog
:cl:
balance: Pinocchio no longer takes damage from glass shards
/:cl:


Kitsunemitsu (approved):
Approved. Easy good buff

PR #2782, opened 2025-03-25
City Update: Justice Nerf/BL Buff

EnderMage99:
## About The Pull Request

This PR makes all melee weapons no longer get the justice buff when dealing damage to humans in City Gamemodes
Also, This buffs all of the blade lineage weapons by making their empowered attack deal 2% more damage per 1% HP they have.

## Why It's Good For The Game

For the gamemodes where PVP is most likely to happen, it ends rather quickly. And with how damage on weapons accounts for dealing with mobs which have over 750 HP on average (At high levels when PVP happens), while humans are only able to have a max of 260 HP, causes the weapons to be overturned for PVP. This nerf should make PVP more interesting, since players will have time to think about their actions, instead of everything ending under 3 seconds.

Also, With these buffs it would make BL very weak since they relay on ending fights in one hit, So this buff should help them deal more damage. Along with the justice nerf, it will let them take more hits before going down.
## Changelog
:cl:
balance: rebalanced weapons in city gamemodes
/:cl:


Kitsunemitsu (changes requested):
Spoke with the developer about how to make this less strange and gross.

Kitsunemitsu:
Also please buff BL weapons to deal double damage if this goes through, they should be able to one shot people.

Kitsunemitsu:
Requires a TM, but it will be done.

PR #2787, opened 2025-03-26
Status Effects+

EnderMage99:
## About The Pull Request

This PR will add the following status effects:
_All of the following status effects work on both mobs and humans._

- [x] Protection / Fragile: Reduces or increases all damage taken from all 4 damage types by (10% * stack). (Max stack of 10).

- [x]  (Damage Type) Protection / (Damage Type) Fragile: Reduces or increases all damage taken from a specific type by (10% * stack). (Max stack of 10).

- [x]  Damage Up/Down: Increases or decreases your melee damage by (10% * stack). (Max stack of 10).

- [x]  (Damage Type) Damage Up/Down: Increases or decreases your melee damage by (10% * stack) if it is the same as it's damage type. (Max stack of 10).

- [ ]  Rupture: When you take RED or BLACK, it causes you to take (stack) HP damage and then reduces the stack by half. (Deals 4x more damage to mobs), (After not getting inflicted with any new Sinking for 10-20 seconds, remove all Rupture), (Max stack of 50).

- [ ]  Sinking: When you take WHITE or BLACK, it causes you to take (stack) SP damage and then reduces the stack by half. (Deals 4x more damage to mobs), (After not getting inflicted with any new Sinking for 10-20 seconds, remove all Sinking), (Max stack of 50).

- [ ]  (Rupture/Sinking/Burn/Bleed) Stasis: When the following status effect would be reduced through natural decay, it does not reduce the stack, and instead this status effect is reduced by 1. (Max stack of 10.)

_More status effects will be added to this RP if I come up with them..._

## Why It's Good For The Game

More status effects will make more things consistent and can be easily used when creating new weapons/mobs.

## Changelog
:cl:
add: Added new status effects
/:cl:


Coxswain-Navigator:
Rupture and sinking already exist, technically.

EnderMage99:
> Rupture and sinking already exist, technically.

Well, Is it the Rupture and Sinking that is outlined in [LC13 2025 Development]?

Coxswain-Navigator:
> > Rupture and sinking already exist, technically.
> 
> Well, Is it the Rupture and Sinking that is outlined in [LC13 2025 Development]?

No... however, I am fairly certain that my reedition of sinking and rupture has our Headercoder's approval and went through quite a bit of refinement which makes it quite different from the ideas in the log. I'll need to talk to her later about closing that particular objective since it has already been completed.
For the record, the current implementation is brute damage for rupture, and brute damage that applies to sanity for sinking. So, basically fixed damage without an associated status effect because making it stack up doesn't really make a whole lot of sense.

PR #2788, opened 2025-03-26
LCL HQ has removed specimen safety suppression allowing for higher abnormality aggression between each othe

KingSuperBOB:
## About The Pull Request
This PR makes these specific abnos get their own faction when sentient, what this does is let them use their faction based attacks on each other so they dont die brutally when killing each other.

## Why It's Good For The Game
More opportunities to RP between abnos, also abno death battles.

## Changelog
:cl:
tweak: gave every abno in LCL with a faction related attack their own faction when sentient
/:cl:


Kitsunemitsu (approved):
Holy shit someone actually using the abno overwrites file as they should

PR #2812, opened 2025-04-03
Adventure Console Combat Update

Kitsunemitsu:
# About The Pull Request

## COMBAT UPDATE
Now you can upgrade all 4 different commands, and 5 stats.
I'm hoping that this will allow players to actually create builds!
### HP 
Your HP can now go up to 150.

### Attack
More Dice options. Base Dice nerfed from 1d6 to 1d4. 
In order of weakest to strongest dice:
1d6 > 3d2 = 1d8 > 2d4 > 1d10 > 1d12 > 2d6 > 3d4

### Block
Block has been changed from a "Lose slower" option to a "Gamble to heal button"
Block Dice - Block dice is now upgraded independently from attack dice. 1d4 > 1d6 > 1d8 > 1d10 >1d12.
Upgrading block dice is cheaper than attack dice
Block Healing - Blocking can now heal you. Starts at 0, and heals up to the Max Block Healing. Upgradable
Max Block Healing - Max HP that blocking can heal you.  Upgradable to 30 Max per combat.

### Run
Run is now a more viable gameplay option.
Because of how safe running can get, upgrades can get SUPER pricey at high levels. 
A basic run build should be cheap, but a high level run build will be super strong.
Run Chance - Run chance is upgradable up to 70%. Starts at 30%.
Run Healing - Run Healing now starts at 0, but can go up to 10 HP.
Run Gold Chance - The chance of getting gold upon running. This means that you cannot get more than one gold per combat,
but at high run chances you'll just be breezing through it.

### Info/Look
Info is now an excellent support option with reason to do.
Info is now unsuccessful if you are attacked during it.
Info Chance - The chance of getting info. Starts at 30% and goes up to 70%.
Info Bonus - Increases the amount of gold you get AFTER killing the monster. Goes up to a bonus 3 gold per encounter.

### Extra
More content - You can now access all abnos by paying a fee of 20 Coin

## SHOP UPDATE
All the garbo from the shop was removed, and replaced with a set of actually usable stuff. 
Prices were all decreased.

### More flavour! More minor changes and lines of text!

# Why It's Good For The Game
Adventure console combat was pretty bland before. I'm hoping that with this, I can create multiple different builds for players to tinker with! Eventually I want choosable difficulties and more!

## Changelog
:cl:
balance: Overhauls the Adventure console combat system
balance: Rebalanced the Adventure Console Shop
/:cl:

### DIFF
--- a/ModularTegustation/lc13_obj/_adventure_console/adventure_layout.dm
+++ b/ModularTegustation/lc13_obj/_adventure_console/adventure_layout.dm
@@ -12,14 +12,16 @@
  */
 
 /datum/adventure_layout
-	///Current players health
+	///Current and max player health
+	var/max_integrity = 100
 	var/virtual_integrity = 100
 	///Coins for flipping.
 	var/virtual_coins = 0
 	///Event progress so that your not stuck with nothing but null events.
 	var/program_progress = 0
-	///Damage in battle senarios. Based on dice values.
-	var/virtual_damage = "1d6"
+	///Damage and block in battle senarios. Based on dice values.
+	var/virtual_damage = "1d4"
+	var/virtual_block = "1d4"
 	///Variable for when they last logged in. Used in StatCatchup()
 	var/last_logged_time = 0
 	///What we are currently displaying.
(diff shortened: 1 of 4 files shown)


InsightfulParasite:
These are good changes.

ModularTegustation/lc13_obj/_adventure_console/adventure_layout.dm
```diff
@@ -429,14 +559,68 @@
...
+			info_chance = value
+
+		if("INFOBONUS")
+			if(value < info_bonus)
+				return FALSE
+			info_bonus = value
+
+		//Extra
+		if("ABNOS")
+			all_abnos = TRUE
```
InsightfulParasite, line 623:
```suggestion
			if(all_abnos)
				return FALSE
			all_abnos = TRUE
```
This is to prevent people buying extra content twice. I know its a bit silly but it prevents complaints.

ModularTegustation/lc13_obj/_adventure_console/adventure_layout.dm
```diff
@@ -456,7 +640,7 @@
 	var/list/new_events = events - event_options
 	for(var/_option in new_events)
 		var/datum/adventure_event/A = _option
-		if(initial(A.require_abno))
+		if(initial(A.require_abno) && !all_abnos)	//Do you require an abno and do you NOT have all abnos?
```
InsightfulParasite, line 643: Im going to look at this a bit more because this should work to make all events appear.

InsightfulParasite: I checked the code and this does work to unlock all events. It may not be obvious at first though due to there being 30+ events.

InsightfulParasite (approved):
All good and done! Approved!

PR #2813, opened 2025-04-03
City-Corp, a Map to replace Xicorp and Cityfixers

Kitsunemitsu:
## About The Pull Request
A new version of the City-Corp hybrid! 
Featuring: 
Cut down roles on both sides
Small, TRADITIONAL LC13 map
Small city area that's easy to move around in.
New role list to replace the flawed WL one.

To do:
Wiki :skull: https://wiki.lc13.net/view/City-Corp

## Why It's Good For The Game
I don't really care for Xicorp because players often get lost.
and Cityfixers gets constant complaints.
This should be good for both LC13 and City players.

## Changelog
:cl:
add: Added City-Corp, a new Hybrid map
/:cl:


InsightfulParasite (approved):
Cursory review seems like its okay. Approved.

PR #2826, opened 2025-04-06
R-Corp Balance update

Kitsunemitsu:
## About The Pull Request
Raccoon sniper shots deal more damage and are faster
Raven Supports get grenades.
Reindeer beams heal faster.

Caves abnos are now preset to exactly the same:
2 Combat
2 Range
1 Tank

One map had like 4 fucking tanks and 3 more abnos than the others.

## Why It's Good For The Game
R-Corp has lost literally every every round in the last like 5 sessions.
I'm trying to correct this.
Generally, 5th pack does better than 4th pack.

## Changelog
:cl:
balance: updated Rcorp to be more effective
/:cl:


code/modules/mob/mob.dm
```diff
@@ -713,6 +713,11 @@
 			to_chat(usr, "<span class='boldnotice'>You must wait for [round((mind.respawn_cooldown - world.time)/10)] more seconds before respawning!</span>")
 			return
 
+	if(SSmaptype.maptype in SSmaptype.combatmaps)
+		if(!(SSmaptype.maptype in SSmaptype.citymaps))
```
EnderMage99, line 717: Can you also add a check for LCL? That is a gamemode that also needs respawns, due to how people will often ghost at the start for a chance at becoming an abno.

EnderMage99 (changes requested):
This PR is almost ready to be merged, there is only one issue that I see with respawns.

ModularTegustation/tegu_items/rcorp/landmarks.dm
```diff
@@ -65,7 +65,6 @@ GLOBAL_LIST_INIT(rhinobuster, list(,
 	/mob/living/simple_animal/hostile/abnormality/rudolta,
```
EnderMage99, line 65: Also, I found what causes the error which prevents a rhinobuster from spawning. It is the leading comma on line 63. If you could fix it as well, that would be helpful.

PR #2831, opened 2025-04-07
Fixes a issue with Zwei West, adds it to roaming fixers.

KingSuperBOB:
## About The Pull Request
Players would use the weapon like a spear to trigger its guard, so the weapon doubles its attack speed and immobilizes without any of the benefits if the player fails to dash, also adds a Zwei West option to roaming association.

## Why It's Good For The Game
Roaming association let's players try out Zwei West and the weapon change should stop the issue where its treated like a high DPS spear.

## Changelog
:cl:
add: Added Zwei West gear to roaming assoc at an uncommon rate
fix: fixed zwei west being able to trigger its lunge without lunging
/:cl:


Kitsunemitsu (approved):
Mostly a bugfix, adding it to Roaming is a nice touch.

PR #2839, opened 2025-04-10
Official Offices [DONE]

EnderMage99:
## About The Pull Request

This PR adds a new tool to Hana called the **"office marker"**. The office marker lets Hana group up players into office, which gives those players the ability to use a new n corp training accelerator, the **"office n corp training accelerator"**

The office n corp training accelerator can only be used by fixers in an office, and causes all other humans within a 9x9 range who share the same office to gain attributes as well. You are able to buy office n corp training accelerators in the same place where you can buy the normal versions, at the same cost. (I am keeping it like this to further encourage offices, this can be raised for balancing.)

Lastly, this update adds **"office scanner gates"**. They spawn in the same location where the "T-Corp locking mechanism" spawn. (This is to make TMing this PR easier, no map conflicts. Also they don't spawn in Office Fixers) For roles that can use n corp training accelerators, they will be alerted of the danger of soloing if they are not a part of an office..

## Why It's Good For The Game

_Currently as the head coder said many times, offices are a core gameplay mechanic and needs to be further enforced. The office scanner gates should be the most simple to understand method of preventing solo players from entering the backstreets. As Tile4 said, if we don't want solo play, we should make it clear to the players. Stuff like bounty hunters and SP drain when alone could confuse players on why they are being punished._ [REMOVED]

 Also, while offices are quite strong with their team work, they require much more ahn to function when compared to solo players. While it is okay that they need more ahn since there are more players, we should reduce the amount to make them more favorable to players. The office accelerators should fulfill that job.

## Changelog
:cl:
add: Added office marker, office n corp training accelerators and office scanner gates
/:cl:


Kitsunemitsu (approved):
Approved!

PR #2842, opened 2025-04-11
Movement smoothness improvement and diagonal move speed

azenkov:
## About The Pull Request

Quality of live improvements. Making client more smooth and making diagonal movement speed the same as cardinal.

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -118,8 +118,13 @@
 		return FALSE
 	//We are now going to move
 	var/add_delay = mob.cached_multiplicative_slowdown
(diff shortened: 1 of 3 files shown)


code/modules/mob/mob_movement.dm
```diff
@@ -140,7 +145,7 @@
 	. = ..()
 
 	if((direct & (direct - 1)) && mob.loc == n) //moved diagonally successfully
-		add_delay *= 2
+		add_delay *= sqrt(2)
```
Kitsunemitsu, line 148: I heavily dislike euclidean movement and think it feels very strange. I will cut the middle at 1.7-1.8x, but 1.41 is too strong for drifting around corners.

Kitsunemitsu (approved):
Made the minor change, going to merge.

PR #2852, opened 2025-04-14
Adventure Console 2.1

InsightfulParasite:
## About The Pull Request
Adventure Events now have new Variables
var/spend_event is false by default. If a option in the event changes it to TRUE then it will no longer be encountered after that option.
var/force_encounter removes all other path options when this event is generated.
Adventure Console now has a new Variable
var/list/spent_events is a list of events that cannot be encountered again. This comes into play when events are being generated. Events are added to this list if completed with the spend_event variable as TRUE.
Adventure Console will now return players to the adventure screen instead of the main menu.

Event Edits:
As a example, legacy.dm has force_encounter and spend_event turned to TRUE.

## Why It's Good For The Game
Requested features by writers and players.

## Changelog
:cl:
add: Adventure Console Support for One Time and Forced Encounters.
tweak: Exiting events in adventure_console now sends the player back to adventure screen. The screen confirming that you want to return to the paths however still remains.
/:cl:

### DIFF
--- a/ModularTegustation/lc13_obj/_adventure_console/adventure_events/_event.dm
+++ b/ModularTegustation/lc13_obj/_adventure_console/adventure_events/_event.dm
@@ -15,6 +15,8 @@
 	var/extra_chance = 0
 	//Short lived text that says things that had happened.
 	var/temp_text = ""
+	//Determines if the event can only be encountered once.
+	var/spend_event = FALSE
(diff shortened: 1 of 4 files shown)


InsightfulParasite:
Testing shows that this works so far.

Kitsunemitsu (approved):
This is important for a PR I am working on, and works.

PR #2865, opened 2025-04-20
Armor datum cleanup + silk upgrading improvement.

Halonexus:
## About The Pull Request
setRating proc on armor datums was pushing red armor into melee, white into bullet and so on, due to the inconsistent way the color damage types were added. I moved the color types to the start of the arglist the same way mob damage coeffs are already done.
Armor datums should not be manually created or modified as they use some kinda caching system to share the same datum instances based on armor tag, so I replaced all the places where they were created with getArmor instead.
The way silks modified armor was occasionally directly modifying armor datums, so I rewrote it to avoid potential armor pitfalls.
Additionally silks now wont downgrade armor to the MAX_ARMOR when the original armor was already better(80 + 5 wont become 40 anymore).
Made the new armor text a bit more readable.

## Why It's Good For The Game
Should fix some potential jank and makes silk application a bit better.

## Changelog
:cl:
tweak: Silk wont downgrade armor unintentionally.
tweak: Made new armor message after silk application more readable.
code: Moved color damage parameters to the start of armor datum arglist for all armor procs.
code: replaced armor datum initializations with getArmor calls.
/:cl:


Kitsunemitsu (approved):
Simple fix.

PR #2880, opened 2025-04-27
Nerfs L Corp Disciplinary Gear

tile4:
## About The Pull Request
Nerfs the EGO shard armor along with the strong EGO shard damage.

## Why It's Good For The Game
L Corp Disciplinary Gear as is is either about as good as the other city gear or better. When you get something that remains conistent between rounds, it shouldn't be as strong as the other options. And not only that, but agents use some of the disciplinary armor since the armor gave the same resistance as EGO as the highest risk level in the facility that Agents can train to, so this trivializes abnormality selection as a result.

## Changelog
:cl:
balance: rebalanced L Corp Disc Gear
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/disciplinary/_disciplinary.dm
+++ b/ModularTegustation/tegu_items/disciplinary/_disciplinary.dm
@@ -10,7 +10,7 @@
 	var/base_damage = 30 //Base damage of the tier
 	var/tier = 1 //used to figure out gun damage
 	//Armor stats
-	var/red_bonus = 20 //50 from the base of 20 in red, so 70
+	var/red_bonus = 10 //40 from the base of 20 in red, so 60
 	var/white_bonus = 10
 	var/black_bonus = 10
 	var/pale_bonus = 10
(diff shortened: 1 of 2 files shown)


Coxswain-Navigator:
How are agents getting enough of this gear to outfit themselves? There should only have ever been enough to gear up the occasional ERA to begin with. It might be fine to nerf them slightly though.

tile4:
> How are agents getting enough of this gear to outfit themselves? There should only have ever been enough to gear up the occasional ERA to begin with. It might be fine to nerf them slightly though.

Combat Pages are an easy source of PE. A level 1 page is enough for a Tier 2 Shard alone.  A level 2 page is more than enough PE for one of the Tier 4 Weapons or Armor.

Kitsunemitsu:
> Combat Pages are an easy source of PE. A level 1 page is enough for a Tier 2 Shard alone. A level 2 page is more than enough PE for one of the Tier 4 Weapons or Armor.

Combat pages are being reworked/removed.

tile4:
> > Combat Pages are an easy source of PE. A level 1 page is enough for a Tier 2 Shard alone. A level 2 page is more than enough PE for one of the Tier 4 Weapons or Armor.
> 
> Combat pages are being reworked/removed.

Well still, it kinda feels wrong that they are so strong yet are in every single round, And considering a good amount of rounds don't have many suppression agents, there usually is enough PE for them to have such gear consistently.

tile4:
This PR wouldn't exactly be needed since that quota bug is fixed and the combat pages are being removed. So I would say this PR is for an outdated problem.

PR #2882, opened 2025-04-28
Actually enables Sirocco

JackFrost2424:
## About The Pull Request

I'm like 90% certain that Sirocco has been unpickable for the last month because the origin was community and most of the gamemodes don't select community abnos. So. You know. Heeheehoohoo.
It got turned into a main mode abnormality so that's an issue.
That or I'm completely wrong and somehow no one has voted for it ever in over a month. Which I think is statistically unlikely.

## Why It's Good For The Game

Having the abno actually pickable is, you know, good.

## Changelog
:cl:
fix: enables Sirocco
/:cl:


Kitsunemitsu (approved):
Bugfix

PR #2884, opened 2025-04-28
Fixes fake shuttle call announcement on shuttle purchase

Destrok171:
## About The Pull Request

Attempting to purchase a shuttle called the autoEvac() proc in subsystem/shuttle.dm. This proc's intended purpose seems to be to be called on the destruction of a comms console or AI (from actual base SS13), and then it checks to see if any other communications consoles or AIs are alive - if there aren't, it sends an emergency shuttle request.

I do not know why autoEvac() gets called by purchasing the shuttle. I also don't know why the shuttle request from autoEvac() doesn't actually call the shuttle.

This PR only adds a check to see if we're on an LC13 map, Corp City or City of Light, and prevents the shuttle from being requested by this proc if that is the case.

I have tested this on Facility Alpha and on City of Light and it seems to work fine. I didn't test it on Corp City because I can't find it in the list of maps on my local server for some reason.

## Why It's Good For The Game

People will stop getting confused by purchasing a shuttle and thinking it automatically calls evac.

## Changelog
:cl:
fix: Purchasing a shuttle no longer creates a shuttle called message
/:cl:

### DIFF
--- a/code/controllers/subsystem/shuttle.dm
+++ b/code/controllers/subsystem/shuttle.dm
@@ -336,6 +336,9 @@ SUBSYSTEM_DEF(shuttle)
 			callShuttle = FALSE
 			break
 
+	if(SSmaptype.maptype in SSmaptype.lc_maps || SSmaptype.maptype == "lcorp_city" || SSmaptype.maptype == "city")
+		callShuttle = FALSE
+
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (approved):
Fix PR

PR #2885, opened 2025-04-28
Flesh Idol Buff

tile4:
## About The Pull Request
Adjusts how Flesh Idol works, to have a qliploth of 4 that drops from each work, but will increase every 5 minutes that pass.

## Why It's Good For The Game
Makes Flesh Idol more fair since unavoidable breaches can be a bit of an issue in some facilities.

## Changelog
:cl:
balance: rebalanced Flesh Idol
/:cl:


Kitsunemitsu (approved):
Approved. Looks good to me.

Kitsunemitsu:
Basically, Flesh idol rn is like on the cusp of being very good, and I think this PR will help

PR #2886, opened 2025-04-30
Damage Type Shuffle Uniform RNG + abno resistances in abno docs

Halonexus:
## About The Pull Request
The mapping logic was causing a weird probability of picking certain mappings:
2025-04-30 21_58_07-Window
Half of the mappings were 5 times more likely to be chosen than the rest.
I couldnt figure out an algorithm to get a uniform distribution so I just manually defined each of the 17 mappings.
2025-04-30 22_32_03-Window
Now its about 5.9% for each mapping.
Also made abnormality records show remapped resistances and breach damage ( some customised info docs might not get converted but should work for most things).

## Why It's Good For The Game
Should be less likely to get same mappings in a row.
Easier to see new resistances.

## Changelog
:cl:
tweak: Changed the way damage type shuffle mappings were chosen to be more uniform.
tweak: Abnormality info docs should show remapped resistances and attack type.
/:cl:


Kitsunemitsu (approved):
This is Sadly the best way to do it haha

PR #2913, opened 2025-05-16
Mini Col Update: BPU Update

EnderMage99:
## About The Pull Request

A Small PR which adds a clinic and association based BPU, along with a small tool for admin events.

1. The only roles that work in the Clinic are able to sign up to the clinic BPU, which is free and does not reduces attributes on clone.
2. Same for the association and their BPU, however that one is not free.
3. The tool is a reusable k-corp ampule which heals all brute damage from dead bodies.

## Why It's Good For The Game

1. The clinic is an important role which NEEDs some way of coming back to life if they are killed, due to them being the only ones who can revive fixers.
2. Association Fixers have attribute requirements for their gear and if they are gibbed and are given a cheap body, they will lose access to all of their skills and their gear. This BPU will allow a simple way of keeping their skills after getting gibbed.
3. Sometimes, Admin events can get a bit bloody, and the doc needs a faster way of bringing fixers back.

## Changelog
:cl:
add: clinic and association bpu, also k-corp quickfix
/:cl:

### DIFF
--- a/ModularTegustation/lc13_machinery.dm
+++ b/ModularTegustation/lc13_machinery.dm
@@ -389,7 +389,7 @@
 	if(SSmaptype.maptype == "office")
 		public_use = TRUE
 		clone_delay_seconds = 60
-		revival_attribute_penalty = -2
+		revival_attribute_penalty = 0
 		cost_multiplier = 2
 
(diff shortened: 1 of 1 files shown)


Coxswain-Navigator:
Development for this PR is no longer ongoing on this branch.

PR #2917, opened 2025-05-19
Repeat Crafting. Allows mass crafting without the tedium.

Caldony:
## About The Pull Request

Allows people to repeat the same crafting recipe a set amount of time. You either add or substract from a number at the top right of the crafting window, and when attempting to craft an item, you will repeat the crafting until you either reach the designed amount, or the crafting is no longer possible due to lacking ingredients or tools.

Things that have been tested:
-This does not allow you to craft in bulk all at once, instead it simply starts crafting again the moment the last craft is over.
-The crafting window will appear busy as normal as long as you are crafting.
-Moving, running out of ingredients or no longer having the right tools will properly stop the crafting from taking place.

Things that would be great to have but did not bother with:
-Ideally, setting a higher repeat amount would also show the increased cost of the recipe on the list, however the list is designed with a proc that is meant to be updated infrequently due to the high amount of data it needs to parse though, and I do not wish to alter the crafting code more than necessary.
-Being able to set the amount of repeat you want manually instead of having to click the + & - would be great, but I'm bad at javascript, and wanted the window to be fast and easy to use. So clicking the value will instead just reset it back to 1.

Things to note, I am not used to javascript, and haven't coded anything serious in a while. I've bugtested this with most eventualities I could think of, but as the crafting code is almost core to SS13, it might need some live testing. However, this should not change how crafting actually works, it's only for convenience.

## Why It's Good For The Game

If a chef wants to make 10 chicken nuggets, they need to click once, wait a few seconds for the nugget to be made, then click again, then repeat that ten times. This is even more tedious with Carnival and their massive silk transfer. This change allows people to craft a lot of things without the needless tedium.

## Changelog
:cl:
tweak: The crafting menu now lets you set the amount of things you want to craft in a row.
/:cl:


Kitsunemitsu:
Welcome back, King.

InsightfulParasite (approved):
It seems to work fine after some tests. Approved.

PR #2923, opened 2025-05-24
Pink Midnight update 2025

Kitsunemitsu:
## About The Pull Request
Adds scaling to pink midnight
Makes pink midnight breach more abnos over time

## Why It's Good For The Game
Pink needed a small update.

## Changelog
:cl:
balance: rebalanced pink midnight
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/misc.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/misc.dm
@@ -9,8 +9,8 @@
 	layer = LARGE_MOB_LAYER
 	pixel_x = -16
 	base_pixel_x = -16
-	maxHealth = 2000
-	health = 2000
+	maxHealth = 20000
+	health = 20000
 	melee_damage_type = PALE_DAMAGE
 	rapid_melee = 2
 	melee_damage_lower = 14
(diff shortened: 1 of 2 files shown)


EnderMage99:
Reviewing this Codewise, there is currently a bug with breachtime, which causes the value to always be 0, due to you using min(). Which, after testing, would just prevent the proc from happening. This can be fixed by changing it to max() instead of min().

code/modules/mob/living/simple_animal/hostile/ordeal/misc.dm
```diff
@@ -25,36 +25,69 @@
...
 	animate(src, alpha = 0, time = 5 SECONDS)
 	QDEL_IN(src, 5 SECONDS)
 	..()
 
+/mob/living/simple_animal/hostile/ordeal/pink_midnight/proc/Breach_Loop()
+	if(!length(whitelist))
+		return
+	if(!src)
+		return
+	var/breachtime = min(0, 40-30*TOUGHER_TIMES(playerscaling*3)+timescaling)
```
EnderMage99, line 55: Reviewing this Codewise, there is currently a bug with breachtime, which causes the value to always be 0, due to you using min(). Which, after testing, would just prevent the proc from happening. This can be fixed by changing it to max() instead of min().

PR #2928, opened 2025-05-26
Green Midnight Buff 2025/05

Destrok171:
## About The Pull Request
**PLEASE DO NOT MERGE THIS FOR A WHILE IT NEEDS SPRITES**

Reworks the Helix of the End from the Green Midnight Ordeal.

- Balance: Health nerf from 50,000 to 40,000 to account for other changes
- Laser phase: Now creates small BLACK damage AOEs akin to Waxing of the Black Sun's breach (only 1 tile, there are a lot of them)
- Laser phase: Now creates BLACK damage AOEs (they have a radius of 1, so the epicenter and adjacent tiles)
- Recharging phase: Now fires off drop pods containing Green Dawn and Green Noon enemies. This has a maximum. The size of the squads deployed will increase as the Helix loses health. The size of the squads also increases faster if you have more agents when the ordeal started.
- Scaling: Minions spawned scale off of living Agents, ERAs and DOs (this will have to be looked at once ERAs become CRAs)
- Scaling: Laser takes longer to recharge for fewer players.

As the fight gets more hectic, staying close to the Helix to deal damage becomes much more difficult and you will have to gauge carefully how much space you have left, and you will definitely need to bring some high DPS BLACK or at least WHITE weapons, AOE weapons will also shine here.

Since there will be actual enemies able to attack you, dumping your armour isn't an option and you have to actually consider a defensive option for the Ordeal.

## Why It's Good For The Game
Green Midnight is the most boring Ordeal that feels like a waste of time if you've been preparing for it for over two hours. It basically has 0 kill pressure on players and you will only ever die to getting overconfident. These changes should make it slightly more interesting, I'm willing to make further changes if the playerbase isn't receptive.

## Changelog
:cl:
add: Sprites for Helix launching drop pods, custom drop pod
tweak: Helix now spawns enemies and causes AOEs to appear during laser phase
balance: Helix health decreased
/:cl:


code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
```diff
@@ -204,9 +243,178 @@
...
+//They are mainly a nuisance that causes people stacking in tight spots to move and shuffle and get eachother killed. Encourages not dumping BLACK armour for the ordeal
+/obj/effect/temp_visual/helix_minilaser
+	name = "helix of the end mini-laser"
+	icon = 'ModularTegustation/Teguicons/32x64.dmi'
+	icon_state = "pillar_strike"
+	duration = 15
+	color = "#7ac21f"
+
+/obj/effect/temp_visual/helix_minilaser/Initialize()
+	..()
```
Gboster-0, line 273: initialize must always have a return value
```suggestion
	. = ..()
```

code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
```diff
@@ -44,6 +44,18 @@
...
+	var/squad_size = 6 // How large our squad of deployed bots should be. Will attempt to deploy bots until we reach this size. This increases when losing enough HP.
+	var/active_minions = 0 // How many of our personally spawned bots are alive right now.
+	var/maximum_squad_size = 16 // We will never increase squad size past this. Adjusted by scaling
+	var/squad_size_increase_step = 2 // This variable controls how many bots are added to a squad per 10% HP lost. Adjusted by scaling
+
+	// I am placing these here because I don't want to calculate turfs in range 9 trillion times.
+	var/list/microbarrage_threatened_turfs = list()
+	var/list/macrolaser_threatened_turfs = list()
+	var/list/danger_close_turfs = list()
+	var/list/microbarrage_target_turfs = list()
```
Gboster-0, line 57: Make sure to set all of these new lists to `null` in `Destroy()` or it'll hard delete
Hard deletes happen when you hold references to other things or when other things hold references to you when you are qdeleted
They cost extra processing time to get done, so you wanna make sure you dont hold them.

code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
```diff
@@ -204,9 +243,178 @@
...
+	icon_state = "warning"
+	duration = 21
+	color = "#7ac21f"
+	pixel_x = -32
+	base_pixel_x = -32
+	pixel_y = -32
+	base_pixel_y = -32
+
+/obj/effect/temp_visual/helix_macrolaser/Initialize()
+	..()
```
Gboster-0, line 295: Another case of no return value on Initialize()
```suggestion
	. = ..()
```

code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
```diff
@@ -44,6 +44,18 @@
 	/// Below that health the green midnight speeds up
 	var/next_health_mark = 0 // Initialize below
 
+	//These variables control how many bots are deployed during the no-lasers-phase and how the scaling for them works.
+	var/squad_size = 6 // How large our squad of deployed bots should be. Will attempt to deploy bots until we reach this size. This increases when losing enough HP.
+	var/active_minions = 0 // How many of our personally spawned bots are alive right now.
+	var/maximum_squad_size = 16 // We will never increase squad size past this. Adjusted by scaling
+	var/squad_size_increase_step = 2 // This variable controls how many bots are added to a squad per 10% HP lost. Adjusted by scaling
```
Gboster-0, line 51: When doing comments like these you wanna put them above the variable and add an extra slash, that way if you hover over the variable anywhere else in the code in visual studio it'll tell you what the comment says.
You can even multi-line these if you want like in the first example if the comment is very large
```suggestion
	/// How large our squad of deployed bots should be.
	/// Will attempt to deploy bots until we reach this size. This increases when losing enough HP.
	var/squad_size = 6
	/// How many of our personally spawned bots are alive right now.
	var/active_minions = 0
	/// We will never increase squad size past this. Adjusted by scaling
	var/maximum_squad_size = 16
	/// Controls how many bots are added to a squad per 10% HP lost. Adjusted by scaling
	var/squad_size_increase_step = 2
```

code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
```diff
@@ -204,9 +243,178 @@
...
+		//We want the last 40% of our budget to be comprised of Green Noons. They are the priority spawn here
+		if(remaining_spawn_budget <= (squad_size * 0.4))
+			squad += /mob/living/simple_animal/hostile/ordeal/green_bot_big/factory
+		else
+		//If we have budget to spare, we can get one of these I guess. Honestly by this point in the shift, they are irrelevant, EXCEPT the Syringe Bots. They are lethal.
+		//These are specifically the /factory subtype because we don't want a lobillion dead bodies to be left behind.
+			squad += pick(list(
+			/mob/living/simple_animal/hostile/ordeal/green_bot/factory,
+			/mob/living/simple_animal/hostile/ordeal/green_bot/syringe/factory,
+			/mob/living/simple_animal/hostile/ordeal/green_bot/fast/factory))
```
Gboster-0, line 320: dont do multi-line lists like this.
if you are multi-lining a list you should indent all its contents and make the end sit on its own line at the end of the list, indented to the same level as the beginning.
Also add a trailing comma, if you do then adding or removing objects from the list will only count as changing 1 line instead of 2 when commiting changes.
(am i making this sound over-complicated?)
```suggestion
			squad += pick(list(
				/mob/living/simple_animal/hostile/ordeal/green_bot/factory,
				/mob/living/simple_animal/hostile/ordeal/green_bot/syringe/factory,
				/mob/living/simple_animal/hostile/ordeal/green_bot/fast/factory,
			))
```

code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
```diff
@@ -204,9 +243,178 @@
...
+		remaining_spawn_budget--
+	return squad
+
+//This proc will deploy the remaining spawn budget (squad size - currently active bots) through drop pods.
+/mob/living/simple_animal/hostile/ordeal/green_midnight/proc/DeployBotSquad()
+	//We get the types we want to spawn in a list here
+	var/list/squad = GenerateBotSquad()
+
+	if(squad.len <= 0)
```
Gboster-0, line 329: instead of `list.len` do `length(list)`, its generally better and may or may not be faster. But its accepted as better
```suggestion
	if(length(squad) <= 0)
```

code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
```diff
@@ -204,9 +243,178 @@
...
+		H.deal_damage(50, BLACK_DAMAGE)
+		to_chat(H, span_userdanger("You're hit by [src.name]!"))
+
+
+//This proc is called to return a list with all the types that should be spawned by the next deployment. DeployBotSquad() trusts whatever list is given to it, so we
+//make sure we don't generate too many mobs in this proc (else midnight could spawn infinite mobs)
+/mob/living/simple_animal/hostile/ordeal/green_midnight/proc/GenerateBotSquad()
+	var/list/squad = list()
+	var/remaining_spawn_budget = squad_size - active_minions
+	while(remaining_spawn_budget > 0)
```
Gboster-0, line 310: minor nitpick, `while()` loops should not really be used unless you know very well what you are doing.
Ya might wanna consider changing it for a `for()` loop instead, yet do feel free to just mark this as complete whenever you read it.

Gboster-0 (commented):
Not greatest code quality but ive seen worse.

code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
Destrok171: I went ahead and did this, but this should only happen when a Helix is gibbed by Smile or manually deleted by an admin, right? I checked on its dead body when it died regularly after changing this, but it still had the lists with the turf references. Is that fine?

And thank you for teaching me about this, where is stuff like these hard deletes documented? Is it BYOND documentation or is it part of SS13 code?

code/modules/mob/living/simple_animal/hostile/ordeal/green/midnight.dm
Gboster-0: It is because they can become infinite, that and its easier to read for() loops than while() loops.

PR #2933, opened 2025-05-28
Fixes Drifting Fox works

Destrok171:
## About The Pull Request
Couple issues here:
- It originally removed you from the petting list every time it tried to calculate work chances. This included opening the Work Console, so if you pet the fox and then opened the console, it'd count as you not having pet it
- You could be added to the petting list several times. As in, if you pet it 10 times, you'd get put in the list 10 times. Funnily enough this helped counteract that previous bug
- It sent the "This abnormality seems to like..." message every time it calculated work chances
- Most important issue: It was missing a . = ..() in WorkChance(). It had a return value if you pet the fox, which would take the normal chances and add 30% for Attachment or remove 10% for anything else. The problem is, if you didn't pet it, there was no return, so it returned... probably NULL by default, which ended up using abysmal work rates, to the point that original Instinct or Insight (45 or 40 respectively) WITH the -10 subtraction from petting it and doing an incorrect work type, ended up being MORE than the chance you got from the buggy code. This ended up in players ALWAYS petting the fox even for Instinct and Insight because otherwise it would actually just go 0/18 PE and breach and kill them.

Everything has been fixed in my testing. Fox works will now be as such: 
- 45% Instinct, down to 35% if you pet it
- 40% Insight, down to 30% if you pet it
- 15%/20%/25%/30%/35% Attachment, with +30% added if you pet it 
- 15% Repression, down to 5% if you pet it

I have not touched the balance per se, but de facto Fox is going to become a much better trainer so it may warrant nerfs. Or not, barely anyone takes it anyway.

## Why It's Good For The Game
Drifting Fox won't be broken

## Changelog
:cl:
fix: Drifting Fox abnormality works now handle petting properly
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/drifting_fox.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/drifting_fox.dm
@@ -91,19 +91,25 @@
 		However, if the umbrellas are broken you will lose 5% for each umbrella broken.<br></b>")
 
 /mob/living/simple_animal/hostile/abnormality/drifting_fox/funpet(mob/petter)
-	pet += petter
+	pet |= petter
 	return ..()
+
+/mob/living/simple_animal/hostile/abnormality/drifting_fox/AttemptWork(mob/living/carbon/human/user, work_type)
+	if(user in pet)
+		if(work_type == ABNORMALITY_WORK_ATTACHMENT)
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (approved):
Looks good to me, looks liek a simple bugfix

PR #2936, opened 2025-06-03
Indigo Noon Rework 2025/06

Destrok171:
## About The Pull Request

Adds variant sweepers to the Indigo Noon, and a few other minor changes. The Ordeal in general should still be weak to ranged kiting, but it will be harder to funnel them into doorways like agents usually do nowadays.
Sprites for the new sweepers were made by insiteparaful

Boring stuff first:

- **Indigo Noons were taken out of the /hostile/ordeal/indigo.dm file and placed into /hostile/ordeal/indigo/noon.dm**. 
The Persistence status effect was also placed into noon.dm, but it can be placed elsewhere or into its own file in the future.

- **Indigo Noon devour healing on other Sweepers buffed from 20 to 40.**

- **All injured Indigo Noons will now attempt to devour fallen Sweepers as they advance.** 
This is a very minor heal, but they are not wasteful.
- **All Indigo Noons have access to the "Persistence" status effect you may remember from Ruina, though it is slightly different.**
Persistence is a buff that can go from 1 stack to 3 stacks, which may prevent lethal damage on a Sweeper and heal them according to how many stacks they have, then losing a stack. Persistence will time out after 35 seconds, All Sweepers can gain 1 stack of Persistence by eating a corpse that doesn't belong to their faction. Chances for Persistence triggering are as follows:

1 stack - 50%
2 stacks - 75%
3 stacks - 100%

These chances are reduced by **overkill damage**, so using higher damage weapons like Totalitarianism is likelier to circumvent Persistence altogether.

Persistence also has a visual overlay to alert players that it is active, and it also makes a sound when it gets triggered.
- **Added Indigo Noon variant: Lanky.**
Lanky Sweepers move and attack faster than normal Sweepers, but deal less damage and have less health.
They have access to a **dash attack** that they can use, which will come off cooldown faster if it doesn't hit anyone. If it does hit someone, the Sweeper restores health and gains a stack of Persistence per target hit - however, it will go on cooldown proportional to the amount of people hit. They may enter an evasive mode after their dash attack, which speeds them up and makes them act erratically.

- **Added Indigo Noon variant: Chunky.**
Chunky Sweepers move and attack slower than normal Sweepers, but deal slightly more damage and have a lot more health.
They are able to use **Extract Fuel** every once in a while after being hit, which temporarily empowers their next hit, making it restore their own health and gain a stack of Persistence if it lands.
When brought to low health, they have a chance to use **333... 1973**, which gives them 3 stacks of Persistence.

## Why It's Good For The Game

More variants added to a simple Ordeal. I think most players appreciate the variance present in Indigo Dawn and Green Dawn, so bringing it to another Ordeal should be good. There is also potential to take the Persistence from these Noons and give it to other Indigo enemies, or to add these Noon variants to Dusk or Midnight.

## Changelog
:cl:
add: Lanky and Chunky Indigo Noon subtypes.
balance: Indigo Noon friendly devour healing increased.
imageadd: Variant sweeper sprites (lanky and chunky) by insiteparaful, Persistence overlay
refactor: Indigo Noon moved from indigo.dm to indigo/noon.dm
/:cl:


Kitsunemitsu (approved):
No issues in playtesting other than balance concerns, but those can be done in a hotfix PR. Ready to get this merged.

PR #2942, opened 2025-06-07
Amber Dawn Infestation

Caldony:
## About The Pull Request
What started as a simple attempt to buff amber dawns has grown a little bigger than I anticipated.

What does this PR do?
It makes amber dawns capable of infesting someone in critical condition. This person will proceed to spawn 3, 4 then 5 bugs over the course of 4.5 minutes out of their body, in which case the infection will stop. The spawning will continue even on the person's death, which should encourage people to destroy the body as soon as possible. Amber dawns spawned from the infection will count towards whatever ordeal is currently ongoing, but that will not stop new dawns from spawning after the ordeal is over if the infection is still active. Only humans can be infected, not monkeys, or abnos, or anything else.

Even if the person is saved from their critical condition, they will still have to survive those 4.5 minutes and the subsequent spawns from it. 

There is a way to stop the infection early, the first is to extract the parasite via surgery, the other method is to eat raw amber meat. This will spawn a buffed up version of an amber dawn, much stronger, harder to kill, and gets stronger with every kill. However, since this one tasted the flesh of its own kind, will now start attacking everything it finds, including other amber dawns. If it kills a human, it'll gib them to spawn more amber dawns that it'll immediately try to kill. At 10 kills of any kind, it'll explode into 30 amber dawns (which should require it to kill about 3 people first, assuming it doesn't find random bodies on the way.)

Ambers are incapable of infecting others during city spawn. The feature is somewhat unpolished and might need live testing to see what needs to be tweaked.

## Why It's Good For The Game

This is admittedly a lot of changes and complexity for dawns, that are supposed to be the simplest ordeals in the game, however I do believe it fits the theme and difficulty just fine. The dawns can only proliferate if they actually manage to get someone in crit, and the special dawn can only spawn if someone is actively fed amber meat, so it can't really happen on accident. This is supposed to give amber dawns the possibility of swarming out of control not due to a timer but to the amount of death in the facility, which is usually something entirely within the agent's control.

## Changelog
:cl:
add: Amber dawns can now infect people in critical condition.
add: The hungriest one will spawn if you eat amber meat while infected.
/:cl:

### DIFF
--- a/code/game/objects/items/food/meat.dm
+++ b/code/game/objects/items/food/meat.dm
@@ -912,7 +912,7 @@
 	name = "perfect meat"
 	desc = "An awful jiggling chunk of meat cut from the hide of a worm. It tastes foul and has the texture of spoiled jelly."
 	icon_state = "wormmeat"
-	food_reagents = list(/datum/reagent/consumable/nutriment/protein = 4, /datum/reagent/yuck = 5)
+	food_reagents = list(/datum/reagent/consumable/nutriment/protein = 4, /datum/reagent/yuck = 5, /datum/reagent/amber = 5)
 	tastes = list("slime" = 1, "jelly" = 1)
 	foodtypes = MEAT | RAW | TOXIC | GROSS
 
--- a/code/modules/mob/living/simple_animal/hostile/ordeal/amber.dm
+++ b/code/modules/mob/living/simple_animal/hostile/ordeal/amber.dm
@@ -33,6 +33,7 @@
 	var/burrowing = FALSE
 
 	var/can_burrow_solo = TRUE // False for amber dawns spawned by dusks that are still alive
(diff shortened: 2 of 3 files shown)


EnderMage99:
Looking at this PR code-wise, I think this PR is good. A thing I would recommend changing is making some of the values in your  /obj/item/organ/amber_bug into variables, so they can be more easily adjusted later. Such as the 1.5 MINUTES or the + 2 in 	var/bug_spawned = feeding_stage + 2. 

I am a bit worried about spawning 30 dawns due to the lag it could cause, even if the condition for it is very unlikely to be fulfilled. Under that same field, I would recommend placing some sort of limit on the number of mobs that can be spawned. Like checking if there are 50+ mobs in ordeal_reference, and not spawning past that limit. Otherwise, there is a chance that these bugs will also devour the server.

Kitsunemitsu:
How did I miss this PR?

PR #2946, opened 2025-06-10
Transfervar fix

Aseidos:
## About The Pull Request

The TransferVar proc had the index and the value variable switched, so your "value" became the index of the list and viceversa (ironically, the current implementations of this proc worked only because of this oversight)
This fixes that and adds a bit of comment that makes clear that the index cannot be a numerical value.

## Why It's Good For The Game

Fixing a proc that has been bugged for quite some time is good (Also I need this for a rework I am doing, so win-win).

## Changelog
:cl:
fix: TransferVar proc now works as intended.
/:cl:


Coxswain-Navigator (approved):
Awesomesauce.

PR #2950, opened 2025-06-16
Adds ego assimilation

vampirebat74:
## About The Pull Request

Adds a feature to certain realized E.G.O (specifically the "Effloresced" E.G.O seen in limbus company) that allows them to create their corresponding weapon by sacrificing any ALEPH weapon. This ability can only be used once per suit.

Additionally, wishing well now creates a global reference to itself, so that its lists can be fetched without calling world.contents

## Why It's Good For The Game

I had intended to add this feature a long time ago, specifically for gasharpoon. At the time, my understanding of general BYOND code was inadequate and I was frustrated by my slow progress.

Code optimization is good for the game. The wishing well changes will be expanded in a separate PR.

## Changelog
:cl:
add: Added a new realized ability
code: realized abilities added to gasharpoon, farmwatch and spicebush. Wishing well now creates a reference to itself as a global variable
/:cl:

### DIFF
--- a/code/modules/spells/ability_types/realized.dm
+++ b/code/modules/spells/ability_types/realized.dm
@@ -1,3 +1,66 @@
+/* E.G.O assimilation */
+/obj/effect/proc_holder/ability/ego_assimilation
+	name = "E.G.O assimilation"
+	desc = "Convert an ALEPH E.G.O into a weapon comptaible with your suit. Can only be used once."
+	action_icon = 'icons/obj/ego_weapons.dmi'
+	action_icon_state = ""
+	base_icon_state = "template"
+	var/target_type = /obj/item/ego_weapon/mimicry
+	var/obj/structure/toolabnormality/wishwell/linked_structure
+
+/obj/effect/proc_holder/ability/ego_assimilation/Perform(atom/target, user)
+	..()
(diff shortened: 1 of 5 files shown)


code/modules/spells/ability_types/realized.dm
```diff
@@ -1,3 +1,70 @@
...
+	name = "E.G.O assimilation"
+	desc = "Convert an ALEPH E.G.O into a weapon comptaible with your suit. Can only be used once."
+	action_icon = 'icons/obj/ego_weapons.dmi'
+	action_icon_state = ""
+	base_icon_state = "template"
+	var/target_type = /obj/item/ego_weapon/mimicry
+	var/obj/structure/toolabnormality/wishwell/linked_structure
+
+/obj/effect/proc_holder/ability/ego_assimilation/Click()
+	..()
```
Gboster-0, line 12: Any reason for this proc to be overriden without setting "." to a value?

Gboster-0 (approved):
Seems good enough from a code standpoint, should probably be checked for balance tho

Kitsunemitsu:
Just easier on us, while it will make these EGO more sought after, the abilities of the others will generally be preferred.

PR #2960, opened 2025-07-01
Fixes Plushie Extractor

Destrok171:
## About The Pull Request
Fixes the already-existing Plushie Extractor. Also refactors it from "obj/item/device/Plushie_Extractor" to "/obj/item/device/plushie_extractor" (removed uppercase from path since I'm pretty sure that doesn't fly). I also changed all of its references in corporation.dm, lcorp.dm and _belt.dm.

It's not quite as simple as that, I also changed it slightly.
The previous version of this item kept a static list variable to which it added the ckey of the user who used it. It never used this list of ckeys for any check, so I have no way of knowing what it was actually meant to do.

Now, we probably don't want to let people print 500 plushies in 1 minute, so the restriction I went with was a simple timer to recharge the item. Judging from the fact it's in a file called "unpowered.dm", battery charge isn't meant to be used here, so I just put it on a 3 minute timer after every use. Other than that, there are no restrictions on its usage. If the abnormality you use this on has a corresponding plushie mapped to it in the "output" static list, and the device is charged, you will be able to make a plushie out of it. Further restrictions can be added if deemed necessary, but like, it's just plushies, I don't think it's needed.

## Why It's Good For The Game
One of the rarest items in the L Corp input should be actually functional. Waste of PE/a crate otherwise.

## Changelog
:cl:
fix: Fixed plushie extractor.
refactor: Changed item path of plushie extractor.
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/gadgets/unpowered.dm
+++ b/ModularTegustation/tegu_items/gadgets/unpowered.dm
@@ -830,7 +830,7 @@
 	else
 		. += span_notice("It is currently recharging.")
 /// I'm unsure if this is necessary, but just to be safe, I want to delete the timer when the object is destroyed.
-/obj/item/device/plushie_extractor/Destroy()
+/obj/item/device/plushie_extractor/Destroy(force)
 	deltimer(charge_timer)
 	charge_timer = null
 	return ..()
--- a/ModularTegustation/tegu_items/gadgets/unpowered.dm
+++ b/ModularTegustation/tegu_items/gadgets/unpowered.dm
@@ -750,52 +750,93 @@
 	to_chat(user, span_nicegreen("You extract [target]'s gift!"))
 	qdel(src)
 
-/obj/item/device/Plushie_Extractor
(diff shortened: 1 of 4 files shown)


ModularTegustation/tegu_items/gadgets/unpowered.dm
```diff
@@ -750,52 +750,93 @@
...
+	. = ..()
+	if(charged)
+		. += span_nicegreen("It is currently charged.")
+	else
+		. += span_notice("It is currently recharging.")
+/// I'm unsure if this is necessary, but just to be safe, I want to delete the timer when the object is destroyed.
+/obj/item/device/plushie_extractor/Destroy()
+	deltimer(charge_timer)
+	charge_timer = null
+	. = ..()
```
Gboster-0, line 836: `Destroy()` should always end with `return ..()`
```suggestion
	return ..()
```

Kitsunemitsu (approved):

PR #2969, opened 2025-07-14
LCL Hostile Faction fix

KingSuperBOB:
## About The Pull Request
LCL abnos are given unique factions so their AoEs can hit each other to engage in PvP. However if a player possesses a abno before the round starts the faction changes will apply THEN the abno will have the Hostile Faction applied meaning its AoE still wont hurt other abnos with the hostile faction, this PR makes the hostile faction only be applied if theres no client, this solves the issue as LCL has multipled faction overrides applied to the player from logins, logouts and initialize

## Why It's Good For The Game
Abnormalities which only had AoE attacks got killed one-sidedly by abnormalities with melee attacks which closed off RP opportunities, this avoids that.

## Changelog
:cl:
tweak: lowered AI vision range to 3 (aggro vision range is still 7)
fix: fixed LCL abnos getting the wrong faction if the player joins too early
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/limbus_labs/!other_overwrites.dm
+++ b/ModularTegustation/tegu_items/limbus_labs/!other_overwrites.dm
@@ -28,14 +28,8 @@
 /mob/living/simple_animal/hostile/abnormality/Initialize()
 	. = ..()
 	if(SSmaptype.maptype == "limbus_labs")
-		vision_range = 7
+		vision_range = 3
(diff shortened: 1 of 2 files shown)


Kitsunemitsu (dismissed):
This doesn't work perfectly. It only checks on roundstart

Kitsunemitsu:
Ah makes sense. Cinema.

Kitsunemitsu (approved):

PR #2986, opened 2025-07-26
Replaces mentions of "City" faction with "Hostile"

InsightfulParasite:
## About The Pull Request
Let me explain to you my viewpoint.
The city faction was created as a nebulous unified faction for all hostile creatures in the city since they would fight eachother when in close proximity.
The hostile faction was created as a nebulous unified faction for all hostile creatures in lavaland and everywhere else since they would fight eachother when in close proximity.

This irritation came about because RCE had a NEW nebulous unified faction called "enemy" which has no purpose in its existence other than to later be corrected by future coders who see the redundancy. I tire of redundancy.

## Why It's Good For The Game
Reduces the amount of redundant code we have. If there is going to be "city", "hostile", and "enemy" and they have no interaction with eachother other than making the definition of default faction difficult to put into a static variable then i personally see it as a hinderance.

## Changelog
:cl:
tweak: Replaced "city" faction with nebulous "hostile"
/:cl:


Kitsunemitsu (approved):
Easy approval.

PR #2991, opened 2025-07-27
Expands Central Command in District4

vampirebat74:
## About The Pull Request

What it says on the tin.

Also maps Shelter from the 27th of March and Backwards Clock into District4, at the Briah layer.

Corroded security agents can now fire at point blank, with their ranged cooldown doubled.

The eastern wing of CC now lacks an open route for the abnormality to path through.

**Before:**
image

**After:**
image

## Why It's Good For The Game

Approaching the briah layer is the most dangerous part of this map. Most players will have their attributes anywhere between 40 and 55 at this point, while potentially contending with three or more WAW-level abnormalities. This, combined with the confined area, made it excessively difficult for unprepared players. This PR expands the initial area that players can explore upon reaching the layer, without needing to trip too many mining events.

## Changelog
:cl:
tweak: mapping changes
balance: ER mob nerf
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/enkephalin_rush/enkr_mobs.dm
+++ b/ModularTegustation/tegu_items/enkephalin_rush/enkr_mobs.dm
@@ -229,7 +229,7 @@
 	maxHealth = 600
 	health = 600
 	ranged = TRUE
-	ranged_cooldown_time = 5 SECONDS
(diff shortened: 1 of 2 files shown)


Coxswain-Navigator (approved):
Mapping PR

PR #2997, opened 2025-07-30
CoL Update: Killing Multitools/Door Hacking

EnderMage99:
## About The Pull Request

Makes all multitools delete themselves in city modes, (Also adds an admin multitool which doesn't remove itself). Also, adds a toggle to all doors, which makes them unhackable by default.

## Why It's Good For The Game

Multi-tools can be used to break doors, and in the current state, doors are used in CoL; they will be impossible to fix without admin help. So, making them hackable by default should fix the problem.

## Changelog
:cl:
tweak: tweaked doors to be unhackable
balance: removed multitools from city modes
/:cl:

### DIFF
--- a/_maps/map_files/Event/city.dmm
+++ b/_maps/map_files/Event/city.dmm
@@ -6422,16 +6422,6 @@
 /obj/item/clothing/suit/apron,
 /turf/open/indestructible/white,
 /area/city/shop)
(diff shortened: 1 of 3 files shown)


Kitsunemitsu (approved):
This is such a gross way to do it but it works for now/

PR #3004, opened 2025-08-01
Expedition Machinery

InsightfulParasite:
## About The Pull Request
New Draft code Structures for RCE.

Im very tired so im copy pasting what i said in the discord.
Okay so all the code in there NOW should work as demostrated in the videos
I didnt make a item to place down cross sections and due to the existence of tunnels you may want to just ignore that code.
All the machines are unfortunately dense so some care may be needed to prevent people from making barricades out of them.
Also if we still want corpse transport the tunnels currently only transport objects not living. So corpses will get stuck on them like gunk in a filter
All machines also process when actrivated but should deactivate when they run out of their desired thing

## Changelog
:cl:
add: R corp factory machinery
/:cl:


Cupax3:
Change the branch you want to commit to to be `vlggms/RCE`

PR #3015, opened 2025-08-06
Love and Justice Realization Buff

Destrok171:
## About The Pull Request

Rebalancing for the QOH realization - makes it stronger. 
Changes:
- Cooldown: 3 Minutes -> 1 Minute
- Base* Beam Damage: 10 -> 16
- Beam Stage Increase Threshold: 150 -> 225
- Insane Ally Base* Sanity Healing: 3 -> 12
- Now heals brute on all allies hit, but for a lesser amount if they're not in critical condition

\* When I say "Base" I mean that is the value before any coefficients are applied to it. In reality, beam damage is multiplied by 1.5 when the beam stage hits 2, and it is also then multiplied by our Justice: for an Agent with 130 Justice that means an additional 2.3 coefficient.

## Why It's Good For The Game

This Realization is extremely bad at the moment in comparison to anything else you can get. It has a particularly long cooldown, low damage, not excellent support capabilities compared to other realizations... It's boring, because you barely get to use it at all. Compare to a realization like Quenched with a fun and powerful ability, or Eulogy or Big Iron which give massive upgrades to their respective weapon which is already fun to use, or even statstick realizations like Wit of Cruelty that just give you pretty good armour...

Needless to say, at present, this Realization is only ever taken so you can enable the speech and hotmic it to the whole facility. Hopefully these changes should put it more in line with the viability of other Realizations.

## Changelog
:cl:
balance: Buffed Love and Justice Realization
/:cl:

### DIFF
--- a/code/modules/spells/ability_types/realized_aimed.dm
+++ b/code/modules/spells/ability_types/realized_aimed.dm
@@ -156,7 +156,7 @@
 		Alt-Click to toggle speech. Crtl-Click to set your own speech."
 	action_icon_state = "arcana0"
 	base_icon_state = "arcana"
-	cooldown_time = 3 MINUTES
+	cooldown_time = 1 MINUTES
 	base_action = /datum/action/spell_action/ability/item/ego_arcana_slave
 
 	var/list/spawned_effects = list()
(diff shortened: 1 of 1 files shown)


Kitsunemitsu (approved):
Yeah sure, this is a hot ass of a Realization so

PR #3027, opened 2025-08-10
Disables unfinished/badly implemented ordeals

EgorDinamit:
## About The Pull Request

Temporarily removes pink and steel ordeals from rotation, as well as indigo midnight.

## Why It's Good For The Game

Those are all either unfinished or simply implemented in a bad way.

Gold ordeals (especially midnight) are pending a proper rework.

In the meantime it will give us time to properly balance the ordeals that have been more or less okay this whole time.


Coxswain-Navigator (approved):
Probably better than leaving these as is.

PR #3029, opened 2025-08-10
LC Armor Values

EgorDinamit:
## About The Pull Request

EGO Armor values are displayed in correct style.
Additionally, it has been placed in default examine message, instead of a link.
image

## Why It's Good For The Game

Something that is much easier to understand.
Saves players from clicking too often.


CrabbytheCrab (approved):
Looks good

CrabbytheCrab (changes requested):
Though I would keep the "can not use ego" message for not having the requirements

EgorDinamit:
> Though I would keep the "can not use ego" message for not having the requirements

This is already displayed by the span style.

CrabbytheCrab (approved):
ok its fine then

PR #3033, opened 2025-08-11
Don't Touch me's span is more reasonable

Anonmare:
## About The Pull Request
Whilst I accomplished my goal of making the button impossible to miss - it ended up being _too_ big. This should keep it eyecatching and the addition of a sound cue ought to be get people's attention without it being the _only_ thing you can notice. The sound cue also shouldn't be too annoying if multiple people spammed the button for whatever reason.

## Why It's Good For The Game
This should be more reasonable

## Changelog
:cl:
tweak: Don't Touch Me's span is made a more reasonable size whilst remaining eyecatching
/:cl:


Gboster-0:
Please test your PR's before putting them on github, that and you made it so the button plays a sound on itself for every player that exists, dont do that

PR #3054, opened 2025-08-17
Ego Armor Slowdown Removal but take twice as long to equip

CrabbytheCrab:
## About The Pull Request

Removes the slow down for holding 2 armors/holding 1 while wearing one but doubles the time it takes to equip it.

## Why It's Good For The Game
The slowdown system had many bugs like how it wouldn't affect you if you equip one of the 2 armors you were holding or how it interacted with other speed changes from armors. the slow down was overall slower than just dragging the armor around. I doubled the time to equip armors to compensate since that does a better job of preventing players from switching armors in risky situations like when a pale abno is melting,

## Changelog
:cl:
del: Removes ego armor slowdown
balance: Ego armors take twice as long to equip
/:cl:


Coxswain-Navigator (approved):
Seems to work pretty well.

PR #3059, opened 2025-08-22
Content Cull & Miscellaneous Updates

Coxswain-Navigator:
## About The Pull Request
Removes a lot of old/abandoned content that is either no longer supported or unfinished and will never be finished.

**Removals :**

- An extra, unused portion of distortion code.
- Unbalanced Halloween races (simply removed from being roundstart choices).
- Plushie Extractor.

**Additions :**

- A new district origination to replace the old Terragov relation.
-  Desc update on various LC tools.
- Agent scanner now scales with prudence.

## Why It's Good For The Game
Less bloat that doesn't work half of the time.

## Changelog
:cl:
del: Removed old things
/:cl:


code/__DEFINES/preferences.dm
vampirebat74, line 1: Nitpick, but districts 1-3 do exist.

code/_globalvars/lists/flavor_misc.dm
vampirebat74, line 1: ditto here

code/__DEFINES/preferences.dm
Coxswain-Navigator: We're not going to let players assume they are from the head.

code/_globalvars/lists/flavor_misc.dm
Coxswain-Navigator: We're not going to let players assume they are from the head.

code/__DEFINES/preferences.dm
vampirebat74: They are districts. They do have backstreets and unsettled regions like Selva Oscura. I don't think that's going to be a big issue

PR #3061, opened 2025-08-24
Adjustable EGO Refactor

LanceSmites328:
## About The Pull Request
Moves Adjustable EGO out of a subtype into a Component that can be placed on any Clothing. This limitation is currently self-enforced by the Component.
The Component takes two arguments: A list of Alternative Styles, all sharing the same `icon` file, and a String of text displayed whenever you _do_ adjust the clothes.
If the first is not provided, the component doesn't Initialize.
If the second is not provided, it defaults to my default text.

## Why It's Good For The Game
Better code. The previous itteration was done out of laziness and the inability to learn Component code. This is a better implementation that allows it to apply to ANY clothing, allowing for adjustable hats and more. This adjustability is still entirely cosmetic; It does not account for damage zones or anything but that should be fine because our armor is Outfit entirely.

## Changelog
:cl:
refactor: /ego_gear/adjustable -> /datum/component/adjust_clothing
refactor: adjustable code from _ego_gear.dm -> /code/datum/component/adjustable_clothes.dm
/:cl:

### DIFF
--- a/ModularTegustation/tegu_items/refinery/crates/syndicate.dm
+++ b/ModularTegustation/tegu_items/refinery/crates/syndicate.dm
@@ -15,7 +15,7 @@
 	rareloot =	list(
 		/obj/item/clothing/suit/armor/ego_gear/city/blade_lineage_cutthroat,
 		/obj/item/clothing/suit/armor/ego_gear/city/thumb_capo,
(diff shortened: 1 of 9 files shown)


Coxswain-Navigator:
Are there more adjustable clothes than what is included in the PR, or just these two? What about realizations and hoods?

PR #3063, opened 2025-08-25
Prevents Big Bird from accidently have extra range from him attacking while moving

tile4:
## About The Pull Request
 What the title says. It prevents Big Bird chomping midwalk so it should avoid the accidental 2 range.

## Why It's Good For The Game
Less unfair instakills

## Changelog
:cl:
fix: fixed Big Bird having extra range sometimes
/:cl:


vampirebat74 (approved):
We're likely to do a more complete revision of Big Bird in the future, but this is a good place to start.

PR #3071, opened 2025-09-01
Experimental Lob Changes

CrabbytheCrab:
## About The Pull Request

- Dual Healing/Quad Shield bullets costs 2 lob.
- Pale Shield costs 1 lob.
- HP/SP/DH Bullets heal a flat 15 hp/sp but heal half as much to working agents.
- Healing Bullet upgrades increases healing by +10 for 1 lob, up to +50(bullets heal for 65).
- Shield bullets starts at 20 shield hp, the rest of their mechanics stay the same.
- Bullet Count costs 1 lob per upgrade but only gives 2 bullets and the cap has been lowered to 24 bullets and now retroactively applies to the manager camera console instead of updating when a meltdown starts.
- Extra Works now applies retroactively to the current work cap before a meltdown instead of updating when a meltdown starts.
- Bonus Meltdown Time costs 1/2/3 lob per upgrade and increases meltdown time by +10/20/30 seconds.
- Bonus Starting Stats costs 1/4/7 lob per upgrade and give +5/10/15 starting stats.
## Why It's Good For The Game

A lot of lob upgrades had issues where their scaling made them bad to fully upgrade due to diminishing returns while making their early upgrades give insane value. Making their values scale or making their price a flat amount should fix that while making some of them a more worth while to purchase. Also fixes some upgrades not applying retroactively.

Bullets are a bit of a mess balancing wise.
In their current state, Healing bullets are in a bad state due to them healing a low percent of health/sanity and their cost. This makes them horrible early game due to them only healing around 2-5 hp per bullet. Their cost is also an issue since to fully heal some one's health or sanity, it requires at most 3-5 bullets depending on healing upgrades which usually means that the manager can only do it 1-4 times before running out of bullets. All these issues stem from them used to and still are to some extent from being able to heal agents mid work, but balancing them around the fact they can work healing agents leads to them being only useful for that and bad everywhere else compared to shields.

Changing Healing Bullets healing to a flat amount of health allows them to be more comparable to Shield Bullets early game while being still having the benefited of multiple damage types at once. To compensate, their healing is reduced by half on working agents as a way to keep them from giving too much value in non combat situations and to dampen non investing in them.

Shield Bullets were broken to some extent due to their base health being 40, this caused them to double or even quadruple someone's health during the early part of the round while costing the same amount of lob to max out as Healing Bullets. Now base shield hp is 20 and pale bullets costs is increased to 1 lob to make Shield Bullets harder to max out compared to Healing Bullets.

TLDR: Makes the 2 agent buffing bullet types less one sided in terms of which one is the best option. Healing Bullets are weaker compared to Shield Bullets but are cheaper to max out and more versatile due to working on working agents and being better vs foes with multiple damage types. Meanwhile, Shield Bullets are the less versatile and more expensive compared to Healing Bullets but provides greater value, has a long duration, and synergizes with sources of healing.

## Changelog
:cl:
balance: rebalanced lob upgrades a bit
/:cl:


Coxswain-Navigator (approved):
For an "experimental" PR, these changes are surprisingly tame. LOB is getting completely overhauled in the cargo update anyways, so I don't see an issue with merging this.

PR #3087, opened 2025-10-20
Fixes map-exclusive and rare lobby music and splash screens

Coxswain-Navigator:
## About The Pull Request
Lobby music and splash screens haven't actually worked properly since... Since forever, really. This PR aims to fix that.

## Why It's Good For The Game
Stuff actually works.

## Changelog
:cl:
config: changes map names to be in line with how the code should read them.
/:cl:


vampirebat74 (approved):
PR looks good. Fixes a few old issues that have been neglected for years and enables lobby music that never worked previously. Participated in a live test that confirmed its functionality.

PR #3151, opened 2026-01-29
Ports Damage Refractors

CrabbytheCrab:
## Ports these prs all by [Destrok171](https://github.com/Destrok171),
https://github.com/Lobotomy-Corporation-13/lc13/pull/188 (This page has a section for the very important files, I would recommend seeing that since the important files changes for the port are the same here)
https://github.com/Lobotomy-Corporation-13/lc13/pull/228
https://github.com/Lobotomy-Corporation-13/lc13/pull/248
https://github.com/Lobotomy-Corporation-13/lc13/pull/254 (Only the aggro related stuff)

## About The Pull Request
Note this summery might suck a bit, sorry if it is

In terms of what I ported, all instances of apply_damage have been replaced with deal_damage(). deal_damage() has some new arguments being `deal_damage(damage_amount, damage_type = BRUTE, source = null, flags = null, attack_type = null, blocked = null, def_zone = null, wound_bonus = 0, bare_wound_bonus = 0, sharpness = SHARP_NONE)` which contains most of the original arguments from apply_damage added on but contains some new ones.

HurtInTurf() Also has arguments for flags and attack_type

Essentially, you can now define the source of the damage, what kind of move it is, and flags related to it. `code/__DEFINES/damage.dm` contains documentation about these flags. 
deal_damage() now calls `PreDamageReaction()` before it sends a COMSIG_MOB_APPLY_DAMGE  signal. If `PreDamageReaction()` returns false and the DAMAGE_FORCED flag isn't enabled, the attack won't go through.

Hostile mobs no longer gain aggro though bullet_act or attack_by, instead they have an override of deal_damage() which calls `RegisterAttackAggro()`. `RegisterAttackAggro()` is a proc that calls RegisterAggroValue() and path finds to the source of the damage if they had no target beforehand with a cooldown.

Before deal_damage returns the final_damage value, it'll call `PostDamageReaction()`, a proc that can function as an all in one for causing something when an mob takes damage like counters or checking health thresholds or changing damage values before deal_damage finishes. Most overrides of apply_damage was replaced with this.

Also adds deal_split_damage, which is like deal_damage, but can be split between damage types evenly.

COMSIG_MOB_APPLY_DAMGE now has information relating to source, flags, and attack_type.
Also adds `AGGRO_DAMAGE `, a ""damage type"" which updates RegisterAggroValue() on hostile mobs
Thrown weapons damage can also now heal panicking humans and be prevented.
Thrown weapons and special moves can now effect Aggro.

Ports `COMSIG_MOB_AFTER_APPLY_DAMGE`, which is like `COMSIG_MOB_APPLY_DAMGE` but fires after most of the damage stuff is done. I decided to slightly redo Fallen Color's passive effect to make use of it and to deshitify it a bit.

Also ports `COMSIG_HOSTILE_GAINEDTARGET`, which is fired when an hostile mob gains a target, It has a bit flag called `COMPONENT_HOSTILE_REFUSE_AGGRO` which can be returned to prevent GiveTarget() from changing an entities target. As a result, most target changes were replaced by either FindTarget/GiveTarget or LoseTarget(). LoseTarget() also now has an argument called `stop_movement = TRUE` which stops the hostile mob's movement if true
The port also refractors how some Mob's counters work like White Fixer, Smock, Everything there, and So Nobody Will Cry.
The port also makes Mouth of God's ability unable to be activated if the source was from the wearer.

Also lowers the damage of some stuff that weren't changes from the hp rework.

Also changed how blocking an attack with Black Swan's weapon works so it actually shoots the goo at the attacker instead of a random enemy.

## Why It's Good For The Game

All of this allows for mobs to track aggro from thrown items/other thing and also allows for more interesting mechanics related to counter attacks or reacting to taking damage. I ported it mainly due to the latter since I have some ideas for reworks for various things that would require it

## Changelog
:cl:
tweak: tweaked a few things
balance: rebalanced a few ego weapons and armor
code: removed most instances of apply damage
refactor: refactored deal damage
/:cl:


code/__DEFINES/damage.dm
Coxswain-Navigator, line 1: I can understand the attack_types for melee and ranged, and you could also justify special. But what do we need the rest for? If we don't have even a single instance of using them for anything, porting them is meaningless.

code/modules/clothing/suits/ego_gear/realized.dm
```diff
@@ -454,38 +454,39 @@ No Ability	260
...
 	for(var/turf/T in view(3, user))
 		new /obj/effect/temp_visual/revenant(T)
-		for(var/mob/living/L in T)
-			if(user.faction_check_mob(L, FALSE))
-				continue
-			if(L.stat == DEAD)
-				continue
-			var/atom/throw_target = get_edge_target_turf(L, get_dir(L, get_step_away(L, get_turf(src))))
-			L.throw_at(throw_target, 1, 1)
-			L.apply_damage(5, WHITE_DAMAGE, null, L.run_armor_check(null, WHITE_DAMAGE), spread_damage = TRUE)
```
Coxswain-Navigator, line 488: You changed damage numbers here. No mention of it in the PR's description. This is not a balance PR.

Coxswain-Navigator (dismissed):
I suggest removing ATTACK_TYPE_THROWING, ATTACK_TYPE_ENVIRONMENT, and ATTACK_TYPE_STATUS.  They can be combined into ranged, other, and other respectively. If you have a justification for them being used, we can talk it over whenever I get a chance.

code/modules/clothing/suits/ego_gear/realized.dm
CrabbytheCrab: I fogtot about these Sorry

Coxswain-Navigator (approved):
Yeah looks good to merge. I'm not going to claim that I went over all 350 files, but the fundamentals have been reviewed and this is mostly a port anyways.

PR #3158, opened 2026-02-14
Reduces the damage and range of Nameless Fetus

CrabbytheCrab:
## About The Pull Request

Due to a big lapse in judgment I made fetus deal way too much damage without realizing how annoying was. Overall Fetus now doesn't have infinite range with its ability to breach while the damage done has been changed from 20 every 20 seconds in a 41x41 area to 8-12 damage every 10 seconds in a 31x31 area. This means that sure it has 50% more dps, but it due to it going off faster, it won't be able to rebreach the same abno multiple times as often while lasting much shorter. The amount of times it can cry has been reworked from 3 per person to a base of 5 times it can cry while each person adds 1 to that. Also makes it gift have its Lc stats.

## Why It's Good For The Game

Deshitifies how it breaches shit while making it not god awful in higher pops

## Changelog
:cl:
balance: Made fetus less painful
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/abnormality/he/nameless_fetus.dm
+++ b/code/modules/mob/living/simple_animal/abnormality/he/nameless_fetus.dm
@@ -126,7 +126,7 @@
 		satisfied = FALSE
 		hunger = 0
 		datum_reference.qliphoth_change(1)
-		visible_message(span_userdanger("The fetus starts to wimper but only for a moment..."))
+		visible_message(span_userdanger("The fetus starts to wimper, but only for a moment..."))
 	else
 		criesleft = 5
 		for(var/mob/living/carbon/human/H in GLOB.player_list)	//Way harder to get a list of living humans.


code/modules/mob/living/simple_animal/abnormality/he/nameless_fetus.dm
```diff
@@ -126,11 +126,12 @@
 		satisfied = FALSE
 		hunger = 0
 		datum_reference.qliphoth_change(1)
-		visible_message(span_userdanger("The fetus is starts to wimper but only for a second!"))
+		visible_message(span_userdanger("The fetus starts to wimper but only for a moment..."))
```
vampirebat74: Needs a comma

CrabbytheCrab: Added it

PR #3170, opened 2026-03-01
Fixes Steam Transport Machine being a damage sponge plus a slight bump to its lower melee damage

CrabbytheCrab:
## About The Pull Request

So Steam's hp scaling was never changed(being 1600 + (400 * gears)) meaning it can have up to 3200 hp with all 4 gears. I lowered it so that it its 400 + (150 * gears) meaning that it can still be pretty bulky late into the round without being a 3200 hp wall. Also fixed its lower melee damage calculation being 4 instead of 5

## Why It's Good For The Game

3200 hp steam machine is a fucking war crime

## Changelog
:cl:
balance: rebalanced Steam Transport Machine's health scaling
/:cl:


Coxswain-Navigator (approved):
I'm a bit concerned about the melee damage being too high, but we can remedy if it becomes an issue.

PR #3176, opened 2026-03-09
Changes to player deletion

Coxswain-Navigator:
## About The Pull Request
Since players can be recovered through the new revival update, there's less reason to delete a player's items when they die. Also ensures that death() is called when players do die.
This PR likely does not currently cover every instance of this happening, but it should cover most.

## Why It's Good For The Game
Game is balanced in a way that makes sense.

## Changelog
:cl:
balance: eliminated player deletions that don't call death() or destroy the player's items.
/:cl:


vampirebat74:
So true

vampirebat74 (approved):

vampirebat74:
All I need is for you to remove the changes to Skin Prophet, since the rework removes them anyways.

PR #3182, opened 2026-03-25
fix coxswain's ci

Inorien:
something isn't working again, this PR is to figure out whats going on in the CI


Inorien:
fix seems to just be adding 
```
sudo dpkg --add-architecture i386
sudo apt-get update -qq
sudo apt-get install -y libcurl4t64:i386
```
to the `compile_all_maps` and `run_all_tests` stages, this time it looks like its the 32-bit libcurl.so.4 issue 🫃🫃🫃

PR #3200, opened 2026-03-27
local codebase literally can not stop breaking its CI

Inorien:
* update pyYaml to something that works with newer setuptools
* update Pillow and pygit2 to something that works with python 3.12

the reason #3198 didn't work is that that file is only really relevant on Windows (it downloads a .exe), in the CI it uses whatevers on the PATH which is 3.12 on this runner


Inorien:
checkem

PR #3217, opened 2026-04-07
Reworks the staining rose

Coxswain-Navigator:
## About The Pull Request
Pretty much entirely reworks the staining rose abnormality. This version should overall better reflect the danger this abnormality represents.

STAINING ROSE CHANGES:

**OLD:**

- The first person to work on the staining rose became chosen.
- Anyone not chosen working on rose would get de-buffed.
- If the chosen lacked level 4 justice, they would also get de-buffed.
- If rose was not worked every 15 minutes, it caused weather that doubled everyone's damage taken for a few minutes.

Overall, the old staining rose was pretty inconsequential in that it could **not directly deal damage in any capacity outside of work damage**.
While the abnormality's abilities were a considerable downside, players generally considered the staining rose more annoying than dangerous. Historically, players often refused to even engage with this abnormality.

This in itself is unacceptable for an ALEPH-tier abnormality. Here are the changes...

**NEW:**

- To condense the many mechanics of the staining rose, the terms "Chosen", "Bloom", and "Stained" are used.
1. The "Chosen" refers to the player that is currently selected by the staining rose. This is identical to previous, except now the "Chosen" is randomly selected from a pool of qualified candidates after every "Bloom" stage and upon spawning.
2. To qualify, a player must have level 5 in at least one attribute OR be agent level 4 or above.
- Every work tick, the staining rose becomes "stained", visually turning red in color. When the abnormality is fully "stained", or the chosen ignores a meltdown, the "Blooming" phase begins.
- When "Blooming", the abnormality deals 50-999 PALE damage to the chosen, dies, and the cycle begins anew.

**The staining rose has two ways it can breach** - the spread of the "Hanahaki" disease, and the "bloom" stage.
- The "Hanahaki" disease is used for breaches that might not necessarily be the fault of the "Chosen" - in other words, it's more of a generalized breach effect that can affect the entire facility.
- This breach effect occurs when someone other than the "Chosen" works on the staining rose, or the Qliphoth counter reaches zero for any reason when no "Chosen" was selected.

**MISC:**
- Moved all abnormality-related diseases to their own folder.
- Since several old features such as the flower weather were removed, some other code associated with them has also been removed.
- Code improvement - Moved work tick failure/success sound effects from the do_work proc on consoles to the success/failure tick on the abnormality. This makes it much easier to modify with nearly no performance impact, as consoles are less mutable than abnormalities.

## Why It's Good For The Game
It's better than not having the staining rose at all, and it's another ALEPH that is actually interesting for players to navigate.

## Changelog
:cl:
tweak: reworked the staining rose.

/:cl:


CrabbytheCrab (approved):
It looks fine to me

PR #3229, opened 2026-04-14
Fixes Fetus yet again

CrabbytheCrab:
## About The Pull Request

So due to shitty coding, Fetus wasn't able to actually kill the chosen, now it can

## Why It's Good For The Game

Bug fixes are good

## Changelog
:cl:
fix: fixed fetus
/:cl:


Coxswain-Navigator (approved):
At some point I need to personally look at fetus and see if I can make its implementation a bit better. Why reward players for abusing respawns when that is the opposite of what we "should" be doing? My rant about nameless fetus aside, this does what it says it does.

PR #3261, opened 2026-05-20
Health and Pop up changes, fixes, and additions

CrabbytheCrab:
## About The Pull Request

Previously brute damage didn't show up on humans and same thing for white damage on insane people, So this pr fixes that by adding a proc to handle health loss instead of using brute loss. Also I decided to touch up a lot of the pop up icon visuals to look better and added ones for oxygen damage. Also now there are overlays on the heart icon for Toxic and Burn damage which gets more intense the more you take. Also makes the red flash appear better

## Why It's Good For The Game

It looks better plus tox/burn damage was a hard to tell that you took any.

## Changelog
:cl:
add: Added some overlays for having toxic or fire damage
fix: fixed some problems with damage icons
imageadd: added oxygen damage icons
/:cl:


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -775,7 +775,7 @@
 		clear_fullscreen("oxy")
 
 	//Fire and Brute damage overlay (BSSR)
-	var/hurtdamage = getBruteLoss() + getFireLoss() + damageoverlaytemp
```
Coxswain-Navigator, line 778: Why do the additional math here when the numbers are compared to maxHealth further down anyways? This makes the math all wrong.

Coxswain-Navigator: Disregard, these changes are actually necesary.

Coxswain-Navigator (approved):
We tested this already, and the code itself looks fine. This is a pretty important change.

PR #3267, opened 2026-06-04
Records officer rework

Coxswain-Navigator:
## About The Pull Request
Gives the RO the bit of love that it deserves. This is my attempt at (sort of) standardizing how officers work by removing inconsistencies between them. Might include some DO stuff in this PR as well.

**Changes**

- The records officer now spawns with all of their special equipment, no more having to know that it's in a closet somewhere.
- The player revive watch has been changed from a scanner to more of an instant revive device. It costs 50% of the RO's current HP to use.
- Changed the info printer's acronym from AIDS to AID, with the upgraded device being the AIDP.

**Additions**

- Added some Dante clock sounds.
- Added the haste watch, which buffs the RO and nearby players with an attack and movement speed buff for a short time. The RO is then given a debuff afterwards.
- Added the AIDP - a information display similar to the ones used by agents, except this one works instantly and can print out information. This should let records officers be significantly more useful in their actual intended purpose

**Removals**

- Removed the records secure closet, as the RO now spawns with these items. (There will be a replacement for these lockers in the cargo update.)
- Removed the time stop watch because it caused many, many bugs. The haste watch is its replacement.
- Removed the portable predictor because abnormality understanding now reveals work rates instead. It has been replaced with the printer.
- Removed the disciplinary general escape researcher which was already unused, as we have no plans to bring it back.
- Removed excessive comments from records officer code.

## Why It's Good For The Game
Trust me its just better. Quite a few things are being left unchanged until we make fundamental differences to the game in the cargo update, however.

## Changelog
:cl:
add: Added new RO gadgets
del: Removed old RO gadgets
tweak: RO now spawns with their equipment like other officers.
/:cl:


WonkyBacon:
It will be great to see the green watch that reserect get more value during key moments with the change from PE boxes to health of RO.

CrabbytheCrab (approved):
This was a good update to a very neglected role overall

PR #3297, opened 2026-08-14
Rend Rework

CrabbytheCrab:
## About The Pull Request

The 5 current Rend Debuffs now use STATUS_EFFECT_REFRESH instead of STATUS_EFFECT_UNIQUE, this allows them to be reapplied. I Also lowered their duration from 6 seconds to 5 seconds to compensate since most weapons have 100% update on those. Also removed some checks and sparks for weapons that applied them since they weren't needed anymore. 

## Why It's Good For The Game

Makes applying rends less janky looking

## Changelog
:cl:
balance: rebalanced rend to be reapply able but has a slightly shorter duration
/:cl:


Coxswain-Navigator (approved):
These seems like pretty good changes to me.

